PR #9, opened 2020-11-13
Document indexes

szarnyasg:
This PR aims to improve the documentation of indexes including [their current lack of persistence](https://github.com/cwida/duckdb/issues/1110).

A couple of questions:

1. The current grammar `CREATE [ UNIQUE ] INDEX [ name ] ON table ({ column | ( expression )})` is not 100% accurate as it does not allow for compound indexes.

2. The current [Indexes page](https://duckdb.org/docs/sql/indexes) in the documentation contains an incomplete table with the supported data types. I'm not sure whether it's worth putting this table here, particularly because all general-purpose types seem to be supported by `create index`:

```sql
create table t(
    my_bigint bigint,
    my_boolean boolean,
    my_blob blob,
    my_date date,
    my_double double,
    my_integer integer,
    my_real real,
    my_smallint smallint,
    my_timestamp timestamp,
    my_tinyint tinyint,
    my_varchar varchar);

create index idx_my_bigint on t(my_bigint);
create index idx_my_boolean on t(my_boolean);
create index idx_my_blob on t(my_blob);
create index idx_my_date on t(my_date);
create index idx_my_double on t(my_double);
create index idx_my_integer on t(my_integer);
create index idx_my_real on t(my_real);
create index idx_my_smallint on t(my_smallint);
create index idx_my_timestamp on t(my_timestamp);
create index idx_my_tinyint on t(my_tinyint);
create index idx_my_varchar on t(my_varchar);
```

### DIFF
--- a/docs/sql/data_types/overview.md
+++ b/docs/sql/data_types/overview.md
@@ -14,7 +14,7 @@ The table below shows all the built-in general-purpose data types. The alternati
 | date |   | calendar date (year, month day) |
 | double | numeric | double precision floating-point number (8 bytes) |
 | integer | int, int4, signed | signed four-byte integer |
-| real | float4 | single precision floating-point number (4 bytes)|
+| real | float, float4 | single precision floating-point number (4 bytes)|
 | smallint | int2 | signed two-byte integer|
 | timestamp | datetime | time of day (no time zone) |
 | tinyint |   | signed one-byte integer|
--- a/docs/sql/indexes.md
+++ b/docs/sql/indexes.md
@@ -15,7 +15,7 @@ Indexes are currently [not persistent](https://github.com/cwida/duckdb/issues/69
 
 <div id="rrdiagram1"></div>
 
-`CREATE INDEX` constructs an index on the specified column(s) of the specified table. Compound indexes on multiple columns/expressions are supported. We currently only support unidimensional indexes.
+`CREATE INDEX` constructs an index on the specified column(s) of the specified table. Compound indexes on multiple columns/expressions are supported. Currently unidimensional indexes are supported, [multidimensional indexes are not supported](https://github.com/cwida/duckdb/issues/63).
 
 ### Parameters
 
--- a/docs/sql/indexes.md
+++ b/docs/sql/indexes.md
@@ -15,15 +15,7 @@ Indexes are currently [not persistent](https://github.com/cwida/duckdb/issues/69
 CREATE [ UNIQUE ] INDEX [ name ] ON table ({ column | ( expression )})
 ```
 
-`CREATE INDEX` constructs an index on the specified column(s) of the specified table. Compound indexes on multiple columns/expressions are supported. We currently only support unidimensional indexes of the following types:
-
-| Name | Aliases | Description |
-|:---|:---|:---|
-| bigint | int8 | signed eight-byte integer |
-| integer | int, int4, signed | signed four-byte integer |
-| smallint | int2 | signed two-byte integer |
-| tinyint |   | signed one-byte integer |
-| varchar | char, bpchar, text, string| variable-length character string |
+`CREATE INDEX` constructs an index on the specified column(s) of the specified table. Compound indexes on multiple columns/expressions are supported. We currently only support unidimensional indexes.
 
 ### Parameters
 
--- a/docs/sql/indexes.md
+++ b/docs/sql/indexes.md
@@ -41,8 +41,10 @@ CREATE [ UNIQUE ] INDEX [ name ] ON table ({ column | ( expression )})
 ```sql
 -- Create an unique index 'films_id_idx' on the column id of table films.
 CREATE UNIQUE INDEX films_id_idx ON films (id);
--- Creates index 's_idx' that allows for duplicate values on column revenue of table films.
+-- Create index 's_idx' that allows for duplicate values on column revenue of table films.
 CREATE INDEX revenue_idx ON films (revenue);
+-- Create compound index 'gy_idx' on genre and year columns.
+CREATE INDEX gy_idx ON films (genre, year);
 -- Create index 'i_index' on the expression of the sum of columns j and k from table integers.
 CREATE INDEX i_index ON integers ((j+k))
 ```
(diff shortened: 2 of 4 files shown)


Mytherin:
> The current grammar CREATE [ UNIQUE ] INDEX [ name ] ON table ({ column | ( expression )}) is not 100% accurate as it does not allow for compound indexes.

That is correct; this is outdated information. When the ART was first added only a single expression was supported. This should be reworked. Ideally I would add a railroad diagram here similar to the other Create statements ([see e.g. Create View](https://duckdb.org/docs/sql/statements/create_view)). The examples should be moved up the page as well, and expanded to contain multi-column indexes.

The page should also be updated to include the new index join algorithm that can potentially make use of an index. 

> The current Indexes page in the documentation contains an incomplete table with the supported data types. I'm not sure whether it's worth putting this table here, particularly because all general-purpose types seem to be supported by create index:

Fair point, this is outdated information as well. When the ART was first added only integer types were supported. This can be removed now.

szarnyasg:
I discussed this PR with Hannes yesterday:
- he also agreed to drop the redundant table
- I'll add a railroad diagram

> The page should also be updated to include the new index join algorithm that can potentially make use of an index.
Cool! Are there any papers to cite on this?

szarnyasg:
Railroad diagrams:

image

image

Mytherin:
Thanks! Looks excellent now. One minor correction: the zonemaps (min/max indices of each column) are persisted and don't need to be recreated on startup. Only the ART is recreated on startup (and only for unique/primary key indices).

szarnyasg:
Okay, fixing that. Can you drop me a reference to the new index join algo?

Mytherin:
The index join can be automatically used by the system if an index exists on a join column. The system is rather conservative on using the index join, though, as the hash join is faster in many situations. The index join is primarily useful in "OLTP-style" queries where two tables are joined only to fetch a small number of rows. Typically index joins are only used when joining a really small table with a really big table, when an index exists on the really big table.

Index joins can be enforced using `PRAGMA force_index_join`. The explain will tell you which join type is used.

e.g.:
```sql


CREATE TABLE t1(i INTEGER PRIMARY KEY);
CREATE TABLE t2(i INTEGER);
insert into t1 select * from range(100);
insert into t2 select * from range(50, 100);
EXPLAIN select count(t1.i) from t1 inner join t2 on (t1.i = t2.i);
┌─────────────────────────────┐
│┌───────────────────────────┐│
││       Physical Plan       ││
│└───────────────────────────┘│
└─────────────────────────────┘
┌───────────────────────────┐                             
│      SIMPLE_AGGREGATE     │                             
└─────────────┬─────────────┘                                                          
┌─────────────┴─────────────┐                             
│         HASH_JOIN         │                             
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │                             
│           INNER           ├──────────────┐              
│            i=i            │              │              
└─────────────┬─────────────┘              │                                           
┌─────────────┴─────────────┐┌─────────────┴─────────────┐
│          SEQ_SCAN         ││          SEQ_SCAN         │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   ││   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             t1            ││             t2            │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   ││   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             i             ││             i             │
└───────────────────────────┘└───────────────────────────┘   

PRAGMA force_index_join;
EXPLAIN select count(t1.i) from t1 inner join t2 on (t1.i = t2.i);
┌─────────────────────────────┐
│┌───────────────────────────┐│
││       Physical Plan       ││
│└───────────────────────────┘│
└─────────────────────────────┘
┌───────────────────────────┐                             
│      SIMPLE_AGGREGATE     │                             
└─────────────┬─────────────┘                                                          
┌─────────────┴─────────────┐                             
│         INDEX_JOIN        ├──────────────┐              
└─────────────┬─────────────┘              │                                           
┌─────────────┴─────────────┐┌─────────────┴─────────────┐
│          SEQ_SCAN         ││          SEQ_SCAN         │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   ││   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             t2            ││             t1            │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   ││   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             i             ││             i             │
└───────────────────────────┘└───────────────────────────┘   
```

Here is an example where the index join will be selected (small table joined with big table):
```sql
create table t1 as select * from range(5) tbl(i);
create table t2 as select * from range(1000000) tbl(i);
create index art_index on t2(i);
explain select count(*) from t1 join t2 using (i);
┌─────────────────────────────┐
│┌───────────────────────────┐│
││       Physical Plan       ││
│└───────────────────────────┘│
└─────────────────────────────┘
┌───────────────────────────┐                             
│      SIMPLE_AGGREGATE     │                             
└─────────────┬─────────────┘                                                          
┌─────────────┴─────────────┐                             
│         INDEX_JOIN        ├──────────────┐              
└─────────────┬─────────────┘              │                                           
┌─────────────┴─────────────┐┌─────────────┴─────────────┐
│          SEQ_SCAN         ││          SEQ_SCAN         │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   ││   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             t1            ││             t2            │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   ││   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             i             ││             i             │
└───────────────────────────┘└───────────────────────────┘ 
```

szarnyasg:
Thanks, pushed a clarification on index join and forcing it with pragma.

Mytherin:
Thanks for the changes! I think this is ready to be merged now.

PR #12, opened 2020-11-15
Drop zero margin from unordered lists

szarnyasg:
Unordered lists didn't have enough space below them due to `margin: 0`.

## Before
screen_2020-11-15_12-36-05

## After
screen_2020-11-15_12-36-25


Mytherin:
Looks better indeed, thanks!

PR #19, opened 2020-12-08
add macro documentation

lnkuiper:
title explains it all

### DIFF
--- a/docs/sql/statements/create_macro.md
+++ b/docs/sql/statements/create_macro.md
@@ -1,0 +1,26 @@
+---
+layout: docu
+title: Create Macro
+selected: Documentation/SQL/Create Macro
+expanded: SQL
+railroad: statements/createmacro.js
+---
+CREATE MACRO - this statement creates a scalar macro in the catalog.
+
+### Examples
+```sql
+-- create a macro for a case expression
+CREATE MACRO ifelse(a, b, c) AS CASE WHEN a THEN b ELSE c END;
+-- create a macro that adds two expressions (a and b)
+CREATE MACRO add(a, b) AS a + b;
+-- create a macro that does a subquery
+CREATE MACRO one() AS (SELECT 1);
+-- create a macro with a common table expression
+-- (parameter names get priority over column names: disambiguate using the table name)
+CREATE MACRO plus_one(a) AS (WITH cte AS (SELECT 1 AS a) SELECT cte.a + a);
+-- macro's are schema-dependent, and have an alias: FUNCTION
+CREATE FUNCTION main.myavg(x) AS SUM(x) / COUNT(x);
+```
+
+### Syntax
+<div id="rrdiagram"></div>
--- a/docs/sql/statements/create_macro.md
+++ b/docs/sql/statements/create_macro.md
@@ -18,9 +18,13 @@ CREATE MACRO one() AS (SELECT 1);
 -- create a macro with a common table expression
 -- (parameter names get priority over column names: disambiguate using the table name)
 CREATE MACRO plus_one(a) AS (WITH cte AS (SELECT 1 AS a) SELECT cte.a + a);
+-- create a macro with default constant parameters
+CREATE MACRO plus_default(a, b=5) AS a + b;
 -- macro's are schema-dependent, and have an alias: FUNCTION
 CREATE FUNCTION main.myavg(x) AS SUM(x) / COUNT(x);
 ```
 
 ### Syntax
 <div id="rrdiagram"></div>
+
+Macro's are bound at creation time to ensure that they are valid. Macro's can still fail when called when invalid arguments are supplied. Internally, macro's are stored as a parsed expression. When they are called, the macro is expanded (i.e. replaced with the stored parsed expression), and the parameters within the expanded expression are replaced with the supplied arguments.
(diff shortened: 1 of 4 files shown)


js/statements/createmacro.js
```diff
@@ -0,0 +1,29 @@
...
+function GenerateCreateMacro(options = {}) {
+	return Diagram([
+		AutomaticStack([
+			Keyword("CREATE"),
+			Choice(0, [
+				Keyword("MACRO"),
+				Keyword("FUNCTION"),
+				]
+			),
+			Keyword("AS"),
```
Mytherin, line 11: This seems to miss the parameter list and the name of the function, no?
Probably want to add something like:

```js
Expression("macro-name"),
Keyword("("),
OneOrMore(Sequence([
  Expression("param-name"),
  Optional(Sequence([
    Keyword("="),
    Expression("default-value")
  ], "skip"),
]), Keyword(",")),
Keyword(")")
```

docs/sql/statements/create_macro.md
```diff
@@ -0,0 +1,26 @@
...
+CREATE MACRO one() AS (SELECT 1);
+-- create a macro with a common table expression
+-- (parameter names get priority over column names: disambiguate using the table name)
+CREATE MACRO plus_one(a) AS (WITH cte AS (SELECT 1 AS a) SELECT cte.a + a);
+-- macro's are schema-dependent, and have an alias: FUNCTION
+CREATE FUNCTION main.myavg(x) AS SUM(x) / COUNT(x);
+```
+
+### Syntax
+<div id="rrdiagram"></div>
```
Mytherin, line 26: Perhaps add an example with default values, and a short description of how the macros are handled (i.e. they get expanded when used in the query)?

Mytherin (commented):
Thanks! Looks great, some minor comments:

js/statements/createmacro.js
lnkuiper: Good point, I was too busy trying to figure out how the railroad works that I forgot the function name completely

docs/sql/statements/create_macro.md
lnkuiper: Added a description of how macro's are handled, and added MACRO to drop.js

js/statements/createmacro.js
lnkuiper: I did something like this, but where normal params must come before default params

docs/sql/statements/create_macro.md
```diff
@@ -0,0 +1,30 @@
...
+-- create a macro with default constant parameters
+CREATE MACRO plus_default(a, b=5) AS a + b;
+-- macro's are schema-dependent, and have an alias: FUNCTION
+CREATE FUNCTION main.myavg(x) AS SUM(x) / COUNT(x);
+```
+
+### Syntax
+<div id="rrdiagram"></div>
+
+Macro's are bound at creation time to ensure that they are valid. Macro's can still fail when called when invalid arguments are supplied. Internally, macro's are stored as a parsed expression. When they are called, the macro is expanded (i.e. replaced with the stored parsed expression), and the parameters within the expanded expression are replaced with the supplied arguments.
```
Mytherin, line 30: This seems a bit too in-depth for users that have no idea what a macro even is :)

Perhaps start off with their purpose, e.g.: "Macros allow you to create shortcuts for combinations of expressions.". I would stay away from terms such as parsed expression, since that is a bit internal-specific.

Also some examples that explain the behavior might help people understand better, e.g.:

```sql
-- failure! cannot find column "b"
create function add(a) as a+b;
-- this works
create function add(a,b) as a+b;
-- error! cannot bind +(VARCHAR, INTEGER)
select add('hello', 3);
-- success!
select add(1, 2);
-- 3
```

And add the default stuff we talked about :)

js/statements/createmacro.js
```diff
@@ -0,0 +1,48 @@
...
+				]
+			),
+			Optional(Sequence([
+				Expression("schema-name"),
+				Keyword(".")
+			]), "skip"),
+			Expression("macro-name"),
+			Keyword("("),
+			Optional(Sequence([
+				OneOrMore(Sequence([
```
Mytherin, line 18: I think this should be ZeroOrMore, rather than OneOrMore, no?

js/statements/drop.js
```diff
@@ -6,6 +6,7 @@ function GenerateDrop(options = {}) {
 			Keyword("DROP"),
 			Choice(3, [
 				Keyword("INDEX"),
+				Keyword("MACRO"),
```
Mytherin, line 9: And function :)

js/statements/createmacro.js
lnkuiper: Didn't know about ZeroOrMore! I guess Optional + OneOrMore is a bit hacky. Done.

js/statements/drop.js
lnkuiper: Done

docs/sql/statements/create_macro.md
lnkuiper: Done, added some more examples, without going too much into the details

Mytherin:
Looks great, thanks!

PR #21, opened 2021-01-01
Add numeric type aliases

szarnyasg:
While working on https://github.com/cwida/duckdb/pull/1256, I encountered a few inconsistencies in the docs. This PR amends them.

Even the tables in the PR omit a few typenames such as `INT16` (=`INT2`), `INT32` (=`INT4`), and `INT64` (=`INT8`). Should we include them or keep them omitted (as they can be confusing for newcomers).


Mytherin:
Thanks for the PR! We should probably fetch this from the DB somehow to avoid this from becoming outdated. Well, later :)

PR #23, opened 2021-01-06
Format text functions, add split functions

### LINKED ISSUE #22: Document list functions
List functions such as `unnest` and `string_split_regex` are currently undocumented. An example for their use:

```sql
D insert into emails values (1, 'a@b.com;c@d.com'), (2, 'e@f.com'), (3, '');
Error: Catalog Error: Table with name emails does not exist!
Did you mean "x"?
D create table emails(id int, addresses varchar);
D insert into emails values (1, 'a@b.com;c@d.com'), (2, 'e@f.com'), (3, '');
D select id, unnest(string_split_regex(addresses, ';')) from emails;
D select id, unnest(string_split_regex(addresses, ';')) as email from emails;
┌────┬─────────┐
│ id │  email  │
├────┼─────────┤
│ 1  │ a@b.com │
│ 1  │ c@d.com │
│ 2  │ e@f.com │
│ 3  │         │
└────┴─────────┘
```

szarnyasg:
Also fixes #22


szarnyasg:
The current output of the table is a bit of a mess.

image

It is fixed by this PR.

One remark: when I put the string
```
"    something"
```
in the Markdown code, the HTML output broke completely. So I figured it's better to use visible spaces (`␣`).

hannesmuehleisen:
Thanks!

PR #41, opened 2021-04-19
Add clarification that gzip compressed CSV files are supported

emres:
Add clarification that gzip compressed CSV files are supported. This is related to the following discussion: https://github.com/duckdb/duckdb/discussions/1645.


hannesmuehleisen:
Thanks

PR #44, opened 2021-04-26
Using clause

szarnyasg:
This PR fixes a small error (incorrect example macro) and documents the `USING` clause introduced in https://github.com/duckdb/duckdb/pull/1659.

This required moving the table list/subqueries used in `USING` and `FROM` to `railroad.js`.


Mytherin:
Thanks for the documentation! Looks good.

PR #48, opened 2021-05-05
Add header anchors

szarnyasg:
I added anchor headings using the trick / HTML template given in https://github.com/allejo/jekyll-anchor-headings.

It uses the [link emoji](https://emojipedia.org/link/) so its look is platform dependent.

image


hannesmuehleisen:
@Franz-Kafka can you have a look at this please?

szarnyasg:
A potential alternative: https://www.bryanbraun.com/anchorjs/

Mytherin:
Thanks for the contribution! This seems good and useful to me. I specifically like the ability to link to specific headers. I am personally not a big fan of how the anchors look, though, at least on MacOS.

Screenshot 2021-05-18 at 14 42 20

Rather than have a separate icon, perhaps we can just make the headers themselves clickable with this behavior?

szarnyasg:
Indeed, they don't look that good and they are emojis so they vary a lot between platforms. So we should either:

1. Use links as you suggested. I implemented this, you can take a look with:
```
cd /tmp && git clone git@github.com:szarnyasg/duckdb-web --branch anchors --depth 1 && cd duckdb-web && jekyll serve
```

2. Use an image file, e.g. based on the examples here: https://www.bryanbraun.com/anchorjs/#examples

Mytherin:
The clickable links look good to me. Thanks for making the changes!

PR #51, opened 2021-05-17
Restructure page on parquet files

lorenzwalthert:
As per https://github.com/duckdb/duckdb/pull/936#issuecomment-841608805. I figured that multi-file reading is actually documented, but hard to find for me (because after reading single files, it goes on with non-reading topics). 

* `Reading` is now named `Single-File Reads` instead of no title (as a hint that there is multi-file reading to come further down the page), 
* `Querying Parquet Files` was renamed to `Inserts and Views`.
* Re-ordered the paragraphs such that all content regarding reading shows up first, then view and inserts, then all writes. 

I admit this re-shuffling is opinionated to some degree.

### DIFF
--- a/docs/data/parquet.md
+++ b/docs/data/parquet.md
@@ -2,7 +2,11 @@
 layout: docu
 title: Parquet
 ---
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks for the contribution! I think the reshuffling indeed makes this more clear.

PR #59, opened 2021-06-17
Document aliases of aggregate functions

szarnyasg:
@pdet I was looking into the `max_by` function. I wanted to extend the documentation as well but I'm uncertain about whether `max_by` is supposed to be an exact alias for `arg_max` or it is a statistical function that should always return floating-point values? (Currently, it returns floating-point values.)

```sql
create table args (a integer, b integer);
insert into args values (1,1), (2,2), (8,8), (10,10);
select arg_min(a,b), arg_max(a,b) from args;
select min_by(a,b), max_by(a,b) from args;
```

```console
D create table args (a integer, b integer);
D insert into args values (1,1), (2,2), (8,8), (10,10);
D select arg_min(a,b), arg_max(a,b) from args;
┌───────────────┬───────────────┐
│ arg_min(a, b) │ arg_max(a, b) │
├───────────────┼───────────────┤
│ 1             │ 10            │
└───────────────┴───────────────┘
D select min_by(a,b), max_by(a,b) from args;
┌──────────────┬──────────────┐
│ min_by(a, b) │ max_by(a, b) │
├──────────────┼──────────────┤
│ 1.0          │ 10.0         │
└──────────────┴──────────────┘
```

### DIFF
--- a/docs/sql/aggregates.md
+++ b/docs/sql/aggregates.md
@@ -13,10 +13,10 @@ When the `DISTINCT` clause is provided, only distinct values are considered in t
 ## General Aggregate Functions
 The table below shows the available general aggregate functions.
 
-| Function | Description | Example | Alias |
+| Function | Description | Example | Alias(es) |
 |:---|:---|:---|:---|
-| `arg_max(arg,val)` |Calculates the arg value for a maximum val value. | `arg_max(A,B)` | `max_by(A,b)` |
-| `arg_min(arg,val)` |Calculates the arg value for a minimum val value. | `arg_min(A,B)` | `min_by(A,B)` |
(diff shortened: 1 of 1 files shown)


pdet:
I think It's supposed to be the same type of a, in that case, an integer.

I'll submit a fix for it soon! thanks for the catch! :-)

szarnyasg:
Cool, thanks for clarifying. Then this PR is good for review, @Mytherin

Mytherin:
The PR looks good to me.

@pdet while you are at it, could you add the `argMin` and `argMax` aliases to match Clickhouse's naming?

pdet:
@Mytherin Clickhouse uses argMin and argMax, no?
https://clickhouse.tech/docs/en/sql-reference/aggregate-functions/reference/argmin/

pdet:
Ah, without the _

Mytherin:
Thanks!

PR #76, opened 2021-09-11
Scalar, range and date range functions

szarnyasg:
@Mytherin I documented functions introduced in https://github.com/duckdb/duckdb/issues/193


Mytherin:
Thanks! Looks great.

PR #102, opened 2021-11-17
Fix some datetime functions examples, formatting

szarnyasg:
@Mytherin I fixed some example outputs of the datetime functions and added some formatting:

GitHub's word-diff is not great so here's a list of the main changes:

* `age` now returns a timestamp and not a date (!)
* `mons` is now `months`.
* `TIMESTAMP '1992-03-27' - TIMESTAMP '1992-03-22'`: previously was `INTERVAL 5 DAY`, changed to `5 days`
* `dayofyear(TIMESTAMP '1992-03-22')`: previously was `81`, changed to `82`
* `monthname(TIMESTAMP '1992-09-20')`: previously was `March`, changed to `September`

There were some missing/extra apostrophes and `generate_series` was mixed up with `range`.

### DIFF
--- a/docs/sql/functions/timestamp.md
+++ b/docs/sql/functions/timestamp.md
@@ -11,55 +11,55 @@ The table below shows the available mathematical operators for `TIMESTAMP` types
 
 | Operator | Description | Example | Result |
 |:---|:---|:---|:---|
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks!

PR #103, opened 2021-11-17
String concat and create_fts_index

szarnyasg:
@Mytherin Two changes:

1. string concat operators looked off as the escape chars were rendered in the HTML:

image

2. the defaults of create_fts_index were not in sync. (cc @lnkuiper)


Mytherin:
Thanks!

PR #124, opened 2021-12-04
Add CREATE MACRO example and document generate_subscripts

szarnyasg:
* Add another `CREATE MACRO` example
* Document the generate_subscripts macro introduced in https://github.com/duckdb/duckdb/pull/2576/


Mytherin:
Thanks!

PR #132, opened 2021-12-15
Rework Python and R connection examples

szarnyasg:
The existing examples did not show how to connect to a file and were somewhat confusing (as using a memory database implies read-only = false so it was unnecessary).


Mytherin:
Thanks!

PR #135, opened 2021-12-16
Split nested docs

Alex-Monahan:
I split the nested docs into 3 separate docs. The summary table and comparison/grouping/function information is all in each document (just at the bottom). 

Splitting the functions document is for another PR!


Alex-Monahan:
I believe that this is ready to be reviewed and then merged! Thanks as always!

Mytherin (commented):
Thanks for the updates! Looks great. Some comments:

Mytherin:
Thanks! Looks great.

PR #150, opened 2022-01-07
Create scala.md

jim-tommaney:
Does it make sense to call out usage in Scala?

### DIFF
--- a/docs/api/scala.md
+++ b/docs/api/scala.md
@@ -1,0 +1,82 @@
+---
+layout: docu
+title: Scala JDBC API
+selected: Client APIs
+---
+## Installation
+The DuckDB Java JDBC API can be used in Scala and can be installed from [Maven Central](https://search.maven.org/artifact/org.duckdb/duckdb_jdbc). Please see the [installation page](/docs/installation?environment=java) for details.
+
+## Basic API Usage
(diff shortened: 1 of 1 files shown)


Mytherin:
Looks good to me, but I don't have particularly strong opinions on anything related to Java. @hannesmuehleisen what do you think?

Alex-Monahan:
Thanks for putting this together @jim-tommaney! I found a few small typos and just had a small question as well.

> Scala uses DuckDB's JDBC API implements the main parts of the standard Java Database Connectivity (JDBC) API, version 4.0. 
changes to:

> Scala uses DuckDB's JDBC API **which** implements the main parts of the standard Java Database Connectivity (JDBC) API, version 4.0. 

> you can use the custom duplicate() method like so:

Would you be able to describe what the duplicate function is doing to allow multiple read-write connections? That feature isn't possible through other DuckDB API's as far as I know, so it sounds very beneficial!

jim-tommaney:
@Alex-Monahan 
Alex, 

I used the java doc as a template for scala, with needed net changes.
https://duckdb.org/docs/api/java


This came from the Java doc, and afaik should still be valid:

If you would like to create multiple read-write connections to the same database file or the same in-memory database instance, you can use the custom duplicate() method like so:

Connection conn2 = ((DuckDBConnection) conn).duplicate();

Alex-Monahan:
Thank you for the explanation! It would be fine with me to merge now!

Mytherin:
Thanks!

PR #154, opened 2022-01-26
Hawkfish isoyear and Interval Data Type

Alex-Monahan:


_data/menu_docs_028.json
```diff
@@ -162,6 +162,10 @@
                   "page": "Timestamp",
                   "url": "timestamp"
                 },
+                {
+                  "page": "Interval",
```
Mytherin, line 166: One problem I see here is that the page is added to the historic menu pages (`menu_docs_028`, etc) but the page itself is only added to the current docs. The page should only be added to the current menu pages (`menu_docs_current.json`) in this case.

PR #158, opened 2022-01-29
Add examples for recursive CTEs and document nested CTEs

### LINKED ISSUE #106: Improve documentation on CTEs
See https://github.com/duckdb/duckdb/issues/2551
I'm "self-assigning" this since I'm working with recursive CTEs quite a lot.

* [x] Add hierarchical CTE example
* [x] Add generic graph query example
* [x] Check whether nested CTEs work (if not, document it)

szarnyasg:
Fixes #106

### DIFF
--- a/docs/sql/query_syntax/with.md
+++ b/docs/sql/query_syntax/with.md
@@ -5,20 +5,166 @@ selected: Documentation/SQL/Query Syntax/With
 expanded: SQL
 railroad: query_syntax/with.js
 ---
-The `WITH` clause allows you to specify common table expressions (CTEs). Regular (non-recursive) common-table-expressions are essentially views that are limited in scope to a particular query. CTEs can reference each-other.
+The `WITH` clause allows you to specify common table expressions (CTEs). Regular (non-recursive) common-table-expressions are essentially views that are limited in scope to a particular query. CTEs can reference each-other and can be nested.
 
-### Examples
+### Basic CTE examples
 
 ```sql
 -- create a CTE called "cte" and use it in the main query
 WITH cte AS (SELECT 42)
 SELECT * FROM cte;
+```
+```
+┌────┐
+│ 42 │
+├────┤
+│ 42 │
+└────┘
+```
+```sql
 -- create two CTEs, where the second CTE references the first CTE
 WITH cte AS (SELECT 42 AS i),
      cte2 AS (SELECT i*100 FROM cte)
 SELECT * FROM cte2;
 ```
+```
+┌─────────┐
+│ i * 100 │
+├─────────┤
+│ 4200    │
+└─────────┘
+```
 
+### Recursive CTE examples
+
+#### Tree traversal
+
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

docs/sql/query_syntax/with.md
```diff
@@ -5,20 +5,166 @@ selected: Documentation/SQL/Query Syntax/With
...
+```
+
+`WITH RECURSIVE` can be used to find all unweighted shortest paths between two nodes. The following query returns all unweighted shortest paths between person 1 and person 3:
+
+```sql
+WITH RECURSIVE paths(startPerson, endPerson, path, endPersonReached) AS (
+   SELECT person1id AS startPerson,
+          person2id AS endPerson,
+          [person1id, person2id]::bigint[] AS path,
+          max(CASE WHEN person2id = 3 THEN true ELSE false END) OVER () AS endPersonReached
```
hawkfish, line 142: No ordering is specified here, so I worry that the data could get reordered if `paths` gets large. It doesn't matter for this small example, but if someone were to use it for a much larger problem, could this be an issue?

hawkfish (commented):
Sorry this is late, but I did have one windowing concern.

docs/sql/query_syntax/with.md
szarnyasg: Good point... I'm not sure.

My thinking is that `max()` is associative, so `endPersonReached` is `true` if *any* of the `person2id`s in the subquery results is 3 (indicating that the graph traversal has reached the end person in that step).

hawkfish: > My thinking is that `max()` is associative, so `endPersonReached` is `true` if _any_ of the `person2id`s in the subquery results is 3 (indicating that the graph traversal has reached the end person in that step).

It is, but the default frame bounds goes from the start of the partition to the current row, so it will only be set for rows where there is a 3 there or before, which might not be correct if the data was reordered somehow.

szarnyasg: Right. So one way to keep them in sync is to sort it with total ordering. For example, the first subquery would become:
```sql
SELECT person1id AS startPerson,
        person2id AS endPerson,
        [person1id, person2id]::bigint[] AS path,
        max(CASE WHEN person2id = 3 THEN true ELSE false END) OVER (ORDER BY person1id, person2id) AS endPersonReached
    FROM knows
WHERE person1id = 1
ORDER BY person1id, person2id
```
This may be a performance hit – I have to microbenchmark it – but if it helps avoid a (very infrequent) Heisenbug, I'm all for it.

hawkfish: I don't think the ordering is needed, just expand the frame to be the entire table? `ORDER (UNBOUNDED PRECEDING TO UNBOUNDED FOLLOWING)` would just set the flag to be true if any path in the table has reached the end.

szarnyasg: Excellent, thanks! Have not used window functions much, so this helps a lot. I sent a follow-up PR: #161

PR #160, opened 2022-01-31
list_slice vs array. array_pre/append. Reorder fns

Alex-Monahan:
Now all of the list examples use list_ functions. The array_ functions are still included on the functions page, but are not in any examples on the data types page. The list functions were also reordered for clarity.


hawkfish (approved):
We should take a pass over the `STRUCT` and other functions. It gets a bit ratty the further down you go.  I was also wondering if it might make sense to split this page up as it is getting rather long.

Mytherin:
Thanks!

PR #161, opened 2022-01-31
Fix window function range in graph traversal example

szarnyasg:
Ensure that the 'end node' of the graph traversal is included even if the rows are reordered.
For details, see the discussion in issue #158.


hawkfish (approved):

hawkfish (approved):

Mytherin:
Thanks!

PR #163, opened 2022-02-10
Document the MD5 hash function and sort text function table

szarnyasg:
It's currently the only cryptographic function so I put it under string (char) functions.


hawkfish (commented):
Thanks - can we just move it up to be alphabetical?

szarnyasg:
@hawkfish sure. I sorted the table as well.

PR #168, opened 2022-02-17
Document disable_optimizer pragma

szarnyasg:


hawkfish (commented):
Thanks for this - I need it yesterday and had to read the code to find it!

szarnyasg:
Thanks, now both are included.

hawkfish (approved):

PR #171, opened 2022-02-22
Glob added to pattern matching

Alex-Monahan:


Alex-Monahan:
I also added an ILIKE example with ESCAPE. Thanks for adding that in @Mytherin! 

This is ready to review.

PR #176, opened 2022-03-02
Added list_contains, list_position, unnest(null), duckdb_functions

taniabogatsch:
Also, this contains all aliases for `list_contains` and `list_position`. I am not sure that the formatting is correct for the section about the `duckdb_functions`. Please check.


Mytherin:
Thanks!

PR #179, opened 2022-03-04
Document JSON extension

lnkuiper:
And move extensions to a separate tab in the documentation menu (just FTS and JSON for now)


lnkuiper:
Ah just seen the TODO in there, fixing

Alex-Monahan:
Hey @lnkuiper, I had a few small questions on the JSON docs. First of all, a HUGE thank you for implementing them!! 

Could you add a bit about how to install and load the JSON extension? I think it's just those 2 commands, but is it any different in the different clients, like Python, etc.? If it is different by client, maybe we should just have a separate "Extension Install" page.

Are we planning to include the JSON extension automatically with some of the bundles? I think it would be great to include in Python, Node and WASM since they all make heavy use of JSON. If so, could you add a quick note about that near the top so folks don't try to install them when it isn't needed?

Thanks!

lnkuiper:
@Alex-Monahan Someone added an `index.md` page for `/docs/extensions` which describes how to install the extension. Although I agree this should be included in the json docs explicitly as well.

I think the JSON extension currently is only automatically bundled in the Linux and OSX CLI's, but I am not quite sure, it's a bit tricky for me to read the deployment pipeline. Maybe @hannes can shed some light on this?

PR #182, opened 2022-03-10
Qualify clause docs and railroad diagrams

Alex-Monahan:
This also contains some changes to the select page's order and the summary of the SQL clauses.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -299,3 +299,7 @@ Gemfile.lock
 _site/
 
 .jekyll-cache/
+duckdb_functions_join_docs_with_notes.tab.txt
+duckdb_functions_join_docs.tab
+duckdb_functions.tab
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -299,3 +299,7 @@ Gemfile.lock
 _site/
 
 .jekyll-cache/
+duckdb_functions_join_docs_with_notes.tab.txt
```
hawkfish, line 302: Did you mean to do this?

hawkfish (approved):

.gitignore
Alex-Monahan: Yes, there are a few files that are generated when I parsed the existing docs markdown for our scalar functions. It would be good to have them in the gitignore once we have that parsing set up.

Mytherin:
Thanks! Looks great.

PR #184, opened 2022-03-10
ODBC Doc

tiagokepe:
Hey all,

I updated the ODBC documentation. It was spited off into separate files for Linux and Windows setup.


docs/api/odbc.md
```diff
@@ -0,0 +1,116 @@
...
+The ODBC driver is a DBMS implementation of the ODBC API, which handles all the internals of that DBMS.
+
+The DM maps user application calls of ODBC functions to the correct ODBC driver that performs the specified function and returns the proper values.
+
+## DuckDB ODBC Assets
+
+DuckDB supports the ODBC version 3.0 according to the [Core Interface Conformance](https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/core-interface-conformance?view=sql-server-ver15). 
+
+We release the ODBC driver as assets for Linux and Windows.
+Users can download them from the [Latest Release of DuckBD](https://github.com/duckdb/duckdb/releases).
```
Mytherin, line 23: Perhaps we should add the ODBC installation also to [the installation page](https://duckdb.org/docs/installation/) under a separate ODBC header in Environment?

docs/api/odbc.md
```diff
@@ -0,0 +1,116 @@
...
+
+**[DuckDB]**: between the brackets is a DSN for the DuckDB.
+
+**Driver**: it describes the driver's name, and other configurations will be placed in the **.odbcinst.ini**.
+
+**Database**: it describes the database name used by DuckDB, and it can also be a file path to a `.db` in the system.
+
+### The ".odbcinst.ini" File
+
+The **.odbcinst.ini** contains general configurations for the ODBC installed drivers in the system.
```
Mytherin, line 53: Perhaps we could take some notes from the [Snowflake Linux ODBC setup instructions](https://docs.snowflake.com/en/user-guide/odbc-linux.html#label-odbc-linux-install-tgz). For example, they seem to ship a `unixodbc_setup.sh` script along with the driver. This might simplify the general setup. 

Some instructions on how to get this working on a clean install of a popular OS, e.g. Ubuntu, might also be useful. For example adding which packages need to be installed (`sudo apt-get install unixodbc`) and generally including instructions that can just be copy pasted into the terminal would simplify the installation process.

docs/api/odbc.md
```diff
@@ -0,0 +1,116 @@
...
+**TraceFile**: the absolute system file path for the ODBC trace file.
+
+
+**[DuckDB Driver]**: the section of the DuckDB installed driver.
+
+**Driver**: the absolute system file path of the DuckDB driver. 
+
+
+
+## Windows Setup
```
Mytherin, line 81: Perhaps we can split off the ODBC setup into separate files similar to the Snowflake pages, so we have e.g. `ODBC/Linux Setup`, `ODBC/Windows Setup`, etc

docs/api/odbc.md
```diff
@@ -0,0 +1,116 @@
...
+For example, the DSN registry for DuckDB would look like this:
+
+![`HKLM->SOFTWARE->ODBC->ODBC.INI->DuckDB`](/images/blog/odbc/odbc_ini-registry-entry.png)
+
+
+The `ODBCINST.INI` contains one entry for each ODBC driver and other keys predefined for [Windows ODBC configuration](https://docs.microsoft.com/en-us/sql/odbc/reference/install/registry-entries-for-odbc-components?view=sql-server-ver15).
+
+### ODBC Windows Installer
+
+To simplify and facilitate the installation process, we provide the ODBC installer (`odbc_install.exe`) with the Windows asset.
```
Mytherin, line 100: I would start off with the installer in the Windows setup section, as that seems to me like the easiest (and recommended?) way of installing ODBC for Windows users. The rest can be provided as more details.

Is it also possible to double click on the executable, or is running the command through the command prompt required? Windows users in general are less used to shells so double clicking would be easier.

Mytherin (commented):
Thanks a lot for the PR! Looks great. Some comments:

docs/api/odbc.md
tiagokepe: > Perhaps we should add the ODBC installation also to [the installation page](https://duckdb.org/docs/installation/) under a separate ODBC header in Environment?

For sure, this is the next step.

docs/api/odbc.md
tiagokepe: Yeah, I didn't see that guide before, it seems very neat (a good guide to follow).

docs/api/odbc.md
Mytherin: That sounds fine when launched via double click, but I would avoid launching message boxes when installing through a command line interface.

tiagokepe: I see, we have a /CI option already that disables message boxes.

Alex-Monahan:
I just had a read through it and it looks good! There are one or two small typos (I saw a DuckBD in there... ;-) ), but I can make those edits after this gets merged if that's the easiest way.

I had a couple of small thoughts. Do you think you could add a sentence or two about where ODBC can be used in the introduction section? That will help folks understand when they should use ODBC vs. another method like JDBC. 

Maybe it would belong in a blog post, but a few screenshots to get DuckDB up and running in a business intelligence tool (Excel, Power BI, Tableau, etc.) would be really helpful to folks that are just getting started!

docs/api/odbc/windows.md
```diff
@@ -0,0 +1,87 @@
...
+
+**duckdb_odbc.dll**: the DuckDB driver compiled for Windows.
+
+**duckdb_odbc_setup.dll**: a setup DLL used by the Windows ODBC Data Source Administrator tool.
+
+**odbc_install.exe**: a installation script to aid the configuration on Windows.
+
+# Step 2: Extracting ODBC artifacts
+
+Unsing PowerShell, run unzip to extract the files:
```
rvasil, line 26: Typo in "Unsing" + `unzip` actually is not a powershell command. Maybe just instruct user to unzip .zip file without telling how... 
I'm not sure if my experience in installing the driver is correct and intended: odbc_install.exe didn't copy the dll files to Windows\system32 directory as is my experience e.g. with SQLite driver. It simply pointed registry entries to the location where dlls were extracted. If this is the intended setup, the docs should explicitly instruct the user that the location of extracted files is meant to be permanent and files should not be moved in order the driver keeps working.

tiagokepe: > Typo in "Unsing" + `unzip` actually is not a powershell command. Maybe just instruct user to unzip .zip file without telling how... I'm not sure if my experience in installing the driver is correct and intended: odbc_install.exe didn't copy the dll files to Windows\system32 directory as is my experience e.g. with SQLite driver. It simply pointed registry entries to the location where dlls were extracted. If this is the intended setup, the docs should explicitly instruct the user that the location of extracted files is meant to be permanent and files should not be moved in order the driver keeps working. Thanks.

@rvasil thanks for your comments.
You are right about the location of the extracted files being permanent.
The first version of the `odbc_install.exe` only focuses on setting up the ODBC registry entries. A new version should copy the .dll files to `Windows\system32` directory.
I will clarify that in the documentation.

docs/installation/index.html
```diff
@@ -102,130 +105,165 @@ <h3>Usage Example</h3>
...
-	
-	
-	
+
+	<div class="master odbc binary macos">
+		<a href="https://github.com/duckdb/duckdb/releases/download/master-builds/libduckdb-osx-amd64.zip" target="_blank">https://github.com/duckdb/duckdb/releases/download/master-builds/libduckdb-osx-amd64.zip</a>
+	</div>
+
+	<div class="master odbc binary linux">
+		Linux 64-bit: <a href="https://github.com/duckdb/duckdb/releases/download/master-builds/libduckdb-linux-amd64.zip" target="_blank">https://github.com/duckdb/duckdb/releases/download/master-builds/libduckdb-linux-amd64.zip</a><br/><br/>
```
Mytherin, line 229: These files don't appear to be there, perhaps it would be better to leave these empty if there is no possibility of downloading master binaries for the ODBC build?

Mytherin (commented):
Thanks for all the updates! Looks a lot better now. One more comment regarding the installation:

docs/installation/index.html
tiagokepe: Removed the links from C++ and ODBC tabs.

Mytherin:
Thanks for all the changes! This seems ready to merge now.

PR #187, opened 2022-03-14
Fix numerical ranges for hugeint and bigint in numeric documentation page

ajzo90:
for docs/sql/data_types/numeric


hawkfish (approved):

Mytherin:
Thanks for the PR! The original hugeint range was correct, however. We don't allow the smallest hugeint number right now (-170141183460469231731687303715884105728) because our internal operators on hugeint rely on negating numbers which is not possible with this number.

Mytherin:
Thanks

PR #195, opened 2022-03-30
Add arrow to Python overview

Alex-Monahan:


Alex-Monahan:
I am also in the process of adding 3 more guides for Arrow: SQL on Arrow, Import from Arrow, Export to Arrow (following the Pandas guides structure). I'm going to try to cover all the types of Arrow objects we can read/write there (Tables, datasets, recordbatches, and any others I'm missing?).

pdet:
@Alex-Monahan We also support Scanners.

docs/api/python.md
```diff
@@ -60,28 +60,46 @@ print(con.fetchall())
 > Do *not* use `executemany` to insert large amounts of data into DuckDB. See below for better options.
 
 ## Efficient Transfer
-Transferring large datasets to and from DuckDB uses a separate API built around [NumPy](https://numpy.org) and [Pandas](https://pandas.pydata.org). This API works with entire columns of data instead of scalar values and is therefore far more efficient. 
+Transferring large datasets to and from DuckDB uses a separate API built around [NumPy](https://numpy.org) and [Pandas](https://pandas.pydata.org), or [Apache Arrow](https://arrow.apache.org/). This API works with entire columns of data instead of scalar values and is therefore far more efficient. 
 
-DuckDB supports "registering" a Pandas data frame as a virtual table, comparable to a SQL `VIEW`. Below is an example:
+By default, DuckDB will automatically be able to query any variable that is a Pandas or Arrow data frame by name. DuckDB also supports "registering" a data frame as a virtual table, comparable to a SQL `VIEW`. This is useful when querying a data frame that is stroed in another way (as a class variable, or a value in a dictionary) Below is a Pandas example:
```
pdet, line 65: I'm not 100% sure we can scan any arrow variable, for sure we can do datasets (memory and from disk), tables, recordbatchreader and scanners.
stroed ==> stored

docs/api/python.md
```diff
@@ -60,28 +60,46 @@ print(con.fetchall())
...
+```
+> DuckDB keeps a reference to the Pandas or Arrow data frame after registration. This prevents the data frame from being garbage-collected by the Python runtime. The reference is cleared when the connection is closed, but can also be cleared manually using the `unregister()` method.
 
 When retrieving the data from DuckDB back into Python, the standard method of calling `fetchall()` is inefficient as individual Python objects need to be created for every value in the result set. When retrieving a lot of data, this can become very slow.
 
-In DuckDB, there are two additional methods that can be used to efficiently retrieve data: `fetchnumpy()` and `fetchdf()`. `fetchnumpy()` fetches the data as a dictionary of `NumPy` arrays. `fetchdf()` fetches the data as a Pandas DataFrame.
+In DuckDB, there are two additional methods that can be used to efficiently retrieve data: `fetchnumpy()`, `fetchdf()`, and `fetch_arrow_table()`. `fetchnumpy()` fetches the data as a dictionary of `NumPy` arrays. `fetchdf()` fetches the data as a Pandas DataFrame. `fetch_arrow_table()` fetches the data as an [Arrow table](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html).
```
pdet, line 102: we also have fetch_df_chunk, and fetch_record_batch

Alex-Monahan:
Great feedback! I will likely just list all of the Arrow types we can read here, but refer to a guide that will have an example for each type!

Good call out on the different ways to fetch also - I can mention that here and refer to the upcoming guide as well.

Alex-Monahan:
Ok, I made some changes - let me know if that looks better!

Alex-Monahan:
Alright, now it's ready for another review. Thanks for the feedback - it's a lot more thorough this way!

docs/api/python.md
```diff
@@ -60,33 +60,67 @@ print(con.fetchall())
...
+* `df()` fetches the data as a Pandas DataFrame
+* `fetchdf()` is an alias of `df()`
+* `fetch_df()` is an alias of `df()`
+* `fetch_df_chunk(vector_multiple)` fetches a portion of the results into a DataFrame. The number of rows returned in each chunk is the vector size (1024 by default) * vector_multiple (1 by default).
 
-Below is an example of using this functionality:
+### Apache Arrow
+* `arrow()` fetches the data as an [Arrow table](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html)
+* `fetch_arrow_table()` is an alias of `arrow()`
+* `fetch_record_batch(chunk_size)` returns an [Arrow record batch reader](https://arrow.apache.org/docs/python/generated/pyarrow.ipc.RecordBatchStreamReader.html) with `chunk_size` rows per batch
```
pdet, line 117: I just checked this one is actually ```fetch_arrow_reader(chunk_size)```

Otherwise I think it looks pretty good!

Alex-Monahan: I'm on version 0.3.3-dev1409 and it looks like fetch_record_batch is working and fetch_arrow_reader is failing with:
```python
AttributeError: 'duckdb.DuckDBPyConnection' object has no attribute 'fetch_arrow_reader'
```

Mytherin:
@pdet is this ready to merge?

pdet:
Yes, looks great! :)

PR #196, opened 2022-03-31
Rework WITH RECURSIVE graph search examples. Fixes #189

### LINKED ISSUE #189: Revise and extend WITH RECURSIVE examples
A while ago, I contributed the examples for `WITH RECURSIVE` (#158 #188)
https://duckdb.org/docs/sql/query_syntax/with

I have since found that these queries are quite inefficient, especially the bidirectional search if there is no path between the start and end nodes.

It would also want to add a proper bidirectional search algorithm, where the two BFS frontiers are advanced simultaneously from the start and the end node.

I don't have time to tackle this now, so I'm opening this issue as a reminder and will get back to it later this spring.

szarnyasg:


szarnyasg:
@Alex-Monahan thanks for the feedback. I added more comments, both in the text explaining what we do and in the queries as comments.

Mytherin:
Thanks! Looks good to me.

PR #197, opened 2022-04-04
Python Arrow How-To-Guides

Alex-Monahan:
Hi Folks!

Let me know what you think! I followed the Pandas guides as an example. 

I didn't even know about Arrow Scanners, so that's worth knowing! Quick random question for my own learning: any idea on the speed of using an Arrow Scanner vs. just querying a dataset directly with DuckDB using a where clause? Arrow mentions that they use Async IO, so I didn't know if that would have any impact.

### DIFF
--- a/docs/guides/python/export_arrow.md
+++ b/docs/guides/python/export_arrow.md
@@ -1,23 +1,48 @@
 ---
 layout: docu
-title: Export To Pandas
-selected: Export To Pandas
+title: Export To Apache Arrow
+selected: Export To Apache Arrow
 ---
 
-# How to export data to a Pandas DataFrame
+# How to export data to Apache Arrow
+All results of a query can be exported to an [Apache Arrow Table](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html) using the `arrow` function. Alternatively, results can be returned as a [RecordBatchReader](https://arrow.apache.org/docs/python/generated/pyarrow.ipc.RecordBatchStreamReader.html) using the `fetch_record_batch` function and results can be read one batch at a time.
 
-The result of a query can be converted to a Pandas DataFrame using the `df()` function.
-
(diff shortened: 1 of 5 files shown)


docs/guides/python/sql_on_arrow.md
```diff
@@ -0,0 +1,92 @@
+---
+layout: docu
+title: SQL on Apache Arrow
+selected: SQL on Apache Arrow
+---
+
+# How to execute SQL on Apache Arrow
+DuckDB can query multiple different types of Apache Arrow objects. 
+
```
pdet, line 9: For completion maybe also add an example with recordbatchreaders?

docs/guides/python/export_arrow.md
```diff
@@ -0,0 +1,48 @@
...
+
+# connect to an in-memory database
+con = duckdb.connect()
+
+my_arrow_table = pa.Table.from_pydict({'i':[1,2,3,4],
+                                       'j':["one", "two", "three", "four"]})
+
+# query the Apache Arrow Table "my_arrow_table" and return as an Arrow Table
+results = con.execute("SELECT * FROM my_arrow_table").arrow()
+```
```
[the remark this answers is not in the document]
Alex-Monahan: Thanks for the feedback- I've added a section for that in the latest commit!

docs/guides/python/sql_on_arrow.md
Alex-Monahan: I also added an example for this! I'm learning a fair bit about Arrow... they have a few different options for objects, huh? I'm sure you feel that more than I do...

Alex-Monahan:
I believe this is ready for another review @pdet! Thanks again!

Mytherin:
Thanks!

PR #198, opened 2022-04-05
Basic Relational API on Pandas guide

Alex-Monahan:
The first of what should at some point be several Relational API guides! I think it may make sense to do more of the API reference documentation before going too much further with additional guides though. I was just on a roll with guides and wanted to get something out there...

What are your thoughts, @pdet? I appreciate it!


pdet:
This looks nice as a quick summary of the relation API, but my main question here is how deep it should go? Because, of course, the RAPI goes way deeper than that. I think that ```Please see [these additional examples](https://github.com/duckdb/duckdb/blob/master/examples/python/duckdb-python.py), and [the available relational methods on the DuckDBPyRelation class](https://github.com/duckdb/duckdb/blob/master/tools/pythonpkg/duckdb-stubs/__init__.pyi).``` should become tutorials themselves.

Alex-Monahan:
> This looks nice as a quick summary of the relation API, but my main question here is how deep it should go? Because, of course, the RAPI goes way deeper than that. I think that `Please see [these additional examples](https://github.com/duckdb/duckdb/blob/master/examples/python/duckdb-python.py), and [the available relational methods on the DuckDBPyRelation class](https://github.com/duckdb/duckdb/blob/master/tools/pythonpkg/duckdb-stubs/__init__.pyi).` should become tutorials themselves.

You are correct - there is a lot more left to document! I was thinking that we may want to do some API reference docs before doing more tutorials since all of the various functions aren't really documented yet. I have a few more guides planned on other topics, but then I can dig in and document this in more depth. Does that sound ok?

Mause (approved):

docs/guides/python/relational_api_pandas.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: docu
+title: Relational API and Pandas
+selected: Relational API and Pandas
+---
+
+# How to use the Relational API to query Pandas
+
+DuckDB offers a relational API that can be used to chain together query operations. These are lazily evaluated so that DuckDB can optimize their execution. These operators can act on Pandas DataFrames, DuckDB tables or views (which can point to any underlying storage format that DuckDB can read, such as csv or parquet files, etc.). Here we show a simple example of reading from a Pandas DataFrame and returning to a DataFrame.
```
hawkfish, line 9: "returning a DataFrame"

Alex-Monahan: Fixed - good catch!

PR #203, opened 2022-04-19
Added timezonetz type to SQL types overview page

Loreith:
Just a tiny patch to an inconsistency in SQL documentation. Hope this is helpful and not just noise!


docs/sql/data_types/overview.md
```diff
@@ -23,6 +23,7 @@ The table below shows all the built-in general-purpose data types. The alternati
 | `SMALLINT` | `INT2`, `SHORT` | signed two-byte integer|
 | `TIME` | | time of day (no time zone) |
 | `TIMESTAMP` | `DATETIME` | combination of time and date |
+| `TIMESTAMP WITH TIME ZONE` | `TIMESTAMPTZ` | combination of time and date using timezone |
```
hawkfish, line 26: Good catch!  

The SQL standard is very confusing on this type and the time zone is not actually part of the stored data. Perhaps "that uses the current time zone"?

Mytherin:
Thanks!

PR #207, opened 2022-05-01
Add caveat for FTS index

patrickpang:
As a follow-up of https://github.com/duckdb/duckdb/issues/3543


hannes:
Thanks for adding the note, but don't you think immediately after the initial sentence is a bit early for this?

patrickpang:
I see. How about A separate caveat section in the end?

Alex-Monahan:
Looks fine to me!

patrickpang:
Thank you!

PR #221, opened 2022-05-13
Fix arrow variable name in comment

keckelt:
The comments in the [Import from Arrow](https://duckdb.org/docs/guides/python/import_arrow) guide referenced a `my_df` variable (as in the [pandas example](https://duckdb.org/docs/guides/python/import_pandas)), but here `my_arrow` is used.


Mytherin:
Thanks!

PR #225, opened 2022-05-31
Add more documentation for NULL values and the IN clause

Mytherin:


Alex-Monahan:
This looks great! I made a few small tweaks, otherwise we are good to merge.

PR #231, opened 2022-06-02
Generated column doc

Tishj:
This PR adds documentation for generated columns to the duckdb docs.

Syntax for it is added to the Syntax railroad diagram of `create_table`, this portion is made expandable under the name `generated-column`.  

This is what it looks like before it's expanded:
image
And this is after:
image

I've also added a brief description of generated columns to `create_table`. We might want to move this description to its own page later, but I felt this was the best place for it, since generated columns are only supported in `CREATE TABLE` statement for now.

PS:
I've had some issues with a bug where the VIRTUAL|GENERATED options were clipping through the options below, which almost magically disappeared, so please verify that this also works for you by running the site locally with jekyll

### DIFF
--- a/docs/sql/statements/create_table.md
+++ b/docs/sql/statements/create_table.md
@@ -58,7 +58,7 @@ CREATE TABLE IF NOT EXISTS t1(i INTEGER, j INTEGER);
 
 The `[type] [GENERATED ALWAYS] AS ( expr ) [VIRTUAL|STORED]` syntax will create a generated column. The data in this kind of column is generated from its expression, which can reference other (regular or generated) columns of the table. Since they are produced by calculations, these columns can not be inserted into directly.
 
-It is also possible to leave out the type when declaring a generated column. Since an expression has a return type, its type will be automatically derived from the expression.
+When possible, DuckDB will infer the type of the generated column based on the expression's return type. This often allows the type to be left out when declaring a generated column. 
   
 Generated columns come in two varieties: `VIRTUAL` and `STORED`.  
 The data of virtual generated columns is not stored on disk, instead it is computed from the expression every time the column is referenced (through a select statement).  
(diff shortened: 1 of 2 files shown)


Alex-Monahan:
Looks good! I reordered some sentences a bit just for clarity and I added a few examples. Would you mind double checking my examples? I haven't pulled down the code to test since it hasn't reached master just yet.

Tishj:
Nice! Yes that works, you could also leave off the `GENERATED ALWAYS` part in the example :)

Tishj:
I am not sure about the sentence: "Since an expression has a return type, its type will be automatically derived from the expression." This makes it feel like that will always happen regardless if you specify the type or not, which isn't the case.

Alex-Monahan:
> I am not sure about the sentence: "Since an expression has a return type, its type will be automatically derived from the expression." This makes it feel like that will always happen regardless if you specify the type or not, which isn't the case.

Ah, interesting! Is there a way to know whether or not this will succeed? Or should we just reword it to be more of a "DuckDB will attempt to automatically derive the type from the return type of the expression"?

Alex-Monahan:
Let me know if you think this is ready to merge!

Alex-Monahan:
> I missed your reaction earlier, it should always succeed.
> 
> Actually setting the type explicitly is how you could cause problems, your column might become unusable because the data you insert might not be able to be cast to the type of the generated column that references the column.
> 
> For example, a problematic sequence of actions:
> 
> ```sql
> D CREATE TABLE tbl (
> > col INTEGER,
> > gcol DATE as (col)
> > );
> D INSERT INTO tbl VALUES (5);
> Error: Constraint Error: Constraint Error: Incorrect DATE value for generated column "gcol"
> ```

Ah, now I understand. I will make some adjustments!

Alex-Monahan:
Looks good! Only tweak would be that varient should be variant!

PR #235, opened 2022-06-09
Document TRY_CAST

Mytherin:


docs/sql/expressions/cast.md
```diff
@@ -10,16 +10,23 @@ railroad: expressions/cast.js
...
+-- Conversion Error: Could not convert string 'hello' to INT32
+SELECT TRY_CAST('hello' AS INTEGER);
+-- NULL
 ```
 
 The exact behavior of the cast depends on the source and destination types. For example, when casting from `VARCHAR` to any other type, the string will be attempted to be converted.
 
 Not all casts are possible. For example, it is not possible to convert an `INTEGER` to a `DATE`. Casts may also throw errors when the cast could not be successfully performed. For example, trying trying to cast the string `'hello'` to an `INTEGER` will result in an error being thrown.
 
+`TRY_CAST` can be used in case the preferred behavior is not to throw an error, but to instead return a `NULL` value. `TRY_CAST` will never throw an error, and will instead return `NULL` if a cast is not possible.
```
hawkfish, line 28: s/in case/when/
s/to instead/instead to/

hawkfish (commented):
Grammar nits.

Issue #247, opened 2022-07-10
Document python client 


Mause:
The client is documented using docstrings fairly well already, we can probably just expose and publish that as is to start with?

Probably just gluing together sphinx and the sphinx-markdown-builder would be a great start

Mytherin:
Sounds good to me. Ideally this will be auto-generated from the code indeed.

PR #254, opened 2022-07-27
feat: python docs

Mause:
Pretty basic for now, only needed some patches to fix some of the weirdness in the Python API. Might require someone with more css knowledge/experience to make it look nicer

Fixes https://github.com/duckdb/duckdb-web/issues/247


Mause:
@pdet your eyes on this would need appreciated 😁

Gemfile
```diff
@@ -0,0 +1,48 @@
+source "https://rubygems.org"
```
Mytherin, line 1: What is the goal of including the `Gemfile` here? I think we need to be careful since github actions is quite finicky on what it accepts/uses. I'm not sure if this will work for the actual website.

Mause: Locking the versions means the builds are more reproducable, and it also made it easier to use a preview service (Vercel in this case, see https://duckdb-web.vercel.app).

Happy to remove however

Mytherin: Ah, that seems fine to me.

pdet:
It looks good, I had a quick view on the md file, and the function defs seem to be correctly generated, although the file seems to have some formatting issues towards the end?

Mause:
@pdet oh I see what you mean, I'll see if regenerating without the pretty_print flag helps

Mause:
@pdet okay, how does that look?

Mytherin:
Thanks, LGTM

PR #261, opened 2022-07-28
Fix an typo in art-storage

douenergy:
I think there is a typo. 
2-bit-trie.png show the index 7,10,12  , but the article write 7,10,14


Mytherin:
Thanks!

PR #269, opened 2022-08-03
feat: link to out-of-tree scanner usage docs

Mause:
Not sure if we want to inline these docs, or just leave them in the readme?


Mause:
@Mytherin also, would it be worth linking to the workflows that contain built versions of these extensions? They don't seem to be in the s3

Mytherin:
Thanks!

The fact that they are not in S3 is an oversight, they should be uploaded there.

PR #277, opened 2022-08-14
feat: update config docs

Mause:
Looks like the default values for `enable_profiling` and `profiling_mode` have changed?


Mytherin:
Could we perhaps work towards creating a single script that updates all of the docs, which we can then be run once as part of the release+archive cycle?

Mause:
> Could we perhaps work towards creating a single script that updates all of the docs, which we can then run once as part of the release+archive cycle?

I think being able to run them individually has it's merits, but happy to add a script that's just calls them all?

Mytherin:
Yes, that sounds great

Mause:
@Mytherin added here: https://github.com/duckdb/duckdb-web/pull/278/files#diff-12f4513287b674b88d7f3fb3365952db1682a45159884ac044f21b7e7b337ae5

Mytherin:
Thanks!

PR #282, opened 2022-08-23
Fix time zone name

Giorgi:
The time zone name is `America/Los_Angeles`


hawkfish (approved):
Good catch!  The tests have it right though, so I don't know how that happed - I usually copy and paste.

Giorgi:
Probably you wanted to hit Ctrl+s but you only clicked s. Happens to me sometimes 😃

PR #292, opened 2022-09-06
CSV import: I added the stdin mode

aborruso:
It's a precious thing, but still undocumented ( [#4619](https://github.com/duckdb/duckdb/discussions/4619#discussion-4363600) )


Alex-Monahan:
Hmm, thank you for adding! I think it would make more sense to put it in its own section, as this capability only exists in the CLI, not in the other clients. Or maybe in the CLI doc itself?

aborruso:
> Hmm, thank you for adding! I think it would make more sense to put it in its own section, as this capability only exists in the CLI, not in the other clients. Or maybe in the CLI doc itself?

The title of that file is `CSV Loading`, and that's another way to load a CSV (via stdin).
That's why I entered it here.

But if you prefer I put it in the cli section

aborruso:
And thank you to you @Alex-Monahan !

Mytherin:
Thanks! Looks good to me.

aborruso:
Thank you very much

PR #294, opened 2022-09-08
[FIX] typo why_duckdb

raphaelauv:


Mytherin:
Thanks for the PR!

This is actually not a typo. While DuckDB is a column-oriented RDBMS, it is at its fundamental level a relational database system. I do agree that we could perhaps mention "column-oriented" more prominently elsewhere.

PR #307, opened 2022-09-24
feat: document read_json_objects

Mause:


lnkuiper (changes requested):
Thanks!

Maybe worth mentioning that this reads newline-delimited JSON only?

PR #317, opened 2022-10-03
Add examples to aggregate/window functions

Mytherin:
Also disable font ligatures for code blocks so `!=` does not render as `≠`


Alex-Monahan:
Thanks Mark!

PR #319, opened 2022-10-04
autocomplete relative pos and close on enter

Alex-Monahan:
Love the search! Super fast, super easy, nice search suggestions. 2 small tweaks:

The autocomplete is covering up the main list, so I changed it to relative position. I also made it so that if you hit enter and you haven't picked anything in the autocomplete list, it will still close the autocomplete list (so that you can see the main list).

NOTE: I am not sure how this would work on mobile. I did test with a very small browser window, so maybe that triggered the mobile layout? What's the best way to test that?

### DIFF
--- a/css/search.css
+++ b/css/search.css
@@ -29,7 +29,8 @@ input[type=submit] {
 	color: #fff;
 }
 .autocomplete-items {
-	position: absolute;
+	/* relative position does not cover up the search results */
+	position: relative;
 	border: 1px solid #d4d4d4;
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

> NOTE: I am not sure how this would work on mobile. I did test with a very small browser window, so maybe that triggered the mobile layout? What's the best way to test that?

There hasn't really been a mobile layout (or really any serious layouting) yet.

PR #320, opened 2022-10-04
feat: restore substrait docs

Mause:
while the repo is private, we should continue to host the docs somewhere

### DIFF
--- a/docs/extensions/substrait.md
+++ b/docs/extensions/substrait.md
@@ -4,4 +4,109 @@ title: Substrait
 selected: Documentation/Substrait
 ---
 
-See [repo](https://github.com/duckdblabs/substrait#support) for usage
+The main goal of this extension is to support a both production and consumption of substrait query plans in DuckDB.
+
+This extension is mainly exposed via 3 different APIs. 1) The SQL API, 2) The Python API, 3) The R API.
+Here we depict how to consume and produce substrait query plans in each API.
+
+Additionally, see the [repo](https://github.com/duckdblabs/substrait) for further usage details.
+
+### SQL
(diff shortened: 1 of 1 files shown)


docs/extensions/substrait.md
```diff
@@ -4,4 +4,109 @@ title: Substrait
 selected: Documentation/Substrait
 ---
 
-See [repo](https://github.com/duckdblabs/substrait#support) for usage
+The main goal of this extension is to support a both production and consumption of substrait query plans in DuckDB.
+
+This extension is mainly exposed via 3 different APIs. 1) The SQL API, 2) The Python API, 3) The R API.
+Here we depict how to consume and produce substrait query plans in each API.
+
+Additionally, see the [repo](https://github.com/duckdblabs/substrait) for further usage details.
```
pdet, line 12: We should remove this link, no?

docs/extensions/substrait.md
```diff
@@ -4,4 +4,109 @@ title: Substrait
...
+     
+     To consume a substrait blob the ```from_substrait(blob)``` function must be called, from the connection, with a valid substrait BLOB plan.
+     ```python
+     query_result = con.from_substrait(proto_bytes)
+    ```
+
+### R
+Before using the extension you must remember to properly load it. To load an extension in R, you must execute the sql commands within a connection.
+```r
+con <- dbConnect(duckdb::duckdb(config=list("allow_unsigned_extensions"="true")))
```
pdet, line 87: This was prolly my mistake before, but since the repo is private this line could be
`con <- dbConnect(duckdb::duckdb())`

and the order should be reversed

```
dbExecute(con, "INSTALL('substrait')"))
dbExecute(con, "LOAD('substrait')")
```

PR #329, opened 2022-10-07
Pandas Analyze sample size documentation

Tishj:
This PR fixes my old attempt at documenting this setting

Previously I added it to the archive ..

And the placement of it was not great either, so now I've moved it to the "import from pandas" documentation page, which feels like a more suitable place for this information.


Alex-Monahan:
Thanks!
I have not seen that set global syntax before. How is that different from a pragma or a connection config option? Maybe a sentence on the pragmas or configuration pages to explain it?

PR #361, opened 2022-11-03
corrected typing and grammar mistakes

nishduckdb:
Made the following changes:
1. a instead of an
2. corrected index name


pdet:
Thanks! LGTM!

PR #363, opened 2022-11-03
Search

Franz-Kafka:
search implemented. including shortcuts. 
search string highlighting in results text tbd.


Mytherin:
Thanks! That looks excellent. The animation on mobile is genius.

One more request - could we show the search on the main page as well, rather than only in the documentation?

Alex-Monahan:
I am getting 404s when clicking on the search results when searching on the guides page (on mobile). I think the path needs to be absolute instead of relative because it was sending me here:
https://duckdb.org/docs/guides/docs/extensions/httpfs

PR #369, opened 2022-11-07
fix: restore python docs

Mause:


Mause:
Will rerun the autogeneration once I merge this

PR #384, opened 2022-11-09
fix: add missing FLOAT alias

Mause:


Mytherin:
That was fast :) Thanks

Mause:
> That was fast :) Thanks

The web editor is very convenient 😁

PR #402, opened 2022-11-16
Update to 1-based index when retrieve a map value

yothinix:
Change proposed in this pull request:

* Update to 1-based index on [Retrieving from Maps](https://duckdb.org/docs/sql/data_types/map#retrieving-from-maps) example

related to
* https://github.com/duckdb/duckdb/issues/3447


Mytherin:
Thanks!

PR #405, opened 2022-11-19
Add starts_with function and operator doc

papparapa:
Added `starts_with` function and equivalent operator `^@`, that are marged into duckdb/duckdb feature branch.
[#5334](https://github.com/duckdb/duckdb/pull/5334)
Also fixed font in the same file.


Mytherin:
Thanks!

Issue #408, opened 2022-11-21
Example usage of NaN is missing quotation marks


david-cortes:
The docs mention about the value `NaN` having a special meaning:
https://duckdb.org/docs/sql/data_types/numeric

Then gives out examples using it:
https://duckdb.org/docs/sql/functions/numeric

One such example is this:
```sql
isnan(NaN)
```

But getting a NaN in duckdb requires quoting it - i.e. this doesn't work:
```sql
SELECT NaN
```
```
Error: rapi_prepare: Failed to prepare query SELECT NaN
Error: Binder Error: Referenced column "NaN" not found in FROM clause!
LINE 1: SELECT NaN
               ^
```

Has to be specified like this:
```sql
SELECT 'NaN'::float
```

PR #420, opened 2022-11-26
Fix infinity and nan cast

papparapa:
[#408](https://github.com/duckdb/duckdb-web/issues/408)
Fixed cast in `isinf` and `isnan` examples.


Mytherin:
Thanks!

Issue #77, opened 2021-10-07
Add UnNest to Nested Types Page


Alex-Monahan:
I just wanted to document this here before I forgot! I'm happy to make this change.

I think that the UnNest function should be mentioned in the Nested types section in addition to the SELECT overview. 

In general, I think there might be a few other Postgres functions that aren't in the docs just yet, but we can track those in separate issues.

Mytherin:
Yes, that makes sense. I think the UNNEST function actually warrants an extra page, since it is a "special" function in the sense that it changes the result cardinality. The functions in the Nested types section could contain UNNEST with a reference to this separate page.

Alex-Monahan:
Looks great, thank you!!

PR #421, opened 2022-11-27
Clarify multiprocess read/write support

davetapley:
Is still confusing people, see e.g.:
https://github.com/duckdb/duckdb/issues/77
https://github.com/duckdb/duckdb/issues/5317

### DIFF
--- a/index.html
+++ b/index.html
@@ -203,7 +203,7 @@ Pkg.add("DuckDB"){%- endhighlight -%}
 					<ul>
 						<li>High-volume transactional use cases (e.g. tracking orders in a webshop)</li>
 						<li>Large client/server installations for centralized enterprise data warehousing</li>
-						<li>Writing to a single database from multiple concurrent processes, or:</li>
+						<li>Writing to a single database from multiple concurrent processes</li>
 						<li>Multiple concurrent processes reading from a single writable database</li>
 					</ul>
 				</div>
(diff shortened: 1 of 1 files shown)


Alex-Monahan:
Great thought! I'm going to add an FAQ entry for it. I think it deserves some extra explanation!

pbowyer:
> Great thought! I'm going to add an FAQ entry for it. I think it deserves some extra explanation!

Definitely. As a potential user it's a real "gotcha" when coming from sqlite or any DB. Not having concurrent writes I get, but having read #77 why there's the inability to read when there's a writer connected or vice versa is still a mystery, if the writes are going to the WAL.

Alex-Monahan:
> > Great thought! I'm going to add an FAQ entry for it. I think it deserves some extra explanation!
> 
> Definitely. As a potential user it's a real "gotcha" when coming from sqlite or any DB. Not having concurrent writes I get, but having read #77 why there's the inability to read when there's a writer connected or vice versa is still a mystery, if the writes are going to the WAL.

Ah, so the reason for that is in the case where the WAL checkpoints (inserts into the DB and then clears out the WAL). Handling that situation gracefully in the multiprocess case is not trivial.

Alex-Monahan:
Thanks again for the PR! I also added some information in the FAQ for more details. https://duckdb.org/faq

PR #425, opened 2022-12-01
Fix `strlen` example

camdencheek:
The example listed for the `strlen` function incorrectly shows an example for `length` instead.


Mytherin:
Thanks!

PR #427, opened 2022-12-01
fix: use json module to generate search index

Mause:
only really introduced some minor changes around escaping characters


Mytherin (approved):
Thanks! That is indeed much cleaner. LGTM

PR #431, opened 2022-12-03
Concurrency and Labs/MotherDuck FAQ's

Alex-Monahan:
Hi folks! I tried to paraphrase some previous answers by @Mytherin on the subject of concurrency and MVCC, but I could definitely use a review on this one! I also added an FAQ entry for the Foundation/Labs/MotherDuck relationships, so I would love another set of eyes on that one as well. 

Many thanks!


faq.md
```diff
@@ -22,7 +22,17 @@ https://github.com/duckdb/duckdb-web). Pull requests to fix issues or generally
...
 
+### How does DuckDB handle concurrency within a single process?
+DuckDB supports multiple writer threads using a combination of MVCC (Multi-Version Concurrency Control) and optimistic concurrency control. As long as there are no write conflicts, multiple concurrent writes will succeed. Appends will never conflict, even on the same table. Multiple threads can also simultaneously update separate tables or separate subsets of the same table. Each subset is ~20,000-80,000 rows depending on the width of the rows. Optimistic concurrency control comes into play when two threads attempt to edit (update or delete) the same row of data at the same time. In that situation, the second thread to attempt the edit will fail with a conflict error.
```
Mytherin, line 32: This part might be a bit too detailed for a FAQ entry:

> Multiple threads can also simultaneously update separate tables or separate subsets of the same table. Each subset is ~20,000-80,000 rows depending on the width of the rows.

While this is granularity at which we do locking (ColumnSegments) - that should be transparent for the user as long as the same row is not being updated. I would remove the sentence "Each subset is ~20,000-80,000 rows depending on the width of the rows." here and talk only about the optimistic concurrency control.

Mytherin (commented):
Thanks! Looks great as always. One minor comment - otherwise it's all good from my side:

Issue #424, opened 2022-11-30
Python Execute SQL Suggested Clarification with "Python dataframe"


d33bs:
Hello - thank you for the amazing work on DuckDB and related documentation found here! I recently noticed a minor issue with some of the documentation and wanted to suggest a correction.

https://duckdb.org/docs/guides/python/execute_sql currently includes the following sentence:
```
Use df if you would like the result to be returned as a Python dataframe instead.
```

This may be improved for clarity with the following (avoiding an implied native Python dataframe format):
```
Use df if you would like the result to be returned as a Python Pandas dataframe instead.
```

Please let me know if you have any questions or if I may clarify the above.

lnkuiper:
Sounds good. Feel free to send a PR if you spot any inaccuracies or anything unclear in our documentation

PR #432, opened 2022-12-03
Add pandas dataframe specificity

d33bs:
Addresses #424 to clarify which type of Python dataframe is returned from `.df()` when using the Python API.


Mause:
Thanks!

PR #442, opened 2022-12-08
Minor update R API docs

eitsupi:
- Use `<-` instead of `=`, which is often recommended by coding rules.
- Use base pipe `|>` instead of magrittr pipe `%>%`.
- Use the `na.rm = TRUE` option of `mean()` in `summarise()`. If this is not specified, a warning message is displayed.
  ```
  Warning message:
  Missing values are always removed in SQL aggregation functions.
  Use `na.rm = TRUE` to silence this warning
  ```
- Use `collect()` at the end of a dbplyr query. This will convert the result of the computation into a data.frame.


docs/api/r.md
```diff
@@ -95,9 +95,10 @@ library("dplyr")
 con <- dbConnect(duckdb::duckdb())
 duckdb::duckdb_register(con, "flights", nycflights13::flights)
 
-tbl(con, "flights") %>% 
-  group_by(dest) %>%
-  summarise(delay = mean(dep_time))
+tbl(con, "flights") |>
+  group_by(dest) |>
+  summarise(delay = mean(dep_time, na.rm = TURE)) |>
```
Mause, line 100:
```suggestion
  summarise(delay = mean(dep_time, na.rm = TRUE)) |>
```

eitsupi: Sorry for typo...

eitsupi:
@Mause Any chance to check this again? Thanks.

Mause:
> @Mause Any chance to check this again? Thanks.

Sorry, I'm not familiar with R, so I'm not comfortable reviewing this. Perhaps @hannes can take a look?

Mytherin:
LGTM, thanks!

Issue #429, opened 2022-12-02
Add generalized documentation on SQL prepared statement dollar symbol ($1) parameter usage


d33bs:
Prepared statements with DuckDB clients may use question marks (?) or dollar signs ($1, $2, etc), similar to what's found within the C client documentation ([reference]( https://duckdb.org/docs/api/c/prepared.)). Using the dollar sign notation allows the parameter references to be reused within the SQL statement. Generalized documentation on this topic may assist those making use of DuckDB with other clients, for example with the Python client (where the functionality exists but is not yet documented).

szarnyasg:
See also #767

szarnyasg:
Fixed via #1678

PR #447, opened 2022-12-11
Add dollar symbol prepared statement notes for Python API

d33bs:
This PR comes in the spirit of stepping towards addressing #429 with what I understand so far about how this syntax operates for DuckDB API's. Something I wasn't sure about was whether prepared statements work for all API's or only some. In the meantime I thought it might be important to cover that it is supported by the Python API.

Thanks in advance for any thoughts or feedback you may have!


Alex-Monahan:
Thanks very much! This is great. Love the duck duck goose reference...

d33bs:
Cheers, thank you @Alex-Monahan !

PR #456, opened 2022-12-15
Fix timezone function example

papparapa:
Fixed `timestamp` function example results.


hawkfish (approved):
Thanks! Next time, instead of doing the math in my head I'll just run the query...

PR #458, opened 2022-12-17
Issue #5712: Document localtimestamp

hawkfish:
Also collect local time operations into their own section.


Alex-Monahan:
Looks great! Thanks as always!

PR #459, opened 2022-12-18
feat: document storage info

### LINKED ISSUE #177: Document the `storage_info` pragma
Hey folks!

On Discord, @hannes mentioned the `storage_info` pragma as a way to get the storage footprint of a table:

```sql
PRAGMA storage_info('my_table');
```

The columns provide a ton of info:

```
row_group_id,
column_name,
column_id,
column_path, 
segment_id,
segment_type,
start,
count, 
compression,
stats,
has_updates,
persistent,
block_id,
block_offset
```

I'm looking to calculate the storage size of a table, so I assume that means summing `count` here.

This would be an awesome pragma to document if possible! Very useful for those of us telling the machine to do a lot of big calculations in DuckDB.

Mause:
Fixes #177 
Fixes https://github.com/duckdb/duckdb/discussions/5730

### DIFF
--- a/.github/workflows/jekyll.yml
+++ b/.github/workflows/jekyll.yml
@@ -49,7 +49,7 @@ jobs:
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
-    if: github.ref == 'refs/heads/master'
+    if: github.ref == 'refs/heads/master' && github.repository == 'duckdb/duckdb'
     runs-on: ubuntu-latest
     needs: build
     steps:
--- a/docs/sql/pragmas.md
+++ b/docs/sql/pragmas.md
@@ -222,3 +222,30 @@ PRAGMA disable_checkpoint_on_shutdown;
 ```sql
 PRAGMA temp_directory='/path/to/temp.tmp'
 ```
+
+### storage_info
+
+```sql
+PRAGMA storage_info('table_name');
+```
+
+returns the following, per column in the given table
+
+| name           | type      | description                                           |
+|----------------|-----------|-------------------------------------------------------|
+| `row_group_id` | `BIGINT`  ||
+| `column_name`  | `VARCHAR` ||
+| `column_id`    | `BIGINT`  ||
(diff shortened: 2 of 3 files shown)


Tishj:
We could also make the storage version read a function on the `duckdb` module
`duckdb.storage_version()` -> return storage version of current latest version
`duckdb.storage_version(file_name)` -> try to run the code you listed as example, throwing an exception if the input is invalid

Mause:
> We could also make the storage version read a function on the `duckdb` module `duckdb.storage_version()` -> return storage version of current latest version `duckdb.storage_version(file_name)` -> try to run the code you listed as example, throwing an exception if the input is invalid
> 
> What do you think?

What's the usecase you envision with those?

Mytherin:
> We could also make the storage version read a function on the `duckdb` module `duckdb.storage_version()` -> return storage version of current latest version `duckdb.storage_version(file_name)` -> try to run the code you listed as example, throwing an exception if the input is invalid
> 
> What do you think?

I was thinking about adding this list to DuckDB internally and having it report the DuckDB version when opening an old file.

Mause:
> > We could also make the storage version read a function on the `duckdb` module `duckdb.storage_version()` -> return storage version of current latest version `duckdb.storage_version(file_name)` -> try to run the code you listed as example, throwing an exception if the input is invalid
> > 
> > What do you think?
> 
> I was thinking about adding this list to DuckDB internally and having it report the DuckDB version when opening an old file.

Do you mean without having to parse it from the error message?

https://github.com/duckdb/duckdb/blob/16f362852baeb64e6a725eac12a9d45982d40901/src/storage/single_file_block_manager.cpp#L156-L165

Mytherin:
Yes, exactly.

Mause:
I think it would still be worth having this documented regardless?

Mytherin:
Yes, for sure

Mytherin:
Sure, I have pushed a minor fix, otherwise this looks good to me.

Mytherin:
Thanks!

PR #464, opened 2022-12-19
Update window function documentation.

voberoi:
This change adds to two clarifications to the window function docs:

1. How to specify multiple windows in the same `WINDOW` clause.
2. Where to find `WINDOW` in the canonical order of a select statement.

---

This was on my to-do list and I'm just getting around to it. I expressed some confusion in Discord a week or two ago and @Alex-Monahan suggested I updated the docs.

I generated the docs and tested the link -- it works.

Happy to make any changes -- just let me know!

### DIFF
--- a/docs/sql/window_functions.md
+++ b/docs/sql/window_functions.md
@@ -215,6 +215,35 @@ ORDER BY 1, 2
 
 The three window functions will also share the data layout, which will improve performance.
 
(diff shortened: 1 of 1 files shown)


hawkfish (approved):
It might be worth noting in future that these two clauses use the same partitioning and ordering, so all 6 functions will be computed together. If (say) one of the sorts was `DESC`, then they would be computed in two groups of 3 with different sorting.

PR #466, opened 2022-12-20
fix: only check links post deploy

Mause:
not much point checking before the deploy is finished if new pages are being added?


Alex-Monahan:
Maies sense to me! I assume if the build breaks we still don't deploy? Should be good!

Mause:
> Maies sense to me! I assume if the build breaks we still don't deploy? Should be good!

Yup!

PR #470, opened 2022-12-22
Add documentation on how to create enums from data file

Tmonster:
Small update to the enum documentation so users can see how to create an enum from a `csv`, `parquet` etc. file.


Alex-Monahan:
This looks great! I'm going to merge this and then add another example of a more general case (just using values in a select rather than from a csv).

Alex-Monahan:
See #479 for the additional enum documentation

PR #471, opened 2022-12-23
Add attach documentation

Mytherin:


Alex-Monahan:
This looks great! When would you like this merged - before or after this feature gets released? It's not often that the docs are ahead of release like this (thank you!!). I'm also thinking of adding in a guide with an example of when attaching a DB comes in handy to get folks thinking creatively...

Mytherin:
Now is good - we might want to rename "current" to "dev" and add the latest released version to the documentation history.

PR #478, opened 2023-01-03
Document POSTGRES_SCAN_PUSHDOWN usage

davetapley:
I only found it by looking at source:
https://github.com/duckdblabs/postgres_scanner/blob/70a6eb867207599e6cb9225d22f09c55d973dd00/postgres_scanner.cpp#L1063


Alex-Monahan:
Thanks for adding this in!

PR #480, opened 2023-01-08
row_group_size parameter

Alex-Monahan:
Added the row_group_size parameter to the parquet data import page. Also added a paragraph about how to choose what to set it to.


Alex-Monahan:
Sorry, doing a small amount of yak shaving so this is not ready to be merged. Need to split the copy to/from parameters since only some are valid in each case. Might want to split by CSV and parquet although there is some overlap... Not sure yet!

Alex-Monahan:
Ready to merge if anyone wants to take a quick look. Otherwise after a day or two I'll go ahead and merge it!

PR #481, opened 2023-01-08
Add date part specifier synonym docs

papparapa:
Added documentation about [date part specifier synonyms that have been merged into duckdb master branch](https://github.com/duckdb/duckdb/pull/5845).


Alex-Monahan:
Thank you for both the code and the docs! It's really awesome when folks do both - we really appreciate it!

PR #483, opened 2023-01-09
Fix arg_min & arg_max descriptions (user feedback)

Alex-Monahan:


docs/sql/aggregates.md
```diff
@@ -38,8 +38,8 @@ The table below shows the available general aggregate functions.
 
 | Function | Description | Example | Alias(es) |
 |:---|:---|:---|:---|
-| `arg_max(arg,val)` |Calculates the arg value for a maximum val value. | `arg_max(A,B)` | `argMax(A,B)`, `max_by(A,b)` |
-| `arg_min(arg,val)` |Calculates the arg value for a minimum val value. | `arg_min(A,B)` | `argMin(A,B)`, `min_by(A,B)` |
+| `arg_max(arg,val)` |Sorts by the `val` expression and finds the row with the maximum `val`. Calculates the `arg` expression at that row. | `arg_max(A,B)` | `argMax(A,B)`, `max_by(A,b)` |
```
hawkfish, line 41: Hmm the point is it *doesn't* sort!

> Returns the value of `arg` corresponding the the maximum/largest value of `val`?

We can't change the name of the function, but maybe the arguments could be improved? (`val`, `ordering`)?

Alex-Monahan: Oh, so it is doing a max aggregate function and recording the row then?

hawkfish: Yes! It records `arg` and `val`, updates both when the new value of `val` is larger/smaller and returns `arg` at the end (finalise). It occurs to me that this could be very powerful if you stuck all the values you want from the row into a struct(!)

Alex-Monahan: > Yes! It records `arg` and `val`, updates both when the new value of `val` is larger/smaller and returns `arg` at the end (finalise). It occurs to me that this could be very powerful if you stuck all the values you want from the row into a struct(!)

Ooh, that's a nice idea... Maybe I can add a quick guide with an example like that.

Alex-Monahan: One other thought! Could you use a struct in the val as well to use multiple criteria for determining your max? Ex: `{"ts":timestamp_column, "tiebreaker": tiebreaker_column}`?

Issue #495, opened 2023-01-19
ANY_VALUE aggregate function is not documented


droher:
The ANY_VALUE aggregate function was added here: https://github.com/duckdb/duckdb/issues/4092
But there is no documentation of it on the site. I'd be happy to put up a PR for it if that works.

PR #497, opened 2023-01-19
Add ANY_VALUE aggregate to docs

droher:
Adds a row for the `ANY_VALUE` aggregate added in https://github.com/duckdb/duckdb/pull/4259.
Resolves https://github.com/duckdb/duckdb-web/issues/495

### DIFF
--- a/docs/sql/aggregates.md
+++ b/docs/sql/aggregates.md
@@ -38,6 +38,7 @@ The table below shows the available general aggregate functions.
 
 | Function | Description | Example | Alias(es) |
 |:---|:---|:---|:---|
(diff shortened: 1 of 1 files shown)


docs/sql/aggregates.md
```diff
@@ -38,6 +38,7 @@ The table below shows the available general aggregate functions.
 
 | Function | Description | Example | Alias(es) |
 |:---|:---|:---|:---|
+| `any_value(arg)` |Returns an arbitary _non-null_ value from arg. | `any_value(A)` | - |
```
Mause, line 41: This isn't correct I don't think? It returns the **first** non-null value, not an arbitrary one?

Mause (changes requested):

Mause:
Thanks!

PR #498, opened 2023-01-19
fix: extract search index into own json file

Mause:
This should drastically reduce the size of the generated website, hopefully fixing the deployment timeout


Mause:
700mb+ to 82mb! I'd call that a success!

PR #502, opened 2023-01-22
Fix make_timestamptz example result

papparapa:
This PR fixes `make_timestamptz` example result.

In countries that use Central European Time, September is in the middle of daylight saving time.
Therefore the appropriate offset to be appended to `1992-09-20 15:34:27.123456` is `+02` and `1992-09-20 15:34:27.123456+02` is written as `1992-09-20 06:34:27.123456-07` in America/Los_Angeles time zone.

Actually `make_timestamptz`  returns this expected result.


Alex-Monahan:
Good catch! Thanks!

PR #512, opened 2023-01-26
Update map type docs

LindsayWray:
This PR updates the information regarding MAP type.


Alex-Monahan:
Thank you for the update- much appreciated!

PR #514, opened 2023-01-27
feat: add filesystems guide

Mause:


Alex-Monahan:
This is great @Mause! Super cool that you have allowed DuckDB to benefit from some of the goodness only available in Python-land.

As an aside, is this an approach that you think we could open up to to others in some way, like maybe Python-specific extensions? If so, maybe we include it within the upcoming DuckDB extensions blog post?

PR #516, opened 2023-01-29
Improve PRAGMA database_size docs

tuffnatty:
The current docs make an impression that `PRAGMA database_size` is the way to get individual table sizes. Obviously it is not ;-( . Also, the `database_name` column was missing from the documentation.


Alex-Monahan:
Good catch! Thanks for the PR - we appreciate it!

PR #517, opened 2023-01-30
Fix two-byte blob example (\x in front of every byte)

sebastian-nagel:


Alex-Monahan:
Thank you for the fixes - much appreciated!

PR #521, opened 2023-02-02
overwrite flag for sql_scan

mneedham:


docs/extensions/sqlite_scanner.md
```diff
@@ -118,5 +118,13 @@ SET GLOBAL sqlite_all_varchar=true;
 
 When set, this option overrides the type conversion rules described above, and instead always converts the SQLite columns into a `VARCHAR` column. Note that this setting must be set *before* `sqlite_attach` is called.
 
+## Data Types
```
Mause, line 121: Can you review this header? Doesn't make a lot of sense given the contents of the section

Mause (changes requested):

docs/extensions/sqlite_scanner.md
mneedham: Yeh, good point. How about now?

Alex-Monahan: Looks good, thanks!

PR #527, opened 2023-02-08
Adding videos and Lightning talks to the summary table

pdet:


Mause:
Might be nicer to just have the video titles as the links, rather than a column that is just "Here, Here, Here, Here, Here, Here, Here, Here"

pdet:
True, I'll add a hyperlink to the title

PR #546, opened 2023-02-13
Fix WITH RECURSIVE example graph figure

szarnyasg:
The previous example graph in the page of the [WITH clause](https://duckdb.org/docs/sql/query_syntax/with) was incorrect, I fixed it.


Mytherin:
Thanks!

PR #552, opened 2023-02-16
fix: Don't use private method for polars df init

ritchie46:
`DataFrame._from_dict` is private. Adapted so we use the default constructor.


Mause:
Thanks!

PR #553, opened 2023-02-16
Expand CSV reader documentation

Mytherin:
This PR adds several more pages to the CSV reader explaining additional features (e.g. how the auto-detect works, how to override it, how to read from multiple pages) and also adds a `tips` section that we can hopefully expand in the future with CSV reading tips.


Alex-Monahan:
This is awesome! Thank you for putting this together. It really helps me build up an intuition around how the csv reader will behave, which will really help folks troubleshoot gnarly csv's as you said! Whenever you are ready to merge, feel free! Looks great to me.

Issue #474, opened 2022-12-26
Document direct download links for extensions


danthegoodman1:
For doing something like making a lambda layer, pre-building a container, etc. it's really helpful to pre-download an extension from the domain path `https://extensions.duckdb.org/${REVISION (including "v" prefix)}/${PLATFORM}/${NAME}.duckdb_extension.gz;`

If this could be included under the extensions section, along with the list of `PLATFORMs` that would be very helpful to folks pre-building environments.

Is there a list available right now of the names of the platforms?

_And merry Christmas 🎄_

Alex-Monahan:
This should be closed now! Let me know if something is missing in the merged PR above.

PR #554, opened 2023-02-16
downloading extensions from S3 and loading

Alex-Monahan:


Alex-Monahan:
Related to issue #474 . Let me know if this looks helpful, or if we don't want to advise folks to do this!

Alex-Monahan:
Merging this since I mistakenly left it on my master branch. Happy to remove if it is not good advice for our users!

PR #566, opened 2023-02-21
Add descriptions about output formats

papparapa:
This PR adds descriptions about output formats like duckdox.


Alex-Monahan:
Thanks! This is a helpful reference that I am sure to refer back to myself!

papparapa:
Thanks for escaping bracket! I missed that.

PR #568, opened 2023-02-22
adding data load tool (dlt) - DuckDB destination

TyDunn:
I would like to add data load tool (dlt) - DuckDB destination as a project mentioned under [Other Projects](https://duckdb.org/why_duckdb#other-projects). I'm opening a PR since I thought it would be easier for everyone, but I can open an issue if necessary. Thanks!


Alex-Monahan:
Thanks for integrating with DuckDB! Welcome to the flock!!

PR #575, opened 2023-02-26
Update window_functions.md

maartenbosteels:
Add missing "partition by" and "order by"


Alex-Monahan:
Good find! Thank you for the fix!

PR #581, opened 2023-02-28
Clarify default Calendar setting.

hawkfish:
The default calendar for ICU is actually not `gregorian` - e.g., if you are in China it will use one of the Chinese calendars.


Alex-Monahan:
Great point! Thanks!

PR #582, opened 2023-02-28
Bump to version 0.7.1

carlopi:
Unsure whether other changes are needed (e.g. generate _data/menu_docs_071 ?) or if they happen automagically.


Mause:
Shouldn't we be using the script to do this, instead of manually?

carlopi:
I had not found the script after few minutes of browsing, but I was sure it had to be there...
I will close this, please someone do it the proper way.

PR #583, opened 2023-02-28
Fix #469 updating sidebar interaction model

### LINKED ISSUE #469: duckdb.org/docs - Update left nav so it has a consistent interaction model 💯
When i'm on the docs page and look at the left panel where the nav tree is I notice that some of the parent items have different interactions. I recommend they have the same behavior. For example:

1. On page https://duckdb.org/docs/ you seen "Installation" at the top and displayed as a section header on the sitemap ( https://jam.dev/c/95c12d48-e835-4ad0-9dca-4d417aec5f4a) but when you click "Installation" the new page header read "DuckDB Installation" they should be the same throughout: Should say "Installation" or you should update the left nav to match it. https://jam.dev/c/aabd8c51-5425-4e3b-8caa-157f3a64306b

2. Selected pages in the left nav should be a different color and/or bold to show they're selected ([screen shot](https://jam.dev/c/d56bf42a-1a4b-4f63-8136-901ade4455ab))

3. Some parent links in the tree can be selected and some can't, they should all have the same behavior.
- "Guides" opens the "Guides" page while "Documentation" just expands and collapses the tree branch. That's weird? 

Those are just a few little fixes that will improve your nav and interaction model, hope that helps - much love from the MotherDuck team 👍

carlopi:
As per issue here: https://github.com/duckdb/duckdb-web/issues/469

I think this should improve navigation, adding a reachable Sitemap top-level + having only leaves having content (this is more visible on mobile, whether a click means both open sub-menu AND go to page)

I personally think this makes sense, but would be nice to have someone (@Alex-Monahan, @Mause, ?) taking a proper look at this.

### DIFF
--- a/_data/menu_docs_dev.json
+++ b/_data/menu_docs_dev.json
@@ -9,8 +9,11 @@
       "page": "Guides",
       "open": true,
       "slug": "guides/",
-      "url": "index",
       "mainfolderitems": [
+        {
+          "page": "Overview",
+          "url": "index"
+        },
         {
           "page": "Data Import & Export",
           "slug": "import",
@@ -207,9 +210,12 @@
             },
             {
               "page": "CSV Files",
-              "url": "csv",
               "slug": "csv",
               "subsubfolderitems": [
+                {
+                  "page": "Overview",
+                  "url": "/data/csv"
+                },
                 {
                   "page": "Auto Detection",
                   "url": "auto_detection"
--- a/_data/menu_side.json
+++ b/_data/menu_side.json
@@ -6,9 +6,12 @@
       "mainfolderitems": [
         {
           "page": "Testing",
-          "url": "/dev/testing",
           "slug": "sqllogictest",
           "subfolderitems": [
+            {
+              "page": "Overview",
+              "url": "/dev/testing"
+            },
             {
               "page": "SQLLogicTest",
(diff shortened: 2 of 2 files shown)


Alex-Monahan:
Ah, let me see if I understand what you mean. Each section header should be a link that takes you to the overview page directly? Or do we want to avoid that behavior and only expand the tree menu when clicking / tapping on a header?

carlopi:
I am not completely sure this makes sense, but the idea behind this is:
Each 'leaf' node is an article, and each 'internal' node only expand the menu. Moving eventual content from the 'internal' nodes to an added 'Overview' leaf node as first child of the internal node.

Plus moving the content currently reachable from here (https://duckdb.org/docs/), but that you can't go back from the docs itself while navigating, to a 'Sitemap' own category (that itself is a 'leaf' node of the tree).

I am not completely sure this works, since I have a hard time building this on work mac thanks to dependency hell, so would not trust this without a proper look from someone else.

Mause:
I'm supportive of this - we should also fix the Guides heading though

Alex-Monahan:
Now I understand! This is much nicer. Thanks! 

@Mause - I believe this fixes the guides heading already? Or is there more to fix?

Mause:
@Alex-Monahan oops you're right 😅

PR #584, opened 2023-02-28
Document split_part function

Alex-Monahan:


Alex-Monahan:
If the docs guy adds a feature, maybe he should document it... ;-)

PR #590, opened 2023-03-02
CLI:stdin/stdout, duck head prompt,direct commands

Alex-Monahan:
More CLI capabilities!

### DIFF
--- a/docs/api/cli.md
+++ b/docs/api/cli.md
@@ -136,14 +136,6 @@ D .help sh
 .show                    Show the current values for various settings
 ```
 
-## Configuring the CLI
-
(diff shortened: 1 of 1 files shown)


docs/api/cli.md
```diff
@@ -476,3 +504,39 @@ D load 'fts';
 ```
 
 <!-- SQL parameters do not appear to work -->
+
+## Reading from stdin and writing to stdout
+
+When in a Unix environment, it can be useful to pipe data between multiple commands. 
+DuckDB is able to read data from stdin as well as write to stdout using the file location of stdin (`/dev/stdin`) and stdout (`dev/stdout`) within SQL commands, as pipes act very similarly to file handles.
```
Mause, line 511:
```suggestion
DuckDB is able to read data from stdin as well as write to stdout using the file location of stdin (`/dev/stdin`) and stdout (`/dev/stdout`) within SQL commands, as pipes act very similarly to file handles.
```

Mause (changes requested):

PR #594, opened 2023-03-03
Update JDBC docs for latest PRs

MariusVolkhart:
PRs merged that lead to these changes:
- https://github.com/duckdb/duckdb/pull/6376
- https://github.com/duckdb/duckdb/pull/6375
- https://github.com/duckdb/duckdb/pull/6377

### DIFF
--- a/docs/api/java.md
+++ b/docs/api/java.md
@@ -34,11 +34,13 @@ ro_prop.setProperty("duckdb.read_only", "true");
 Connection conn_ro = DriverManager.getConnection("jdbc:duckdb:/tmp/my_database", ro_prop);
 ```
 
-The JDBC `DriverManager` API is a relatively poor fit for embedded database management systems such as DuckDB. If you would like to create **multiple connections to the same database**, it would be somewhat logical to just create additional connections with the same URL. This is however only supported for read-only connections. If you would like to create multiple read-write connections to the same database file or the same in-memory database instance, you can use the custom `duplicate()` method like so:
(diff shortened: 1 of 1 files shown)


docs/api/java.md
```diff
@@ -34,12 +34,14 @@ ro_prop.setProperty("duckdb.read_only", "true");
 Connection conn_ro = DriverManager.getConnection("jdbc:duckdb:/tmp/my_database", ro_prop);
 ```
 
-The JDBC `DriverManager` API is a relatively poor fit for embedded database management systems such as DuckDB. If you would like to create **multiple connections to the same database**, it would be somewhat logical to just create additional connections with the same URL. This is however only supported for read-only connections. If you would like to create multiple read-write connections to the same database file or the same in-memory database instance, you can use the custom `duplicate()` method like so:
```
Mause, line 37: Can we keep mention of the limitations mentioned here?

Mause (changes requested):

docs/api/java.md
MariusVolkhart: The limitations don't apply anymore, that's why I removed this section. See [this comment](https://github.com/duckdb/duckdb/pull/6375#discussion_r1113713688). There's a test that opens 2 write, and separetly, 2 read, connections: https://github.com/duckdb/duckdb/blob/65a9ad239bd1f66029b401d2d505997464d04ff0/tools/jdbc/src/test/java/org/duckdb/test/TestDuckDBJDBC.java#L1474

Mause: Ah sweet, no worries then!

PR #595, opened 2023-03-04
docs: show how to import from Cloudflare R2

elithrar:
This change shows how to import Parquet data from Cloudflare's R2 object storage via its S3 compatibility API.

Working example: https://twitter.com/elithrar/status/1631985701918048256?s=20

image

Disclaimer: I work for CF, and had a few users ask how to wire this up, hence the PR!


Alex-Monahan:
Fantastic, thanks for the help! This will be useful to a number of folks.

PR #596, opened 2023-03-04
csv.md to overview.md and move into csv folder

Alex-Monahan:
Fixes the csv link with the new nicer menu interactivity


Mytherin:
I actually did this intentionally so the old link would keep working. Now there are a bunch of broken links on the website, and Google no longer indexes the current CSV reading docs correctly (instead linking to an archived version). Perhaps we should revert or find a way to preserve the old link?

Mytherin:
Have added this in #606

PR #599, opened 2023-03-05
pattern matching flag example and RE2 link

alexboks:
Add an example of using regex flag and link to more docs on RE2


Mause:
Thanks!

PR #603, opened 2023-03-06
Add MacOS section for ODBC

blackrez:
Hello, I use odbc on MacOS and I add the documentation. I only tested it with unixODBC built from homebrew.

I know there is an another solution to use odbc on MacOS but I will try later.


Alex-Monahan:
Thank you for building this documentation! We really appreciate it.

As I am a Windows user, I can't quite test this, and I'm used to having ODBC be built into my OS! Maybe someone else on the team could have a look?

PR #606, opened 2023-03-06
Add redirect of old CSV link to new CSV link

Mytherin:
This adds the `jekyll-redirect-from` plugin as well, we might need to add that somewhere in the requirements


Mause:
It's already included, from the original github pages template we used: https://github.com/duckdb/duckdb-web/blob/master/Gemfile#L18

PR #608, opened 2023-03-06
Formatting for readability

rmoff:


Alex-Monahan:
Howdy! Unfortunately, this breaks the format of the accordions we have in place (so those extra paragraphs are always visible and not minimized). 
image

Alex-Monahan:
Thanks for the bolding though! I'll merge that portion.

PR #613, opened 2023-03-08
Document hash() SQL function

pmm-motif:
Looks like `hash` function (https://github.com/duckdb/duckdb/pull/3077) is not documented. I added a short description for it. 

Not sure if it's in the right section (since it doesn't apply just to strings), but concluded that the proximity of `md5` makes it a good fit


Alex-Monahan:
Thank you! That is a good spot. I also added it to the utility functions doc, as md5 is in both of those places as well...

PR #616, opened 2023-03-08
Use same table name in java code examples

dinsaw:
In the first example, we have named the table as `items`, in the second code example its name is `test`.  This resolves that, making it frictionless for users (like me) who like to copy-paste the entire code.


Mytherin:
Thanks! LGTM

PR #618, opened 2023-03-09
lakeFS Blog Post. Images resized to 680 width

Alex-Monahan:
Post is dated tomorrow (2023-03-10), so it won't be visible if we merge today (which is what we want!)


Mause:
> so it won't be visible if we merge today (which is what we want!)
Is this a new Jekyll feature I'm not aware of? I'm not sure how a static site generator could implement this

Alex-Monahan:
> > so it won't be visible if we merge today (which is what we want!)
> 
> Is this a new Jekyll feature I'm not aware of? I'm not sure how a static site generator could implement this

It just looks at the filename (2023-03-10) and won't render if the date is in the future... Nothing too fancy!

PR #622, opened 2023-03-10
:student:  Update CASE statement > Add AS keyword

adriens:
# :grey_question: About

The current documentation for `CASE` statement does not emebed the `AS` alias, which makes queries easier to use by putting alias (to create VIEWs or TABLEs on top of the `case` statement)

# :moneybag: Benefits

Merging this PR would make code sample even easier to use.


Alex-Monahan:
Thanks for the edits! That will make the output cleaner if folks run those examples for sure. We appreciate it!

adriens:
Thanks @Alex-Monahan for the kind words :pray:

PR #627, opened 2023-03-13
Add JSON import/export docs/guides

lnkuiper:
I used the CSV import/export docs as an example. I will revisit this when I do more development on the JSON extension.

### DIFF
--- a/_data/menu_docs_dev.json
+++ b/_data/menu_docs_dev.json
@@ -51,6 +51,14 @@
               "url": "s3_export"
             },
             {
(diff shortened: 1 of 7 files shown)


Alex-Monahan:
Thanks so much for writing this up!

Of course, only after I added examples to the json_import file did I see that you had already included similar content in the JSON extension doc... I'm happy to remove if you'd like. Just let me know and I'll edit and merge!

Alex-Monahan:
Good stuff! Thanks again!

PR #640, opened 2023-03-16
Add homebrew installation doc

mehd-io:
There is a nice duckdb [homebrew formula](https://formulae.brew.sh/formula/duckdb#default) but it's not referenced anywhere in the documentation.
I believe this could be the default way for any macos user, it's added to the `PATH`, and managing upgrade is straightforward.

Not sure about the location where to put it through, happy to hear your thoughts

### DIFF
--- a/_includes/installation.html
+++ b/_includes/installation.html
@@ -159,7 +159,8 @@ DBInterface.execute(con, "CREATE TABLE integers(i INTEGER)"){%- endhighlight -%}
 	</div>
 
 	<div class="latest cli binary macos">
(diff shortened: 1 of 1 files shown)


Alex-Monahan:
Thanks! Looks good to me. I am not a Mac user though, but thanks for keeping both options in there!

Mause:
@Alex-Monahan do you know if the performance issues have been fixed with the brew version? I recall a discord conversation about the default brew compilation flags that caused decreased performance

carlopi:
To my knowledge, they have not been touched. The last commit has been just after 0.7.1 release: https://github.com/Homebrew/homebrew-core/commits/master/Formula/duckdb.rb.

mehd-io:
@Mause could you give more info about this? I've been using it through `homebrew` but I haven't faced any perf issue so far 🤔 Happy to dive a bit more...

PR #646, opened 2023-03-21
Replace TRUE with 1 and FALSE with 0 in FTS docs

wpdonders:
The FTS documentation show the full call signature for the PRAGMA  `create_fts_index` with defaults. Although some named parameters are defined as `BOOL`, you cannot pass `TRUE` or `FALSE` when invoking the PRAGMA. The full example does not use non-default values for the named parameters `strip_accents`, `lower` or `overwrite`, so there is no issue there. However, when adapting the full example to set `overwrite` to a non-default value (`TRUE`) an error occurs:

`Error: Parser Error: Named parameter requires a constant on the RHS`

This error does not occur when passing `1` instead of `TRUE` and `0` instead of `FALSE`, which should be reflected in the documentation. I'm unsure whether this is something that can also be addressed upstream such that you _can_ pass `TRUE` or `FALSE`.

### DIFF
--- a/docs/extensions/full_text_search.md
+++ b/docs/extensions/full_text_search.md
@@ -11,7 +11,7 @@ The extension adds two `PRAGMA` statements to DuckDB: one to create, and one to 
 ### PRAGMA create_fts_index
 ```
 create_fts_index(input_table, input_id, *input_values, stemmer='porter', stopwords='english',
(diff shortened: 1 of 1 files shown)


Alex-Monahan:
Thank you for improving our docs! If you would like to see this fixed within DuckDB itself, could you add an issue to the main repo? That will ask you for some additional details and all of that.

Much appreciated!

PR #648, opened 2023-03-21
Iframe shell.duckdb.org on wasm.md page

carlopi:


Alex-Monahan:
Great idea! Thanks!

I think we may want to inject the shell in a number of places at some point.

PR #649, opened 2023-03-22
Add documentation of GCD and LCM functions

kryonix:
The functions GCD and LCM have been added to duckdb. This pr adds them to the documentation as well.


Alex-Monahan:
Fantastic, thank you! Would you mind also adding any alias names that you created? We like to list those in the docs also.

kryonix:
Sure thing, makes sense! I've added them.

Alex-Monahan:
Thanks again! Looks great. Those will come in handy I'm sure!

PR #657, opened 2023-03-23
Issue #3423: Document POSITIONAL JOIN

hawkfish:
And other joins too.
Also, fix railroad for CROSS JOIN.


docs/sql/query_syntax/from.md
```diff
@@ -28,6 +28,129 @@ SELECT * FROM table_name TABLESAMPLE 10%;
...
+FROM west s1, west s2
+WHERE s1.time > s2.time 
+  AND s1.cost < s2.cost;
+```
+
+#### Positional Joins
+
+When working with data frames or other embedded tables of the same size, 
+the rows may have a natural correspondence based on their physical order.
+In scripting languages, this is easily expressed using a loop:
```
Mause, line 101: Not sure I'd call C++ a scripting language!

PR #659, opened 2023-03-26
Add function list, to explicitly add `parquet_scan` as alias

### LINKED ISSUE #589: Improve documentation of difference between read_parquet and parquet_scan or state that they are identical
See [this page](https://duckdb.org/docs/data/parquet.html#filenames).

image

It should be explicitly stated that `read_parquet` and `parquet_scan` are identical - if they are.

Tishj:
This PR fixes #589


Mause:
Thanks!

PR #674, opened 2023-04-04
add date function examples with a variable interval

OrangieLou:
I noticed it was not possible to simply add or subtract variable intervals from a date. I had imagined something like this:
```
SELECT DATE '1992-03-22' + INTERVAL d.days DAY FROM (VALUES (5), (11)) AS d(days);
```
But this gives me a Parser Error. In the comments of a benchmark query I found something that tipped me off with this workaround:
```
SELECT DATE '1992-03-22' + INTERVAL 1 DAY * d.days FROM (VALUES (5), (11)) AS d(days)
```
I thought it a nice addition to the docs as this workaround was not obvious to me.


Alex-Monahan:
Great idea, thank you! I also remember this being confusing to me when I first saw it (on Teradata of all places...). Much appreciated!

PR #677, opened 2023-04-09
Add DuckDB Wasm example from repository

bluehat974:
Add DuckDB wasm examples (from duckdb-wasm repository)
https://github.com/duckdb/duckdb-wasm/blob/master/packages/duckdb-wasm/README.md

There is also more examples added (convert arrow table to JSON and export parquet)


carlopi:
Thanks!
I think splitting the Wasm section in sub-sections is the way to go, and this seems a solid direction.
I would probably have wasm/overview.md be just a copy of old wasm.md, until that content do not find an alternative place.

The other pages probably would need a bit more text, at the very least on top, and potentially we could add some other sections (even later) like:
* how to embed duckdb-wasm, which files, how to serve them, example of deployments
* security would be nice to mention somewhere (basically that data processing happens on-device with no logging)

bluehat974:
@carlopi 
I rollbacked to old wasm.md content and added small introduction for each pages in wasm folder.

For other ideas (embed duckdb-wasm or security), I need to think more and probably do it in another PR and do incremental develoment (there is so much knowledge I wish to copy/paste from discord or the unit test).

docs/api/wasm/extensions.md
```diff
@@ -0,0 +1,11 @@
+---
+layout: docu
+title: Extensions
+selected: Client APIs
+---
+
+Default [extensions](../../extensions/overview) enabled in DuckDB-wasm are HTTPFS and Parquet.
```
carlopi, line 7: I think also json and fts (= full text search) are packaged by default.

docs/api/wasm/data_ingestion.md
```diff
@@ -0,0 +1,149 @@
+---
+layout: docu
+title: Data Ingestion
+selected: Client APIs
+---
+
+DuckDB-wasm has multiple way to import data. 
+
+Basically, there are two steps to import data into DuckDB.
```
carlopi, line 9:
```suggestion
DuckDB-Wasm has multiple way to import data.

There are two steps to import data into DuckDB.
```

carlopi (approved):
Left another round with a few minor suggestions, but looks good to go after that.

I think on the overview.md it really needs a rework (as you suggested initially), but probably it makes more sense if I try to come up with a new structure and then we iterate starting from that.

Thanks a lot on this contribution, potentially if there are item that you think are worth inserting but don't have a place yet a possible way to keep track of them is opening an issue about DuckDB-Wasm documentation. Thanks again.

docs/api/wasm/extensions.md
carlopi: Actually, correcting myself, Parquet, FTS are alwasy packaged by default, while JSON and Excel are build-time opt-in. See https://github.com/duckdb/duckdb-wasm/blob/968ba722350d9be3988d7f44ac328c0ef0c8423a/lib/CMakeLists.txt#L238, I believe JSON and Excel

HTTPFS it's not really an extension, but a different re-implementation of it that comes bundled by default.

docs/api/wasm/overview.md
```diff
@@ -1,19 +1,23 @@
...
+
+## Getting Started with DuckDB-Wasm
+
+A great starting point is to read the [DuckDB-Wasm launch blog post](https://duckdb.org/2021/10/29/duckdb-wasm.html)!
+
+Another great resource is the [Github repository](https://github.com/duckdb/duckdb-wasm).
+
+For details, see the full [DuckDB-Wasm API Documentation](https://shell.duckdb.org/docs/modules/index.html).
+
+_DuckDB-Wasm is fast! If you're here for performance numbers, head over to [our benchmarks](https://shell.duckdb.org/versus)._
```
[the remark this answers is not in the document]
bluehat974: I liked this url as it compare different browser database tools (sql.js, arquero...)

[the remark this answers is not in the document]
carlopi: Makes also sense. +1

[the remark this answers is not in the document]
bluehat974: Removed

docs/api/wasm/extensions.md
bluehat974: File changed

Alex-Monahan:
Thank you for your contribution! This will be a real help to folks.

PR #698, opened 2023-04-20
Add extra clarification for intervals and their use cases

Mytherin:
Clarify the interval type and describe the use cases more clearly, plus add a reference to datediff (see duckdb/duckdb#7170)


Alex-Monahan:
Looks great! I did not know about the `interval (i) year` trick... Here I've been multiplying an interval by a column! Good trick to know...

PR #717, opened 2023-04-25
add spatial blogpost

Maxxen:
Adds a post about the spatial extension, largely adapted from the extension readme.

Thoughts:
- Tried having some fun with the title but not sure if it is too much or to unclear?
- I could elaborate more on _why_ we chose to go our own way and not base everything on e.g. Googles S2 or PostGIS's LWGEOM, but I'm not sure if that's too technical or relevant here.
- Not sure how much I should assume readers are familiar with geospatial SQL in general and concepts like coordinate projections or spatial joins. But I think I have somewhat of a balance with some explanations hidden in collapsible segments, what y'all think?

Note that I haven't uploaded binaries so that they are downloadable from the CLI yet, will tackle that tomorrow, and maybe add a link/example once that's done.


Mytherin:
The blog post looks great! Perhaps it also makes sense to have some documentation ready for people that want to try the extension after reading the blog post?

PR #718, opened 2023-04-26
Document constraints and expected index behavior

Mytherin:


Alex-Monahan:
Looks great, super helpful! Thanks as always!

PR #719, opened 2023-04-26
Add documentation or regexp_matches examples.

Tmonster:
When this PR merges https://github.com/duckdb/duckdb/pull/7264, we should probably update the docs.

This is also to show users that the wildcard token (i.e `.`) does not match newline by default.


hannes:
Can you resolve the conflict please?

hannes:
thanks

PR #744, opened 2023-05-12
added support for ** glob

lverdoes:
Added some examples of how to use `**` for csv, parquet.
And added a row to the wildcard table


Mause:
Thanks!

PR #760, opened 2023-05-15
Add MultiFileReader params to JSON docs

lnkuiper:
Forgot these in the previous PR


carlopi:
Hi! Was reading around the docs, unsure if also these bits: https://duckdb.org/docs/sql/statements/copy#json-options potentially also needs updating?

lnkuiper:
@carlopi You're right, thanks! I've updated the copy docs too

PR #761, opened 2023-05-15
Update docs with semi and anti joins

Tmonster:
Anti and semi joins are in 0.8.0 but the docs have not yet been updated. 

https://github.com/duckdb/duckdb/pull/6480

### DIFF
--- a/docs/sql/query_syntax/from.md
+++ b/docs/sql/query_syntax/from.md
@@ -96,6 +96,25 @@ WHERE s1.time > s2.time
   AND s1.cost < s2.cost;
 ```
 
+#### Semi and Anti Joins
+
+Semi joins allow a user to return rows from the left table that have at last one match in the right table. Anti joins allow a user to return rows from the left table that have _no_ matches in the right table. When using a semi or anti join the resulting table will never have more rows than the left hand side table. Semi anti joins provide the same logic as [(NOT) IN](../expressions/in) statements.
+
+```sql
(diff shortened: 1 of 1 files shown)


docs/sql/query_syntax/from.md
```diff
@@ -96,6 +96,25 @@ WHERE s1.time > s2.time
   AND s1.cost < s2.cost;
 ```
 
+#### Semi and Anti Joins
+
+Semi joins return rows from the left table that have at last one match in the right table. Anti joins return rows from the left table that have _no_ matches in the right table. When using a semi or anti join the result  will never have more rows than the left hand side table. Semi and anti joins provide the same logic as [(NOT) IN](../expressions/in) statements.
```
Mause, line 101:
```suggestion
Semi joins return rows from the left table that have at last one match in the right table. Anti joins return rows from the left table that have _no_ matches in the right table. When using a semi or anti join the result will never have more rows than the left hand side table. Semi and anti joins provide the same logic as [(NOT) IN](../expressions/in) statements.
```

Mause (changes requested):
Just a small typo

Alex-Monahan:
Wow, yet another case of TIL. I love the non-duplicating feature of this. 

I have a quick question for you @Tmonster, for my own understanding. Will a NOT IN list get re-written into one of these joins, or the other way around? ;-)

PR #764, opened 2023-05-15
Add documentation for Pythons `create_function`

Tishj:
Adds documentation for `create_function` and `remove_function`

### DIFF
--- a/docs/api/python/function.md
+++ b/docs/api/python/function.md
@@ -1,0 +1,135 @@
+---
+layout: docu
+title: Python Function API
+selected: Client APIs
+---
+
+You can create a DuckDB function out of a python function so it can be used in SQL queries.
+Just like regular [functions](../../sql/functions/overview.md) they need to have a name, a return type and parameter types.
+
+### Creating Functions
+
+The `create_function` method is used to add a function, the parameters to this method are:  
+
+`name: str`  
+The name that will be used to identify the function.  
+  
+`function: Callable`  
+The python function that will be used.
+
+`argument_type_list: Optional[List]`  
+The types of the arguments that the function expects.
+
+`return_type: Optional[DuckDBPyType]`  
+The type of the value returned from the function.
+
+`type: Optional[str]`  
+The type of python function, either `native` or `arrow`.
+
+`null_handling: Optional[str]`  
+The way NULL value parameters should be handled.
+
+`exception_handling: Optional[str]`  
+The way exceptions that occurred during execution of the function should be handled.
+
+### Type Annotation
+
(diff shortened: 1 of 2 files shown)


docs/api/python/function.md
```diff
@@ -0,0 +1,137 @@
+---
+layout: docu
+title: Python Function API
+selected: Client APIs
+---
+
+You can create a DuckDB function out of a python function so it can be used in SQL queries.
+Just like regular [functions](../../sql/functions/overview.md) they need to have a name, a return type and parameter types.
```
Mause, line 8:
```suggestion
Just like regular [functions](../../sql/functions/overview) they need to have a name, a return type and parameter types.
```

docs/api/python/function.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Creating Functions
+
+The `create_function` method is used to add a function, the parameters to this method are:  
+
+`name: str`
```
Mause, line 14: This section can probably just be a link to the API reference instead?

Mause (changes requested):

docs/api/python/function.md
Tishj: I don't think that's descriptive enough.
https://duckdb.org/docs/api/python/reference/#duckdb.DuckDBPyConnection.create_function

docs/api/python/function.md
```diff
@@ -0,0 +1,137 @@
+---
+layout: docu
+title: Python Function API
+selected: Client APIs
+---
+
+You can create a DuckDB function out of a python function so it can be used in SQL queries.
+Just like regular [functions](../../sql/functions/overview) they need to have a name, a return type and parameter types.
+
+### Creating Functions
```
Mytherin, line 10: Could we start off with an example at the top similar to what we do on other pages? Often that's what people are looking for in the docs.

Tishj: I've added another example using faker at the top :)

docs/api/python/function.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Null Handling
+By default when functions receive a NULL value, this instantly returns NULL, as part of the default null handling.  
+When this is not desired, you need to explicitly set this parameter to `'special'`.
+
+```py
+def dont_intercept_null(x):
+	return 5
+
+duckdb.register_scalar_udf('dont_intercept', dont_intercept_null, [BIGINT], BIGINT)
```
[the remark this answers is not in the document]
Tishj: Yea you're right, I've fixed and verified all of the examples now

PR #773, opened 2023-05-16
Add docs for Building DuckDB

### LINKED ISSUE #756: Add documentation for installation from source
Add documentation for using the Makefile

Should include:
Extension define list: BUILD_TPCH, BUILD_TPCDS, BUILD_PYTHON etc..

Debug flags:
DEBUG_NO_INLINE, etc..

Build types:
release, reldebug, relassert, debug, etc..

Tishj:
This PR fixes #756 

It doesn't contain every option, a lot of the debug / miscellaneous flags are not documented as they seemed too esoteric.

### DIFF
--- a/dev/building.md
+++ b/dev/building.md
@@ -109,7 +109,8 @@ When this flag is set, the [SQLSmith](https://github.com/duckdb/duckdb/pull/3410
 ### Debug Flags
 
 #### `CRASH_ON_ASSERT`
-When an assertion triggers (`D_ASSERT(condition);` in the code), this throws an InternalException. With this flag enabled it instead directly causes a crash.
+`D_ASSERT(condition)` is used all throughout the code, these will throw an InternalException in debug builds.  
+With this flag enabled, when the assertion triggers it will instead directly cause a crash.
 
 #### `DISABLE_STRING_INLINE`
 In our execution format `string_t` has the feature to "inline" strings that are under a certain length (12 bytes), this means they don't require a separate allocation.  
--- a/dev/building.md
+++ b/dev/building.md
@@ -48,7 +48,7 @@ BUILD_PYTHON=1 make debug
 When this flag is set, the [Python](../docs/api/python/overview.md) package is built.
 
 #### `BUILD_SHELL`
-When this flag is set, the [CLI](../docs/api/cli.md) is built, this is usually enabled by default.
+When this flag is set, the [CLI](../docs/api/cli) is built, this is usually enabled by default.
 
 #### `BUILD_BENCHMARK`
 When this flag is set, our in-house Benchmark testing suite is built.
--- a/dev/building.md
+++ b/dev/building.md
@@ -117,7 +117,10 @@ In our execution format `string_t` has the feature to "inline" strings that are 
 When this flag is set, we disable this and don't inline small strings.
 
 #### `DISABLE_MEMORY_SAFETY`
-Our data structures that are used extensively throughout the non-performance-critical code have extra checks to ensure memory safety, including preventing dereferencing `nullptr`, or index out of bounds errors.  
+Our data structures that are used extensively throughout the non-performance-critical code have extra checks to ensure memory safety, these checks include:  
+- Making sure `nullptr` is never dereferenced.  
+- Making sure index out of bounds accesses don't trigger a crash.  
+
 With this flag enabled we remove these checks, this is mostly done to check that the performance hit of these checks is negligible.
 
 #### `DESTROY_UNPINNED_BLOCKS`
--- a/dev/building.md
+++ b/dev/building.md
@@ -8,6 +8,9 @@ expanded: Development
 To build DuckDB we use a Makefile which in turn calls into CMake.  
 It's not advised to directly call CMake, as the Makefile sets certain variables that are crucial to properly building the package.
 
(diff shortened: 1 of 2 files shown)


dev/building.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### `benchmark`
+This build is a shorthand for `release` with `BUILD_BENCHMARK=1` set.
+
+#### `tidy-check`
+This creates a build and then runs [clang tidy](https://clang.llvm.org/extra/clang-tidy/) to check for issues or style violations through static analysis.  
+The CI will also run this check, causing it to fail if this check fails.
+
+#### `format-fix` | `format-master`
+This doesn't actually create a build, but uses [clang-format](https://clang.llvm.org/docs/ClangFormat.html) to check for format issues in the code.
```
Mause, line 36: It also runs cmake-format too right?

dev/building.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### `release`
+This build has been stripped of all the assertions and debug symbols and code, optimized for performance.
+
+#### `debug`
+This build runs with all the debug information, including symbols, assertions and DEBUG blocks.  
+The special debug defines are not automatically set for this build however.
+
+#### `relassert`
+This build does not trigger the `#ifdef DEBUG` code blocks, but still has debug symbols that make it possible to step through the execution with line number  information and `D_ASSERT` lines are still checked in this build.
```
Mause, line 23:
```suggestion
This build does not trigger the `#ifdef DEBUG` code blocks, but still has debug symbols that make it possible to step through the execution with line number information and `D_ASSERT` lines are still checked in this build.
```

dev/building.md
```diff
@@ -0,0 +1,133 @@
...
+### Package Flags
+
+For every package that is maintained by core DuckDB, there exists a flag in the Makefile to enable building the package.  
+These can be enabled by either setting them in the current `env`, through set up files like `bashrc` or `zshrc`, or by setting them before the call to `make`, for example:
+```bash
+BUILD_PYTHON=1 make debug
+```
+
+#### `BUILD_PYTHON`
+When this flag is set, the [Python](../docs/api/python/overview.md) package is built.
```
Mause, line 48:
```suggestion
When this flag is set, the [Python](../docs/api/python/overview) package is built.
```

dev/building.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### `BUILD_SHELL`
+When this flag is set, the [CLI](../docs/api/cli.md) is built, this is usually enabled by default.
```
Mause, line 51:
```suggestion
When this flag is set, the [CLI](../docs/api/cli) is built, this is usually enabled by default.
```

dev/building.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### `BUILD_BENCHMARK`
+When this flag is set, our in-house Benchmark testing suite is built.
+More information about this can be found [here](https://github.com/duckdb/duckdb/tree/master/benchmark).
+
+#### `BUILD_JDBC`
+When this flag is set, the `[Java](../docs/api/java.md) package is built.
```
Mause, line 58:
```suggestion
When this flag is set, the `[Java](../docs/api/java) package is built.
```

Mause (changes requested):
Would probably be worth mentioning ninja, and perhaps the `CMakeUserPresets.json` file?

dev/building.md
Tishj: Okay I'll verify and update them

dev/building.md
Tishj: I don't know anything about `CMakeUserPresets.json`, and my duckdb repo doesn't contain it, so I'm not entirely sure what you want me to say about it here.

Alex-Monahan:
I just received a Mac for the new gig, so I'll be able to test these instructions out as a complete C++ novice! That one college class was a little while ago... ;-)

Alex-Monahan:
Hey @Tishj and @Mause! I'm getting started trying to compile DuckDB, but I'm wondering if we may want some additional steps at the very start. The readme in the main repo mentions that cmake, python3, and a C++11 compiler are required. Would it be possible to document those steps also? I know they are unnecessary for someone already using C++ already, but it could help newcomers. I tried to build up a set of steps for Windows and it proved too tricky, so maybe we could just have Mac and Linux instructions? 

Even if we don't want to officially document it, are there best practices that you recommend for those steps? On Mac, should I just use homebrew for everything?

Alex-Monahan:
For Mac, I think the steps could be as simple as: 

1. Install python from Python.org
2. Install homebrew
3. brew install cmake
4. brew install ninja
5. Go to ~/.zshrc and add an environment variable to use ninja
	1. export GEN="ninja"
6. Clone the repo 
7. Run make in the top level directory

carlopi:
I think you only python, cmake, and a C++ compiler (unsure what's the equivalent package to linux's build-utils).
I would keep ninja / changing settings as optional, to keep complexity low.

And one thing I noticed with Pedro trying duckdb-wasm is that it would be cool if running make will test and potentially say "you miss python!, please follow instructions at link" (at least for basic dependencies)

Mause:
I think for now we just merge this and iterate on it?

PR #774, opened 2023-05-16
Improve Python docs for Connection

### LINKED ISSUE #643: Improve Python-API `DuckDBPyConnection` documentation
We have switched to creating duplicate wrapper methods for all connection methods, to use the default connection.
I don't think we have any of this documented, and it can cause confusion when the methods are used interchangeably (`con.sql` and `duckdb.sql` for example)

Tishj:
This PR fixes #643 

Mention `':default:'` and mention that the methods on `duckdb` use this connection by default.

### DIFF
--- a/docs/api/python/dbapi.md
+++ b/docs/api/python/dbapi.md
@@ -9,7 +9,7 @@ The standard DuckDB Python API provides a SQL interface compliant with the [DB-A
 ### Connection
 
 To use the module, you must first create a `DuckDBPyConnection` object that represents the database.
- The connection object takes as parameter the database file to read and write from. If the database file does not exist, it will be created (the file extension may be `.db`, `.duckdb`, or anything else). The special value `:memory:` (the default) can be used to create an **in-memory database**. Note that for an in-memory database no data is persisted to disk (i.e. all data is lost when you exit the Python process). If you would like to connect to an existing database in read-only mode, you can set the `read_only` flag to `True`. Read-only mode is required if multiple Python processes want to access the same database file at the same time.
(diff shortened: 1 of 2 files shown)


docs/api/python/dbapi.md
```diff
@@ -6,8 +6,27 @@ selected: Client APIs
...
+### Connection
+
+To use the module, you must first create a `DuckDBPyConnection` object that represents the database.
+ The connection object takes as parameter the database file to read and write from. If the database file does not exist, it will be created (the file extension may be `.db`, `.duckdb`, or anything else). The special value `:memory:` (the default) can be used to create an **in-memory database**. Note that for an in-memory database no data is persisted to disk (i.e. all data is lost when you exit the Python process). If you would like to connect to an existing database in read-only mode, you can set the `read_only` flag to `True`. Read-only mode is required if multiple Python processes want to access the same database file at the same time.
```
Mause, line 12:
```suggestion
 The connection object takes as a parameter the database file to read and write from. If the database file does not exist, it will be created (the file extension may be `.db`, `.duckdb`, or anything else). The special value `:memory:` (the default) can be used to create an **in-memory database**. Note that for an in-memory database no data is persisted to disk (i.e. all data is lost when you exit the Python process). If you would like to connect to an existing database in read-only mode, you can set the `read_only` flag to `True`. Read-only mode is required if multiple Python processes want to access the same database file at the same time.
```

docs/guides/python/jupyter.md
```diff
@@ -45,6 +45,7 @@ pip install matplotlib
 
 ## Library Import and Configuration
 Next, open a Jupyter Notebook and import the relevant libraries. 
+It's possible to have `%sql` commands and `duckdb.sql` share the same [default connection](dbapi.md) by providing `duckdb:///:default:` as the SQLAlchemy connection string.
```
Mause, line 48:
```suggestion
It's possible to have `%sql` commands and `duckdb.sql` share the same [default connection](dbapi) by providing `duckdb:///:default:` as the SQLAlchemy connection string.

```

Mause (changes requested):

Mause:
Thanks!

PR #775, opened 2023-05-16
Add `Vector` docs

### LINKED ISSUE #766: Add documentation for `Vector`
I feel like this is needed.

Since we have many different clients, that all have their own specialized API which can sometimes go pretty in depth, I think it's worth it to have a general place where information about our execution format can live.

This can also include explanations about the `STANDARD_VECTOR_SIZE` constant, with a brief explanation of the vectorized execution engine model.
Currently when you search for `Vector` on the docs the only relevant hit is the C++ DataChunk API.

This can also include information about the various different vector formats we use, and serve as a nice entry point for new contributors as well.

Tishj:
This PR fixes #766 

Maybe this could still use a special mention of the `ArrowAuxiliaryData`

### DIFF
--- a/internals/vector.md
+++ b/internals/vector.md
@@ -1,0 +1,117 @@
+---
+layout: docu
+title: Execution Format
+---
+
+`Vector` is the container format used to store in-memory data during execution.  
+`DataChunk` is a collection of Vectors, used for instance to represent a column list in a PhysicalProjection operator.
+
+### Data Flow
+
+DuckDB uses a vectorized query execution model. This means that our operators are optimized to work on Vectors of a fixed size.
+This fized size is commonly referred to in the code as `STANDARD_VECTOR_SIZE`.  
+The default STANDARD_VECTOR_SIZE is 2048 tuples.
+
+### Internals
+
+`ValidityMask validity`  
+Tracks which elements in the Vector are NULL.  
+
+`data_ptr_t data`  
+The element data of the Vector.  
+
+`VectorBuffer buffer`  
+An interface implemented based on the type of the vector and the data type it's representing.  
+Usually owns the `data` memory.  
+
+`VectorBuffer auxiliary`  
+Optionally stores additional data needed by the Vector, explained later.  
+
+
+### Vector Format
+
+To efficiently represent different shapes of collections, we use different Vector formats.
+
+#### FLAT_VECTOR
+
+Flat vector is a sequential array, this is the default format.
+
+#### CONSTANT_VECTOR
+
+Constant vector is used to represent a single element as an entire vector.  
+For example:  
+```sql
+select lst, 'duckdb' from range(1000) tbl(lst);
+```
+Since `duckdb` is Constant Expression, we know it's a single entry, but operators expect a DataChunk to have the same amount of tuples.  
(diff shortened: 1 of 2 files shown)


internals/vector.md
```diff
@@ -0,0 +1,117 @@
...
+#### STRUCT
+Struct is designed to function as a nested table, with as little overhead as possible.  
+For structs, the `auxiliary` is used to store a list of "child" Vectors.  
+The `data` and `buffer` variables are unused by a struct Vector.
+
+#### MAP
+Internally `MAP` is just a `LIST[STRUCT(key KEY_TYPE, value VALUE_TYPE)]`.
+
+#### UNION
+Internally `UNION` utilizes the same structure as a `STRUCT`.
```
Mause, line 94: Might be worth mentioning the union tag here too?

Tishj: Thanks, I've added it :)

Mause:
Looks good, thanks!

internals/vector.md
```diff
@@ -0,0 +1,118 @@
...
+Instead we can use a ConstantVector to only store 'duckdb' once.
+
+#### DICTIONARY_VECTOR
+
+Dictionary vector is used when a transformation would not change any elements of the data, but only change the order of the elements.  
+For example:  
+```sql
+select * from range(1000) order by 1 desc;
+```
+Instead of duplicating the result of the `range` result and sorting that, we can use the same data, but just represent it in a different shape.
```
Mytherin, line 57: Language nitpick - can we remove all instances of `just` in the text?

Mytherin (commented):
Thanks for the PR! Some comments

internals/vector.md
Tishj: Yea I agree, re-reading it it sounds sloppy

internals/vector.md
```diff
@@ -0,0 +1,118 @@
...
+```
+Instead of duplicating the result of the `range` result and sorting that, we can use the same data, but just represent it in a different shape.  
+How the dictionary vector works internally is that it adds a layer of indirection between the requested index and the retrieved index.  
+In the example above a request for the data at index 0 would instead retrieve the data at index 999.
+
+#### SEQUENCE_VECTOR
+
+Sequence vectors are used when the data is a sequence with a fixed increment.  
+Instead of storing all the elements, we just store the beginning and the increment.
+
```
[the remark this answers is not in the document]
Tishj: Isn't this defined by the API of the language used to interact with the Vector?
My goal for this page was to provide information about the internals of the Vector without including API details, do you think that's okay?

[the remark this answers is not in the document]
Mytherin: Yes, fair enough, that's fine

internals/vector.md
```diff
@@ -0,0 +1,118 @@
...
+			char *ptr;
+		} pointer;
+		struct {
+			uint32_t length;
+			char inlined[12];
+		} inlined;
+	} value;
+};
+```
+The details of this class and the benefits are explained in greater detail [here](https://github.com/duckdb/duckdb/pull/431).
```
Mytherin, line 115: Forgot to mention - can we expand the description here instead of linking to a PR, or add a separate page explaining our string representation?

PR #782, opened 2023-05-17
Fix broken links

carlopi:


docs/api/python/types.md
```diff
@@ -4,7 +4,7 @@ title: Types API
 selected: Client APIs
 ---
 
-The `DuckDBPyType` class represents a type instance of our [data types](../../sql/data_types/overview.md)
+The `DuckDBPyType` class represents a type instance of our [data types](/docs/sql/data_types/overview)
```
Mause, line 7: These should be kept as relative links for when they are archived

Mause (changes requested):

PR #784, opened 2023-05-17
Document unnesting of structs and recursive unnesting

Mytherin:


docs/sql/query_syntax/unnest.md
```diff
@@ -5,15 +5,22 @@ selected: Documentation/SQL/Query Syntax/UNNEST
...
+-- unnest a list, generating 3 rows (1, 2, 3)
+SELECT UNNEST([1, 2, 3]);
+-- unnesting a struct, generating two columns (a, b)
+SELECT UNNEST({'a': 42, 'b': 84});
+-- recursive unnest of a list of structs
+SELECT UNNEST([{'a': 42, 'b': 84}, {'a': 100, 'b': NULL}], recursive := True);
+```
 
-When `UNNEST` is combined with regular scalar expressions, those expressions are repeated for every entry in the list. When multiple lists are unnested in the same `SELECT` clause, the lists are unnested side-by-side. If one list is longer than the other, the shorter list will be padded with `NULL` values.
+The `UNNEST` function is used to unnest lists or structs by one level. The function can be used as a regular scalar function, but only in the `SELECT` clause. `UNNEST` with the `recursive` parameter will unnest lists and structs of multiple levels.
```
Mause, line 18: Small nitpick - might be worth mentioning that you can't unnest maps?

PR #786, opened 2023-05-17
Follow up PR on Vector docs, implementing feedback

Tishj:
@Mytherin feel free to continue the review here, I tried to hit all the feedback you gave.

I replied in the old PR, but I'll repost it here:
My intention for this page was to provide information about the internals of Vector, without going into detail on the API side of things, as those are package specific.

This might also be missing some specific mentions of the compressed execution you requested, I didn't really find a natural way to fit it in, but I've sprinkled it in here and there when explaining the different vector types.


Mytherin:
Did another pass over the text and cleaned it up a bit. In particular I removed the `Internals` as this is C++ specific.

Tishj:
I did like the internals section but I understand.
Maybe I'll replicate it somewhere under the C++ API tab

PR #791, opened 2023-05-19
Clarify when conflict_target is optional

### LINKED ISSUE #600: Improve UPSERT docs
Add that `conflict_target` is not optional for DO UPDATE - if there are more than one unique/primary key constraints on the table.

Alex-Monahan:
Should fix #600.
@Tishj , mind having a quick look?


Tishj:
Looks good to me :+1: Thanks!

PR #796, opened 2023-05-23
Document hidden syntax for CASE and IFNULL

jkew:
This was implemented in DuckDB in:
https://github.com/duckdb/duckdb/issues/156

I verified locally that the documentation looks right, but given that this is my first docs PR I might have missed something.


Alex-Monahan:
Thanks for this! I requested a small change (to use 'default_string' instead of 'NULL' in your examples to remove any potential for confusion), however it looks like you may need to approve that change in some way and update this PR? Sorry, in the past I have just modified PR's and they have been mergeable afterwards!

jkew:
> 



> Thanks for this! I requested a small change (to use 'default_string' instead of 'NULL' in your examples to remove any potential for confusion), however it looks like you may need to approve that change in some way and update this PR? Sorry, in the past I have just modified PR's and they have been mergeable afterwards!

I think I needed to sync my fork? I don't know, but it now looks like your change is in this PR. Thanks for taking a look!

PR #800, opened 2023-05-26
Add correlated subqueries blog post

Mytherin:


_posts/2023-05-26-correlated-subqueries-in-sql.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+We can see the drastic performance difference that subquery decorrelation has when we compare the run-time of this query in DuckDB with the run-time in Postgres and SQLite. When running the above query on the [`ontime` dataset](https://www.transtats.bts.gov/Homepage.asp) for `2017` with roughly `~4 million` rows, we get the following performance results:
+
+| DuckDB | Postgres | SQLite |
+|--------|----------|--------|
+| 0.06s  | >48 Hours    | >48 Hours  |
+
+
+As Postgres and SQLite do not de-correlate the subquery, the query is not just *logically*, but *actually* executed once for every row. As the subquery involves a full table scan, this moves the query from linear complexity, O(n), to quadratic complexity, O(n<sup>2</sup>).
```
lnkuiper, line 266: Maybe here we can add something along the lines of:
"Roughly speaking, this query will take 4 million times longer in Postgres/SQLite than in DuckDB."
I don't think the big O notation really resonates with everyone, but the number 4 million does.

_posts/2023-05-26-correlated-subqueries-in-sql.md
```diff
@@ -0,0 +1,294 @@
...
+     FROM ontime
+     WHERE ontime.origin=ontime_outer.origin AND ontime.dest=ontime_outer.dest
+     GROUP BY uniquecarrier
+     HAVING COUNT(*) > 1000
+);
+```
+
+#### Performance
+
+Logically, correlated subqueries are executed *once per row*. As such, it is natural to think that correlated subqueries are very expensive and should be avoided for performance reasons.
```
lnkuiper, line 222: Maybe to make the difference between regular scalar subqueries and correlated subqueries more clear, we can add something along the line of:
"Whereas scalar subqueries are logically executed *once*, correlated subqueries are logically executed *once per row*. Slight modifications to a query that may seem harmless, such as adding a `WHERE` clause that compares a column from a subquery with a column from the outer query, may have significant performance implications."

_posts/2023-05-26-correlated-subqueries-in-sql.md
```diff
@@ -0,0 +1,294 @@
+---
+layout: post
+title:  "(Correlated) Subqueries in SQL"
```
lnkuiper, line 3: Maybe something along the lines of:
"(Correlated) Subqueries in SQL or: How I changed one line in my query and now it has been running for 48 hours"

lnkuiper (approved):
I love the blogpost, and I left some comments. Hope they're helpful!

_posts/2023-05-26-correlated-subqueries-in-sql.md
lnkuiper: Kind of like "Dr. Strangelove or: How I Learned to Stop Worrying and Love the Bomb"

PR #806, opened 2023-05-27
epoch_xs and make_timestamp(µs)

hawkfish:
Document new high-precision epoch conversion functions.


Alex-Monahan:
Thanks as always @hawkfish! Is it worth mentioning that DuckDB maxes out at microsecond precision in the nanosecond functions?

hawkfish:
Yeah I'm actually not entirely sure what those types do They seem to get cast at some point.

Issue #808, opened 2023-05-30
Add `con.close()` to persistent storage section in Python Overview page


ned2:
I was referring to the [persistent storage section](https://duckdb.org/docs/api/python/overview.html#persistent-storage) of the Python Overview page to work out how to save my DB to disk. This page doesn't mention that there are some circumstances where you need to explicitly close the connection, so it took me quite a while to work out that the reason my saved tables weren't appearing in the catalogue of the saved DB was because I hadn't explicitly closed it. I see that this is documented over on [this docs page](https://duckdb.org/docs/api/python/dbapi#connection), but I only found this after struggling with this problem for a while. 

It would be good to add reference to the need to close the DB in some circumstances. I was thinking maybe just a final line in the code snippet with `con.close()` and a comment indicating when this is explicitly needed.

Would be happy to submit a PR.

szarnyasg:
Fixed by #812

PR #812, opened 2023-05-31
Add con.close() to Persistent Storage example in Python Overview

ned2:
This adds an explicit `con.close()` to the Persistent Storage example in the Python API Overview page.

As described in #808, readers may come away not being aware that an explicit connection close is required in some situations. A notable context is in a Jupyter notebook, where it is common for everything to be in the global scope, so no implicit close will occur (as was my context when it took me some time to work out why expected tables weren't available in my previously saved database).


ned2:
heya @Alex-Monahan, what do you think about this tweak to the docs?

Alex-Monahan:
Looks great, thanks! Sorry, I was on vacation. Back to "the office" for a few days now... ;-)

PR #816, opened 2023-06-01
Replace "high" with "low" in claim about ENUM data types and string column cardinality.

gordonje:
Hello,

I am submitting what I believe to be a valid correction to the section of your docs describing the `ENUM` data type.

I also confess I may not completely understand your intended meaning and how it relates to the distinction between high- and low-cardinality I found in [Wikipedia](https://en.wikipedia.org/wiki/Cardinality_(SQL_statements)#Values_of_cardinality).

Thank you for your amazing project. I use DuckDb everyday and evangelize about it whenever I have the opportunity.

### DIFF
--- a/docs/sql/data_types/enum.md
+++ b/docs/sql/data_types/enum.md
@@ -11,7 +11,7 @@ blurb: The ENUM type represents a dictionary data structure with all possible un
 
 ## Enums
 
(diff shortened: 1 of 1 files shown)


Alex-Monahan:
Good find, thank you!

PR #818, opened 2023-06-03
Clarify file output in profiling.md

NickCrews:
Fixes https://github.com/duckdb/duckdb/issues/7704


NickCrews:
Please double check my interpretation of what actually happens here.

Perhaps should edit https://duckdb.org/docs/sql/pragmas as well.

PR #829, opened 2023-06-13
DuckDB v0.8.1

carlopi:
Bumping to v0.8.1, I changed:
* _config.yml
* _data/versions.csv
* `python3 scripts/archive_docs.py 0.8.1`

I verified locally (`bundler exec jekyll serve`), seems to be working as intended.


Mause (changes requested):
I think we should try to close some of the docs issues before we do this, given how many there are open

carlopi:
Filed #830 moving installation links to 0.8.1, but avoid duplications of docs (for now)

PR #831, opened 2023-06-14
Revert PR #809 on list_intersect & all

carlopi:
PR was merged against feature, so should wait a bit more to go in.
This fixes https://github.com/duckdb/duckdb/issues/7936

This commit will have to be reverted once dev docs will map to feature.


carlopi:
Potentially adding a "feature" label to PRs could more easily allow to differentiate between what should be merged right now, and what should go in later. Or having a 'feature' branch also here, unsure what's more complex.

PR #834, opened 2023-06-15
chore: update information_schema to note catalog is now supported

Mause:
Originally noted in https://github.com/tconbeer/harlequin/issues/89


tconbeer:
Thanks, @Mause !

PR #838, opened 2023-06-19
Add qStudio docs for duckdb

ryanhamilton:
Disclaimer: I am the author of qStudio

qStudio is a free SQL IDE specialized for data analysts, it's been available for 10+ years.
In the distance past I used MonetDB so as soon as I heard of DuckDB I added DuckDB integration.

0.6 We bundled.
0.7 we bundled and added first class ability to create new databases
0.8 we now bundle and provide a pivot and asof join example

Compared to other SQL generic IDEs we provide first class support. i.e. we bundle the driver and allow creation of a DuckDB database from within the IDE. In future I want to provide even more integration. 

I'm keen to make qStudio the best possible IDE for DuckDB (outside python) so if you have any ideas, please let me know.


Alex-Monahan:
Thank you for adding this! I'm sure folks will find it useful.

As a random aside, you can use the DuckDB logo in your menu instead of a Duck icon if you'd like! https://duckdb.org/faq#where-do-i-find-the-duckdb-logo

PR #839, opened 2023-06-19
Outdated column name in summarize.md

jexp:
Replaced `approx_count_distinct` with `approx_unique`


Alex-Monahan:
Good catch, thanks!

PR #845, opened 2023-06-26
nullif and reformat utility.md

Alex-Monahan:


Alex-Monahan:
Nullif support was added a while ago, but now it is documented!

Issue #822, opened 2023-06-06
Flatten() is undocumented


alexboks:
The flatten function is not documented
see https://github.com/duckdb/duckdb/pull/3102

PR #846, opened 2023-06-27
Document list flatten function

Alex-Monahan:


Alex-Monahan:
Should address #822. @alexboks, do you have any feedback on these docs? Thanks for the request!

Issue #821, opened 2023-06-06
Hive partitioning page uses parquet_scan alias, whereas main Parquet page used read_parquet


ned2:
As mentioned in this comment: https://github.com/duckdb/duckdb-web/issues/589#issuecomment-1484202585

The [Hive Partition docs page](https://duckdb.org/docs/data/partitioning/hive_partitioning.html) uses `parquet_scan` without mention that its an alias for `read_parquet`, which is the form defaulted to being used in the [Parquet Loading page](https://duckdb.org/docs/data/parquet/overview.html). Some readers (such as myself initially) might infer that you need to use `parquet_scan` specifically to read data from a hive partitioned dataset.

I reckon it could be good to standardise across the docs, and use `read_parquet` instead of `parquet_scan` in the Hive Partitioning page.

PR #847, opened 2023-06-27
Standardize on read_parquet instead of parquet_scan

Alex-Monahan:
Should address #821


ned2:
Hey @Alex-Monahan, sorry for not getting you feedback before! That looks perfect to me! Standardised, while also making people aware of the alias. Nice one :)

PR #849, opened 2023-06-28
Python UDF API

pdet:
Basic Python UDF API docs and example.


pdet:
Oh, excellent, I'll merge them both

Tishj:
@pdet 
While you're working on this page, can you fix this typo?
> By default DuckDB will assume the created function is a pure function, meaning it will produce the same output when given the same output.

The end should be
> when given the same input.

Alex-Monahan:
Thanks for documenting this! I am betting that I will be using this before too long... 

Some of the examples have parameters that aren't mentioned in the overall list (`side_effects`, etc.). Would it be worth including all of those in the original function signature and list?

PR #851, opened 2023-06-29
Added example SQL statement for ST_Read for the spatial extension

cholmes:
Added one example in the spatial extension, did it to be the same form as the spatial copy, where there was a helpful example. 

As a user I struggled with coming up with the exact right syntax from scratch, thankfully I found the info on https://tech.marksblogg.com/duckdb-gis-spatial-extension.html

So hoping to contribute this back to make things a bit easier for others.

### DIFF
--- a/docs/extensions/spatial.md
+++ b/docs/extensions/spatial.md
@@ -101,7 +101,10 @@ Compute relationships and spatial predicates between geometries.
 
 
 ## Spatial Table Functions
-The spatial extension provides a `ST_Read` table function based on the [GDAL](https://github.com/OSGeo/gdal) translator library to read spatial data from a variety of geospatial vector file formats as if they were DuckDB tables.
+The spatial extension provides a `ST_Read` table function based on the [GDAL](https://github.com/OSGeo/gdal) translator library to read spatial data from a variety of geospatial vector file formats as if they were DuckDB tables. For example to create a new table from a GeoJSON file, you can use the following query:
+```sql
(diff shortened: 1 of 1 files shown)


cholmes:
Note I'd be happy to write up some more docs for the spatial extension, I think the next thing to show is going from wkb to geometries and back out, and there's more I could do. But not sure how much the docs want lots of examples versus aiming to be the core definitions.

I started a bit of a tutorial on doing some practical things with DuckDB and GeoParquet at https://gist.github.com/cholmes/f78c225f971c6d935e4340ce9895b4b2  Happy to take some of those and put them in docs, or clean it up and put it some other good place for tutorials, etc.

And great work on the spatial extension! It's really promising and I hope we see more in this direction.

Reslan-Tinawi:
Hello Chris.

I also didn't find a documentation on `ST_Read` in the DuckDB documentation. But it's explained in the spatial extension repository: https://github.com/duckdblabs/duckdb_spatial

I'm not sure about how things should be organized here because I also wanted to contribute an example of using the extension (should I contribute to the spatial repo or here).

Mause:
> Hello Chris.
> 
> I also didn't find a documentation on `ST_Read` in the DuckDB documentation. But it's explained in the spatial extension repository: https://github.com/duckdblabs/duckdb_spatial
> 
> I'm not sure about how things should be organized here because I also wanted to contribute an example of using the extension (should I contribute to the spatial repo or here).

Ideally all documentation should live here so it's accessible and searchable, and contributions are definitely welcome!

PR #854, opened 2023-07-03
Group by ALL / Order by ALL

### LINKED ISSUE #799: Document ORDER BY ALL
As noted by @Maxxen, it's not clear from the docs that ORDER BY ALL exists, could be worth at least of an example in the ORDER BY page, potentially even explicit reference.

Potentially same for GROUP BY ALL.

Alex-Monahan:
Finally documenting this! This closes #799.


Alex-Monahan:
This also adjusts the documentation to show that we are NULLS LAST now.

PR #867, opened 2023-07-10
CI: Ignore archive directories when running linkchecker

szarnyasg:


szarnyasg:
@Alex-Monahan good point. The problem I see is that with every new version the link check in CI is going to take longer and longer and it will start failing as some sites return too many requests. This is already happening to some extent but it's okay most of the time.

I pushed a commit that tries something new: it first only crawls through the latest version of the documentation. Then, if that succeeds, it checks the full documentation. (So there is duplication there.) I tested it locally and the first check was much faster, ~5 minutes, while the second was 20 min+. Let's see how they work inside the CI.

.github/workflows/jekyll.yml
```diff
@@ -75,8 +75,14 @@ jobs:
...
 
-      - name: Link Checker
-        run: |
-          linkchecker https://duckdb.org/ --config .github/linkchecker/linkchecker.conf --file-output html/utf-8/report.html
-      - run: cat report.html >> $GITHUB_STEP_SUMMARY
+      - name: Link Checker on latest
+        run: linkchecker https://duckdb.org/ --ignore-url 'https://duckdb\.org/docs/archive/.*' --config .github/linkchecker/linkchecker.conf --file-output html/utf-8/report.html
+
+      - run: cat report.html >> ${GITHUB_STEP_SUMMARY}
+        if: success() || failure()
```
Mause, line 82:
```suggestion
        if: always()
```

szarnyasg: Good point, it should still display the summary if the second `linkchecker` step failed.

.github/workflows/jekyll.yml
```diff
@@ -75,8 +75,14 @@ jobs:
...
+      - name: Link Checker on latest
+        run: linkchecker https://duckdb.org/ --ignore-url 'https://duckdb\.org/docs/archive/.*' --config .github/linkchecker/linkchecker.conf --file-output html/utf-8/report.html
+
+      - run: cat report.html >> ${GITHUB_STEP_SUMMARY}
+        if: always()
+
+      - name: Link Checker on everything
+        run: linkchecker https://duckdb.org/ --config .github/linkchecker/linkchecker.conf --file-output html/utf-8/report.html
+
+      - run: cat report.html >> ${GITHUB_STEP_SUMMARY}
```
Mause, line 87: Are we wanting to override the report? Or append to it?

szarnyasg: It would make sense to override. The second step runs a superset of the first step, so if the CI gets to it, we'll only need the output of the second step.

PR #876, opened 2023-07-18
PRAGMA platform and two very minor fixes

carlopi:


Alex-Monahan:
Looks great, thank you!

PR #878, opened 2023-07-19
Add Jekyll plugin to render function docs

Mause:
This isn't integrated into the actual website yet, probably needs some fine tuning (and support for search index generation as well)

@szarnyasg this is in response to our conversation the other day

### DIFF
--- a/.github/workflows/jekyll.yml
+++ b/.github/workflows/jekyll.yml
@@ -35,7 +35,11 @@ jobs:
       - name: Setup Pages
         id: pages
         uses: actions/configure-pages@v3
-      - run: git clone https://github.com/duckdb/duckdb
+      - uses: actions/checkout@v3
+        with:
+          repository: duckdb/duckdb
+          path: duckdb
+        name: Clone DuckDB itself
       - name: Build with Jekyll
         # Outputs to the './_site' directory by default
         run: bundle exec jekyll build --baseurl "${{ steps.pages.outputs.base_path }}"
--- a/.github/workflows/jekyll.yml
+++ b/.github/workflows/jekyll.yml
@@ -35,6 +35,7 @@ jobs:
       - name: Setup Pages
         id: pages
         uses: actions/configure-pages@v3
+      - run: git clone https://github.com/duckdb/duckdb
       - name: Build with Jekyll
         # Outputs to the './_site' directory by default
         run: bundle exec jekyll build --baseurl "${{ steps.pages.outputs.base_path }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/jekyll.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       - name: Setup Pages
         id: pages
         uses: actions/configure-pages@v3
+      - run: git clone https://github.com/duckdb/duckdb
```
szarnyasg, line 38: I think this works just fine but maybe use the `checkout` Action? https://github.com/actions/checkout

carlopi:
Just throwing another idea into the mix: what if there was a single json file with all the content needed checked-in somewhere in the duckdb repository? Something that for each function also has the information about path / category.
This will have to be auto-generated as well from the various functions.json.

Then this would unify and simplify (both local development and CI) to:
- fetch `https://raw.githubusercontent.com/duckdb/duckdb/${HASH}/some/path/all_functions.json`
- iterate on records

I guess adding requirement on having internet available is already sort of assumed, so not that bad, but on this I might be wrong. 

And to change to newer / older duckdb it would be a matter only of changing the hash, without having to deal with keeping an extra local copy of duckdb in sync.

carlopi:
And connected but separated from previous point: I would have somewhere explicit which hash is being currently used + available to be changed if needed. It can be master by default, but in the general case we might want to be able to fiddle with that (say generate documentation for a different moment than latest).

_plugins/duckdb_functions.rb
```diff
@@ -0,0 +1,149 @@
...
+    elsif args.length > 0
+      @html += args.join(' ')
+    end
+    @html += "</#{m}>"
+  end
+end
+
+# @return [Array<Object>, nil]
+def get_functions
+  files = Dir.glob('duckdb/src/core_functions/**/*.json')
```
carlopi, line 39: globbing is most likely non deterministic, unsure if already implicit but or it would be better to add a sort somewhere.

carlopi: Also I am not actually sure to have sense to have this automatically generated on the fly during deployment vs checked in. I don't see how documentation versioning would work for example.

I would personally probably go either for checking-in the repo a json containing all information (and the pages can then build from that on the fly) OR having directly a script (ruby or anything else) generate the markdown to be checked in.

Mause:
Going to continue iterating on this, but will merge for now so we can all see what it looks like

PR #879, opened 2023-07-20
build: add functions.json file for plugin

Mause:


scripts/generate_function_json.py
```diff
@@ -0,0 +1,66 @@
...
+                {
+                    **function,
+                    'parameters': (
+                        function['parameters'].split(',')
+                        if function.get('parameters')
+                        else
+                        []
+                    ),
+                    'category': category,
+                    'result': get_result(function['example']) if function.get('example') else None
```
carlopi, line 51: Pretty cool the self-check on results!

Mause: I actually found some issues due to this, going to submit a pr to the main repo shortly

scripts/generate_function_json.py
```diff
@@ -0,0 +1,66 @@
...
+                }
+                for function in json.load(fh)
+            ]
+
+    functions = sorted(functions, key=lambda fn: fn['name'])
+
+    with open('docs/functions.json', 'w') as fh:
```
carlopi, line 58: One small addition I would consider (even later) is given that we need a binary, we can add a field (either to `docs/functions.json` or some ancillary file) like: "duckdb_version": "some_git_hash". I don't see a strong use right now but for avoiding re-generating if fields is the same or tracking down weird behavioiur, but I guess the extra cost is low.

scripts/generate_function_json.py
```diff
@@ -0,0 +1,66 @@
...
+    except CalledProcessError as e:
+        print(example.strip(), e.stderr.decode())
+        return None
+
+
+def main():
+    functions = []
+    source = Path(args.source)
+    assert source.exists()
+    for file in source.glob('src/core_functions/**/*.json'):
```
carlopi, line 38: The only improvement I could see here is that eventually we might want to go full on with duckfooding, where we use duckdb itself (either from within python or as a bash script) to generate functions.json. 🦆

carlopi (approved):
Thanks @Mause! Looks great!

scripts/generate_function_json.py
Mause: Given the way this stuff works is that we don't commit the new file if it hasn't changed, maybe.

But yeah, the fact that the times in this file change on each run is annoying

PR #881, opened 2023-07-20
Clarify what date subfields means (examples)

Alex-Monahan:
Super small change! This is just to help folks find the date part functions more easily. An expert user had trouble finding them, so I imagine that other users may need a little bit more description as well!


hawkfish (commented):
LGTM!

PR #882, opened 2023-07-21
add doc for exp

phrrngtn:
I was unable to find a documentation entry for it anywhere. The example is based on the one for `ln`


Mause:
Thanks!

PR #887, opened 2023-07-24
docs: fix signature of `ST_AsGeoJSON`

peter-gy:
Update docs to properly describe the signature of the `ST_AsGeoJSON(GEOMETRY) -> VARCHAR` function. 
It should reflect that a `GEOMETRY` input yields a `VARCHAR` output, instead of the current `VARCHAR -> VARCHAR`.

Also sorts `ST_AsGeoJSON` alphabetically along with the other `ST_As....` functions and formats the table.


Maxxen (approved):

Maxxen:
Thanks!

PR #889, opened 2023-07-25
Add covar_samp function

### LINKED ISSUE #702: Missing documentation of COVAR_SAMP()
It seems that `COVAR_SAMP()` is supported but not listed here:
https://duckdb.org/docs/sql/aggregates

Other pairs of `X_POP()` and `X_SAMP()` aggregate functions are listed correctly.

Mause:
Fixes #702

### DIFF
--- a/docs/sql/aggregates.md
+++ b/docs/sql/aggregates.md
@@ -77,6 +77,7 @@ The table below shows the available statistical aggregate functions.
 |:---|:---|:---|:---|
 | `corr(y,x)` | Returns the correlation coefficient for non-null pairs in a group. | `COVAR_POP(y, x) / (STDDEV_POP(x) * STDDEV_POP(y))`| - |
 | `covar_pop(y,x)` | Returns the population covariance of input values. | `(SUM(x*y) - SUM(x) * SUM(y) / COUNT(*)) / COUNT(*) ` | - |
+| `covar_samp(y,x)` | Returns the sample covariance for non-null pairs in a group. | `(SUM(x*y) - SUM(x) * SUM(y) / COUNT(*)) / (COUNT(*) - 1)` | - |
 | `entropy(x)` | Returns the log-2 entropy of count input-values. | - | - |
 | `kurtosis(x)` | Returns the excess kurtosis (Fisher's definition) of all input values, with a bias correction according to the sample size. | - | - |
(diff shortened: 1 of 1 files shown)


szarnyasg:
This is a strange one. When running the query with the expression `(SUM(x*y) - SUM(x) * SUM(y) / COUNT(*)) / (COUNT(*) - 1)` in the description, I get different results:

```
DROP TABLE IF EXISTS aggr;

CREATE TABLE aggr(k int, x decimal(10,2), y decimal(10, 2));
INSERT INTO aggr VALUES(1, 10, NULL);
INSERT INTO aggr VALUES(2, 10, 11), (2, 20, 22), (2, 25, NULL), (2, 30, 35);
```

```sql
SELECT k, COVAR_SAMP(x, y)
FROM aggr
GROUP BY k;
```
```console
┌───────┬──────────────────┐
│   k   │ covar_samp(x, y) │
│ int32 │      double      │
├───────┼──────────────────┤
│     1 │                  │
│     2 │            120.0 │
└───────┴──────────────────┘
```
```sql
SELECT k, (SUM(x * y) - SUM(x) * SUM(y) / COUNT(*)) / (COUNT(*) - 1) AS covar_samp
FROM aggr
GROUP BY k;
```
```console
┌───────┬────────────────────┐
│   k   │     covar_samp     │
│ int32 │       double       │
├───────┼────────────────────┤
│     1 │                    │
│     2 │ 51.666666666666664 │
└───────┴────────────────────┘
```

Postgres 15 returns the same results for these queries.

The documentations of [Firebird](https://firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref30/fblangref30-aggfuncs-stats.html) and [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/covar_samp) use this formula. What am I missing here?

szarnyasg:
I figured it out: the formula does not work in the presence of `NULL` values. With those removed, it works correctly:

```
DROP TABLE IF EXISTS aggr;

CREATE TABLE aggr(k int, x decimal(10,2), y decimal(10, 2));
INSERT INTO aggr VALUES(2, 10, 11), (2, 20, 22), (2, 30, 35);

SELECT k, COVAR_SAMP(x, y)
FROM aggr
GROUP BY k;

SELECT k, (SUM(x * y) - SUM(x) * SUM(y) / COUNT(*)) / (COUNT(*) - 1) AS covar_samp
FROM aggr
GROUP BY k;
```

szarnyasg:
The description clearly says `non-null` pairs, so this PR is good to go.

PR #891, opened 2023-07-27
Use FROM-first syntax where applicable

szarnyasg:
Hi, I'd like to propose this change which promotes the use of the `FROM`-first syntax across our documentation.

Using the CSV loaders is a great place to do this as the majority of its calls happen with `SELECT * FROM ...`. That said, it is not obvious whether we should make this change across all the example as it may be confusing to newcomers. To mitigate this, I added links to the `FROM` documentation page which explains the new syntax.

WDYT?


szarnyasg:
@Mause @Alex-Monahan do you have any thoughts/strong opinions on this?

Mause:
> @Mause @Alex-Monahan do you have any thoughts/strong opinions on this?

I'm not sure that removing the full syntax is worth it if we have to add a note to every code block?

szarnyasg:
I discussed this with Mark and he agreed that it's good to promote the syntax in the documentation but we shouldn't overdo it and make it confusing. My latest commit is an attempt at this.

PR #895, opened 2023-07-27
Fix csv delimiter example

### LINKED ISSUE #868: CSV docs with delimiter parameter instead of delim
The examples of reading CSV with explicit delim incorrectly call the param as `delimiter` instead of `delim`(see [CSV Auto Detection page](https://duckdb.org/docs/data/csv/auto_detection.html)).

Should we update the docs accordingly? Happy to push a PR for this simple change

Cabeda:
Resolves #868


szarnyasg (approved):

szarnyasg:
Thank you!

Issue #905, opened 2023-08-01
Multiple redirect_from instructions cause ambiguity


szarnyasg:
On the current `master` branch, both the dev version and version 0.8.1 have a `redirect_from` clause:

https://github.com/duckdb/duckdb-web/blob/485cc79ef21aa5b51128315493cd2225bcc821b8/docs/data/csv/overview.md#L4-L5

https://github.com/duckdb/duckdb-web/blob/485cc79ef21aa5b51128315493cd2225bcc821b8/docs/archive/0.8.1/data/csv/overview.md?plain=1#L4-L5

The resolution of these is ambiguous, causing https://duckdb.org/docs/data/csv to redirect to https://duckdb.org/docs/archive/0.8.1/data/csv/overview.html. This is not the intended target.

To resolve this, `redirect_from` properties should be adjusted by the `scripts/archive_docs.py` script.

PR #907, opened 2023-08-01
fix: adjust redirect_from properties in 0.8.1 archive pages, fix broken links

szarnyasg:
Towards #905


szarnyasg:
Merging this now and will attempt to fix the archiving script for #905 later.

PR #913, opened 2023-08-02
Add release calendar

szarnyasg:


Mause:
I always assumed there was a reason this wasn't public?

szarnyasg:
There was no real reason, just no one got around to publish it. It's also tricky to find a nice place for it - I placed it under Why DuckDB.

Mause:
> There was no real reason, just no one got around to publish it. It's also tricky to find a nice place for it - I placed it under Why DuckDB.

Maybe we should add a link to the version selector? It's the closest thing we have to a list of versions anywhere

carlopi:
First technical comment: I think the appropriate solution to iframes is using a 'component' of sort, I added [this one](https://github.com/duckdb/duckdb-web/blob/master/_includes/iframe.html) at some point for use in the Wasm overview for embedding the shell itself (https://duckdb.org/docs/api/wasm/overview).
It's fixed ratio, at least for now, with width = width of parent container and height = 0.6*width.
I guess the same can be reused in a few other places instead of using fixed sized elements.

Content wise I was thinking it could be called roadmap or in some way mentioning roadmap (since I guess it's the most searched term) and adding a paragraph saying something to the effect "DuckDB project aims for a release about every 12 weeks of development, with all features, performance improvements and bug fixes merged into the master branch up to that point tagged, packaged and made more readily available. Minor releases are sometimes followed by bug-fixes releases where effort is mostly put towards increasing stability.
Releases are announced on GitHub, on this Website and blablabla, here the most recent one: [link to github].

On where to place this: development section is a bit light, and this could as well fit there as https://duckdb.org/dev/roadmap or equivalent.

szarnyasg:
I discussed this with Hannes today. We decided to keep it simple and just add the iframe under the Development section. We can iterate later if needed.

PR #915, opened 2023-08-02
nit: Replace CRLF line terminator with LF in Markdown/config files

szarnyasg:
I'd like to avoid the `^M` characters in git diffs, therefore I looked for CRLF line endings:
```bash
find . -type f -exec file "{}" ";" | grep CRLF
```
and ran `dos2unix` on them. I only applied this to non-archive Markdown files and config files.


Mause:
Wouldn't it be easier to configure your local git client to ignore line endings in diffs? 😅

szarnyasg:
Good point 😅. That said, I'd like to improve the consistency of the documentation's code (line endings, SQL syntax, spelling, formatting, etc.) within the repository and this PR is a step into that direction. (There's no overlap with other PRs so it will not cause a merge conflict.)

PR #916, opened 2023-08-02
nit: Use uppercase keywords

szarnyasg:
It turns out finding instances of lowercase SQL instructions is difficult – SQL is English language after all, and the term `from` appears everywhere in the documentation. So we need to restrict out scope to code blocks.

We can so so using my `agid` (`ag` + `ignore directory` + `docs/`) macro:

```bash
function agid() {
    if [ ${#} -ge 1 ]; then
        ag --ignore-dir=archive/ $@ docs/
    else
        echo "missing search argument"
        return 1
    fi
}

KEYWORD=from
agid --md -s "\`\`\`[a-zA-Z0-9+]+\n.+${KEYWORD}.+\n\`\`\`"
```

### DIFF
--- a/docs/api/cli.md
+++ b/docs/api/cli.md
@@ -429,7 +429,7 @@ Note that the duck head is built with unicode characters and does not always wor
 -- Duck head prompt
 .prompt '⚫◗ '
 -- Example SQL statement
-select 'Begin quacking!' as "Ready, Set, ..."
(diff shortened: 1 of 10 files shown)


Mause:
I wonder if we can enforce this using Vale or something?

szarnyasg:
Hah, I would not have thought about this but it should work.

PR #921, opened 2023-08-03
Fix vale config

Mause:


.github/workflows/vale.yml
```diff
@@ -6,10 +6,12 @@ jobs:
...
     steps:
+      - uses: actions/checkout@v3
       - name: Remove archive docs
         run: |
           rm -rf docs/archive
-      - uses: actions/checkout@v3
       - uses: errata-ai/vale-action@reviewdog
         with:
           files: docs/
+          filter_mode: nofilter
```
Mause, line 16: Not sure how helpful this line will be, happy to tweak it

szarnyasg:
Looks good, and now the errors show up in the PR page.

szarnyasg:
Okay, it's not perfect yet but `Use 'i.e., ' instead of 'IEEE'"` made me laugh.

PR #928, opened 2023-08-06
fix: don't force copying of entire installation instructions block

Mause:


carlopi:
I was looking at the copy button, this seems a decent simple and unobtrusive approach: https://remarkablemark.org/blog/2021/06/01/add-copy-code-to-clipboard-button-to-jeyll-site/#code.

Tutorial I expect to work sort of out of the box (on injection side, and that's basically the only viable solution), but the harder part is CSS and having it work for wider-than frame code-blocks.

Mause:
> I was looking at the copy button, this seems a decent simple and unobtrusive approach: https://remarkablemark.org/blog/2021/06/01/add-copy-code-to-clipboard-button-to-jeyll-site/#code.
> 
> Tutorial I expect to work sort of out of the box (on injection side, and that's basically the only viable solution), but the harder part is CSS and having it work for wider-than frame code-blocks.

I took that code and tweaked it a bit, let me know what you think

carlopi:
Great!
I haven't not been able to try locally, something ruby/permission related, but code makes sense to me.

szarnyasg:
So I tested this locally. It works nicely:

- it allows copying for the code blocks in the installation instructions
Screenshot 2023-08-07 at 16 56 18

- it does not show for some environment/platform combinations, namely the ones that have URLs
Screenshot 2023-08-07 at 16 55 42

- it also shows up at every code block in the docs
Screenshot 2023-08-07 at 16 54 17

szarnyasg:
Perfect!

szarnyasg (approved):

carlopi:
From the images the text doesn't look vertically in the middle, but this is a very minor detail, plus I guess the round corners are not in-style with the rest of the docs, but this are details that we can review on the go, it's very nice, looking forward to seeing this live!

carlopi:
Opened issue #931 on this, to keep track of it.

PR #939, opened 2023-08-08
Document changes to parallel CSV reader

szarnyasg:


pdet:
I don't think we should mention the memory consumption. 
The max difference should be of 3 to 1 buffers (i.e., 94mb vs 32 mb).
Anything else should be a bug with buffers not being properly released.
But removing the experimental is fine!

szarnyasg:
Thanks!

PR #941, opened 2023-08-08
Secondary menu should point to stable release

### LINKED ISSUE #940: Secondary menu should point to stable documentation
#912 changes the "Documentation" button of the front page to point to stable but the secondary menu (i.e. top bar) still points to the `dev` documentation.

szarnyasg:
Fixes #940

### DIFF
--- a/_includes/menu_secondary.html
+++ b/_includes/menu_secondary.html
@@ -1,5 +1,6 @@
 <nav>
 	<div class="search_icon"></div>
+	{% assign archive_url_slug = '/docs/archive/' | append: site.currentduckdbversion %}
 	{% if site.data.secondarymenu.toc2[0] %}
 		{% for item in site.data.secondarymenu.toc2 %}
 			{% if item.mainfolderitems[0] %}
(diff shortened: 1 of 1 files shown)


carlopi:
Even after this, there are plenty of navigation patterns that will still lead to seeing /docs, given it's the default everywhere, example: internal links to https://duckdb.org/docs/installation/.

If we want to have people browser stable docs by default, then I think the proper steps would be move /docs/ to /docs/next/, and then /docs/archive/v0.8.1 to /docs/.

That would ensure all links (internals AND external) point to stable docs.

Then on releases, /docs/ could go to /docs/archive/v0.8.1, /docs/next copied over to /docs, and on we go.

szarnyasg:
Good point. However, I don't like the slug `next/` very much. Some alternatives: `snapshot` (sounds Java-ish to me), `dev` (is already used for the dev documentation),`latest`. WDYT?

Mause:
Maybe `nightly`?

szarnyasg:
I feel like nightly has gone out of fashion somewhat. I checked Postgres and it uses `current`, which I quite like:
https://www.postgresql.org/docs/current/

Mause:
Yeah about that: https://github.com/duckdb/duckdb-web/pull/544

szarnyasg:
Haha, okay. Naming things is hard.

szarnyasg:
I spun up the naming discussion into an issue at #960, so we can focus on the PR here. Let me know whether this is good the merge.

PR #942, opened 2023-08-08
Rewrite redirect clauses in archive script. Fixes #905

### LINKED ISSUE #905: Multiple redirect_from instructions cause ambiguity

szarnyasg:
Fixes #905

### DIFF
--- a/scripts/archive_docs.py
+++ b/scripts/archive_docs.py
@@ -3,6 +3,8 @@ import os
 import shutil
 import datetime
 import subprocess
+import frontmatter
+
 
 # Usage instructions: 
 # Update _config.yml to specify the new version number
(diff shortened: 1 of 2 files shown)


szarnyasg:
I implemented this for the basic case, when there is no `--date` provided.

The case with `--date` is tricky: the script uses Git's stdout and stderr output in binary format. So running text processing on those is not possible. I am on the fence on whether we should keep the `--date` option. I don't think it adds much value (we can always navigate to a given commit manually and run the script) but it adds a noticeable complexity to the Python script.

scripts/archive_docs.py
```diff
@@ -79,10 +80,32 @@ def list_tree(source):
...
-def copy_file(source_path, target_path):
+def copy_file(source_path, target_path, version):
 	print(f"{source_path} -> {target_path}")
 	if revision == None:
-		shutil.copy(source_path, target_path)
+		if source_path.endswith(".md"):
+			with open(source_path) as f, open(target_path, "w") as of:
+				# parse YAML metadata and adjust "redirect_from" fields
+				doc_text = f.read()
+				doc_groups = doc_text.split("---\n")
```
Mause, line 90: Rather than copying the frontmatter parsing and generation code, perhaps we should switch to https://pypi.org/project/frontmatter/ or move the code into a shared package?

szarnyasg:
Pushed a new version of this PR which uses the [`python-frontmatter`](https://pypi.org/project/python-frontmatter/) package.

Mause (approved):

PR #943, opened 2023-08-08
Move the dev/ pages under docs/. Fixes #938

### LINKED ISSUE #938: The internals/ and dev/ pages should be under docs/
The **Development** node in the menu tree, which includes the pages from the `internals/` and `dev/` directories should be under the `docs/` directory and it should be included in the archive for each release.

The advantage of this would be that the dev documentation would also be versioned. We can keep `https://duckdb.org/internals/storage` with a redirect.

szarnyasg:
I moved the `dev` and the `internal` directories under `docs` and adjusted the JSON menus accordingly.

As explained in #938, this helps versioning these pages. It also allows for a uniform handling of {docs, guides, dev, internal} pages.

~~However, note that merging this PR will break existing URLs such as https://duckdb.org/internals/storage. Maybe we should add some redirects to ensure that these still work?~~ resolved

Let me know what you think.


carlopi:
internals/storage is referred from any duckdb version in a somehow frequently seen error message, so I would aim for that link to be always be valid (redirect / copy of content / keeping that page there and moving the rest). It's also special in the sense that's a page that has to be valid cross version.

The rest is fine either way.

szarnyasg:
Thanks, very good point. I pushed a commit that fixes this.

szarnyasg:
@Mause @Alex-Monahan pls let me know what you think.

Mause:
Can you explain why you want to version these pages? As far as I can see, unless someone wants to work on an old version, they won't need the old dev docs

szarnyasg:
After v1.0, I expect quite a few people to anchor down on a given version, so it will be beneficial if these are also versioned, archived. They should also be part of the PDF.

docs/dev/internals/storage.md
```diff
@@ -2,6 +2,8 @@
 layout: docu
 title: Storage
 expanded: Development
+redirect_from:
```
Mause, line 5: This file shouldn't be moved - there should only be one copy that contains the information for all the versions at once

szarnyasg:
I think it cannot hurt for 0.9 either.

And a major benefit or merging this now would be that it would allow for a uniform handling of {docs, guides, dev, internal} pages for the single PDF document, which IMHO should also contain the dev and internal pages.

Mause (changes requested):
Storage should be moved outside of versioning

szarnyasg:
At the moment, it is a bit unclear to me where the developer docs should be. It could also be a wiki that's separate from the main documentation.
In any case, my priority now is to get the user documentation plus guides out in a single-PDF build, so I'll close this now and revisit it later.

PR #944, opened 2023-08-08
CI: Do not fail on Vale errors

szarnyasg:
I propose to configure CI so that it doesn't fail on Vale errors for the time being. It will take some time to get the `e.g.`s and `i.e.`s fixed in the main repo.


Mause:
We could also change the filter to only trip on changed files?

Mause (approved):

PR #953, opened 2023-08-09
Add multi-arch Jekyll Docker image

### LINKED ISSUE #951: Add instructions for how to use Jekyll with Docker
I noticed that people are reluctant to install Ruby and Jekyll so they cannot build the site locally. Providing a script that runs it from Docker would help people build the site locally.

szarnyasg:
This PR add scripts to run run `jekyll serve` through Docker. It's a multi-arch image so it also runs fast on Apple Silicon.

There is no `builder` version (with Linux and Ruby packages pre-baked into the image), and I did not want to roll out a Docker image for this...

So currently the Docker script installs a few packages.

Here are the times required to deploy the `dev` version on my laptop (after a full `git clean -xdf .`):
- native `jekyll`: no setup time + 6.3 s deployment time
- Docker `jekyll`: 13.4 s setup time + 8.0 s deployment time

So the difference is noticeable but the Docker setup is perfectly usable.

Fixes #951

### DIFF
--- a/scripts/docker-serve-full.sh
+++ b/scripts/docker-serve-full.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -xeuo pipefail
+
+# navigate to the repository root
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/.."
+
+scripts/docker-stop.sh
+
+. scripts/docker-vars.sh
+
+docker run \
+    --rm \
+    --interactive \
+    --name ${JEKYLL_DOCKER_CONTAINER_NAME} \
+    --volume="${PWD}:/srv/jekyll:Z" \
+    --publish 4000:4000 \
+    --publish 35729:35729 \
+    ${JEKYLL_DOCKER_IMAGE_NAME} \
+    bundle exec jekyll serve --host 0.0.0.0 $@
--- a/scripts/docker-stop.sh
+++ b/scripts/docker-stop.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -xeuo pipefail
+
+# navigate to the repository root
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/.."
+
+. scripts/docker-vars.sh
+
+docker rm --force \
+    ${JEKYLL_DOCKER_CONTAINER_NAME=duckdb-jekyll}
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+Dockerfile
+.git/*
+.github/*
+_site/*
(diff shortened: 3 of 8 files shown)


scripts/docker-serve-full.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -xeuo pipefail
+
+# navigate to the repository root
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/.."
+
+JEKYLL_DOCKER_IMAGE_VERSION=pages-v228-minimal
```
Mause, line 8: Does this container not provide multiarch? https://github.com/duckdb/duckdb-web/blob/master/.devcontainer/devcontainer.json#L6C12-L6C61

scripts/docker-stop.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -xeuo pipefail
+
+# the Jekyll version used by GitHub is given at https://pages.github.com/versions/,
```
Mause, line 5: We don't use the versions at that url, have a look at the Gemfile and lock file to see which we use

scripts/docker-serve-full.sh
```diff
@@ -0,0 +1,22 @@
...
+    --rm \
+    --interactive \
+    --name ${JEKYLL_DOCKER_CONTAINER_NAME} \
+    --volume="${PWD}:/srv/jekyll:Z" \
+    --publish 4000:4000 \
+    --publish 35729:35729 \
+    rockstorm/jekyll:${JEKYLL_DOCKER_IMAGE_VERSION} \
+    /bin/bash -c \
+        "apk add --update gcc g++ make ruby-bundler && \
+         bundle install && \
```
Mause, line 21: Maybe these should be in a Dockerfile so they don't have to be installed and built every time? And the image can be reused

Mause (changes requested):

scripts/docker-stop.sh
szarnyasg: removed this outdated comment, this was for the `jekyll/jekyll` image

scripts/docker-serve-full.sh
szarnyasg: It does, but it's built-in Jekyll is too new: v4.3.2. So we would need to reinstall Jekyll to match the one used by us (3.9.2).

szarnyasg:
`scripts/docker-build.sh` runs from scratch in 17s.

docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM rockstorm/jekyll:pages-v228-minimal
```
Mause, line 1: Did you mean to add this file?

szarnyasg: Ah, of course not. Thanks.

Issue #956, opened 2023-08-10
Consider Adding DuckDB VSCode extension to list of supported SQL editors


archiewood:
I've authored a VSCode extension which allows you to explore, write queries, as run them against DuckDB files.

https://marketplace.visualstudio.com/items?itemName=Evidence.sqltools-duckdb-driver

If this is of interest, I can submit a PR for review.

Mause:
By all means!

PR #963, opened 2023-08-10
Add docs for VSCode extension

archiewood:
Addresses #956 

CleanShot 2023-08-10 at 11 35 49@2x


archiewood:
Pointed it to the (compressed) version hosted on our GH

Issue #955, opened 2023-08-10
Consider adding Evidence to list of supported Data Viewers


archiewood:
As of latest version, Evidence.dev supports 0.8.1, and ships a duckDB with it's default template.

https://evidence-template.netlify.app/

Happy to write the page and PR in.

PR #965, opened 2023-08-10
Adds docs for Evidence

archiewood:
Addresses #955 

CleanShot 2023-08-10 at 17 45 19@2x


hannes:
Thanks for the PR, but we're regrettably not going to merge this. There a ton of tools out there that use DuckDB in some way and we feel that this kind of documentation should be in the respective tool's documentation and not in this one. Maybe it makes sense to add a reference to Evidence here: https://github.com/davidgasquez/awesome-duckdb

PR #970, opened 2023-08-11
Remove SQL editors except one

szarnyasg:
To reduce the clutter and the number of 3rd party tools in the documentation (in accordance with the [contributor's guide](https://github.com/duckdb/duckdb-web/blob/master/CONTRIBUTING.md)), we are removing most SQL editors.


Alex-Monahan:
Hmm, is there another place these could live? Maybe the DuckDB awesome list? Or maybe we can link to the sites of the IDEs from a single page?

PR #978, opened 2023-08-12
Add source link

### LINKED ISSUE #274: Add "Show source" link on documentation pages
For the documentation of the latest version, a "show source" link is an easy way to tell users where the source of the documentation lives, and also to encourage contributions. Should be easy enough to add to https://github.com/duckdb/duckdb-web/blob/master/_layouts/docu.html .

Examples: See the "Source" link at the top of https://dplyr.tidyverse.org/reference/arrange.html or https://dplyr.tidyverse.org/articles/grouping.html .

szarnyasg:
Attempt to fix #274. It's a hack now, I could not yet figure out why `margin-top: 1px;` was needed.
Let me know what you think.

Preview images:

Screenshot 2023-08-12 at 15 49 42
Screenshot 2023-08-12 at 15 49 46
Screenshot 2023-08-12 at 15 49 49


carlopi:
One problem is that on mobile space is scarce and adding stuff break in most cases, but on this a shortcut I would suggest is just have the div non show there (given on mobile you probably are not going to submit modifications).

---
And in the past when this was raised the sketch was having only the link to be there only for latest/nightly/dev versions (the one we are actually interested in modifying), and in the older version having either "go to latest/stable docs" or nothing at all.
Reason is having an element that help telling people whether something is updated & contribution are welcome or something is frozen in the past.

css/main.scss
```diff
@@ -1096,7 +1096,17 @@ body.documentation{
 		width: calc(100% - 252px);
 		display: flex;
 		justify-content: space-between;
-		align-items: center;
+		align-items: right;
```
carlopi, line 1099:
```suggestion
		align-items: center;
```
// this is the vertical alignment, likely reverting this makes `margin-top:1px` redundant.

szarnyasg:
Fixed by #1096

PR #980, opened 2023-08-12
Close SQL in side menu

szarnyasg:
I noticed that the SQL menu is always open even if we are browsing something different (see secreenshot). I assume this is not on purpose, is it? In any case, this PR fixes it.
 
Screenshot 2023-08-12 at 16 08 03

### DIFF
--- a/_data/menu_docs_dev.json
+++ b/_data/menu_docs_dev.json
@@ -523,7 +523,6 @@
         },
         {
           "page": "SQL",
-          "open": true,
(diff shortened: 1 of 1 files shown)


Mause:
I assumed this was actually purposeful, as it's what people will be most likely to navigate to the docs for?

szarnyasg:
I also had this hunch – I will ask Hannes tomorrow.

PR #992, opened 2023-08-15
Rework Profiling page

### LINKED ISSUE #991: hole in profiling documentation page
Hello everyone,

First of all, thank you very much for this impressive and very useful library :)

on this page : https://duckdb.org/dev/profiling

in the "[Explain Statement](https://duckdb.org/dev/profiling#explain-statement)" subsection;

The paragraph says : "We can see that the logical_plan contains the unoptimized query plan, involving a cross product and a LIKE operation"

I can "see" no such thing. The page doesn't seem to contain any logical_plan, nor any other mention to a cross product besides this paragraph.

Could anyone please either provide the schema that contains the logical_plan with the mention to the cross product, or update the wording ?

szarnyasg:
Fixes #991

### DIFF
--- a/dev/profiling.md
+++ b/dev/profiling.md
@@ -20,11 +20,15 @@ INSERT INTO exams VALUES (1, 1, 8), (1, 2, 8), (1, 3, 7), (2, 1, 9), (2, 2, 10);
 ### Explain Statement
 The first step to profiling a database engine is figuring out what execution plan the engine is using. The `EXPLAIN` statement allows you to peek into the query plan and see what is going on under the hood.
 
-The `EXPLAIN` statement displays three query plans that show what the plan looks like as it passes the various stages of the execution engine. The `logical_plan` is the initial unoptimized plan as it is created right after parsing. The `logical_opt` is the optimized logical plan, that demonstrates the equivalent but optimized logical plan after it passes the optimization phase. This optimized plan is then transformed into the `physical_plan`, which is the plan that will actually get executed.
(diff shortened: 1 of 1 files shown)


dev/profiling.md
```diff
@@ -62,10 +70,12 @@ EXPLAIN SELECT name FROM students JOIN exams USING (sid) WHERE name LIKE 'Ma%';
                              │   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
                              │ Filters: name>=Ma AND name│
                              │  <Mb AND name IS NOT NULL │
+                             │   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
+                             │           EC: 1           │
                              └───────────────────────────┘
 ```
 
-We can see that the `logical_plan` contains the unoptimized query plan, involving a cross product and a `LIKE` operation. The optimized plan transforms this plan and pushes down the filters, transforming the cross product into a comparison join. It also recognizes that the `LIKE` operator only does prefix filtering, and transforms the more expensive `LIKE` operator into a cheaper `PREFIX` selection.
+Note that the query does is not actually executed – therefore, we can only see the estimated cardinality (`EC`) for each operator.
```
benjamin-du, line 78: Maybe add a note about what EC means / what it represents ?

benjamin-du:
That reviewdog thing seems unhappy with the use of i.e., other than that, your update should make the page clearer, thanks :)

dev/profiling.md
szarnyasg: Good point, I added a (very) brief explanation.

PR #995, opened 2023-08-15
Add markdown-lint

Mause:


szarnyasg:
I write follow `MD022 - Headings should be surrounded by blank lines` in my own writings because I find it nicer to read in raw Markdown form. So I would set it to true and adhere it. WDYT?

szarnyasg:
Looks good, let's give this a go.

Issue #977, opened 2023-08-12
What's the purpose of the selected attribute in front matters?


szarnyasg:
The source files have a `selected` attribute:

```
$ head -n 4 docs/sql/statements/attach.md
---
layout: docu
title: Attach/Detach
selected: Documentation/SQL/Attach
```

This gets rendered as the `data-selected` attribute to the website

```
$ grep data-selected _site/docs/sql/statements/attach.html
<div class="sidenavigation" data-selected="Documentation/SQL/Attach" data-expanded="SQL">
```

However, the `selected` field is often missing or it is different from the page title (like above). Removing it seems to have no effect.

@Mause are we currently using this field?

Mause:
It doesn't look like it actually does anything anymore, I'm not sure of the history behind it

It affects this at least:

https://github.com/duckdb/duckdb-web/blob/49b6d081ee4d5501f6cfb1726c0b86cfc28f1c06/_layouts/page.html#L67-L69

But those CSS classes aren't even included in any pages anymore? They're defined in `css/pure-min.css`, which is a copy of https://purecss.io/menus/, which I guess is what the website was originally based on, but this stuff was never cleaned up

Looking further into the old commits, it looks like it was from the old website, whatever that was: https://github.com/duckdb/duckdb-web/commit/5aafb33b6de4e83c69364d6fd71a7c02f0f3277a

PR #999, opened 2023-08-16
Cleanup old website and benchmarks

Mause:
This is a continuation of the work started in https://github.com/duckdb/duckdb-web/issues/977


szarnyasg (approved):
Very satisfying. Tested locally and it still builds and works fine. Thanks!

Issue #281, opened 2022-08-20
Switch to nightly.link for master builds 


Mause:
Much more user friendly as it allows us to link directly to the most recent artifacts

Eg https://nightly.link/duckdb/duckdb/workflows/LinuxRelease/master

Mause:
@Alex-Monahan what do you think?

Alex-Monahan:
Seems really helpful! I typically am using the pip install pre or the npm equivalent, but this looks great for the other cases!

Mause:
@carlopi can probably repurpose this old issue to switch to your nicer hosting/links for nightlies

szarnyasg:
PR #1030 already fixed this issue. I put this in one of the comments but should have done that in the top post.

PR #1030, opened 2023-08-23
Installation page: use 'main' branch, add direct artifact links

szarnyasg:
This PR adds direct links for the artifacts created by the nightly build using artifacts.duckdb.org.

The MacOS build currently fails so those links fail because there is no older build for `main` to fall back to (only `master`): https://github.com/duckdb/duckdb/actions?query=branch:main+event:repository_dispatch+workflow:OSX

**Status:** this is blocked on the MacOS CI build being broken. A fix is proposed at https://github.com/duckdb/duckdb/pull/8679

### DIFF
--- a/_includes/installation.html
+++ b/_includes/installation.html
@@ -88,7 +88,6 @@ print(cursor.execute('SELECT 42').fetchall()){% endhighlight %}
 	<div class="latest main cli"></div>
 	<div class="latest main odbc"></div>
 	<div class="latest main cli binary"></div>
-	<div class="main odbc"></div>
 	<!-- End Empty Results -->
 
 	<div class="latest julia">
@@ -243,17 +242,17 @@ DBInterface.execute(con, "CREATE TABLE integers(i INTEGER)"){%- endhighlight -%}
 		<a href="https://github.com/duckdb/duckdb" target="_blank">https://github.com/duckdb/duckdb</a>
 	</div>
 
-	<div class="main cli binary linux odbc">
+	<div class="main cli cplusplus binary linux odbc">
 		Linux 64-bit: <a href="https://artifacts.duckdb.org/latest/duckdb-binaries-linux.zip" target="_blank">https://artifacts.duckdb.org/latest/duckdb-binaries-linux.zip</a><br/><br/>
 
 		Linux arm64/aarch64: <a href="https://artifacts.duckdb.org/latest/duckdb-binaries-linux-aarch64.zip" target="_blank">https://artifacts.duckdb.org/latest/duckdb-binaries-linux-aarch64.zip</a><br/><br/>
 	</div>
 
-	<div class="main cli binary macos">
+	<div class="main cli cplusplus binary macos">
 		MacOS binaries (universal): <a href="https://artifacts.duckdb.org/latest/duckdb-binaries-osx.zip" target="_blank">https://artifacts.duckdb.org/latest/duckdb-binaries-osx.zip</a><br/><br/>
 	</div>
 
-	<div class="main cli binary win odbc">
+	<div class="main cli cplusplus binary win odbc">
 		Win 64-bit: <a href="https://artifacts.duckdb.org/latest/duckdb-binaries-windows.zip" target="_blank">https://artifacts.duckdb.org/latest/duckdb-binaries-windows.zip</a><br/><br/>
 	</div>
 
--- a/_includes/installation.html
+++ b/_includes/installation.html
@@ -7,8 +7,8 @@
 			<li class="selected" data-id=".latest">
 				<div> {{ site.currentduckdbversion }} <span class="versioninfo">(Latest Release)</span></div>
 			</li>
-			<li data-id=".master">
-				<div>GitHub master <span class="versioninfo">(Bleeding Edge)</span></div>
+			<li data-id=".main">
+				<div>GitHub main <span class="versioninfo">(Bleeding Edge)</span></div>
 			</li>
 		</ul>
 	</div>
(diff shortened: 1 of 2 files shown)


szarnyasg:
So the links work with `wget` but not with `curl`.
See https://stackoverflow.com/questions/20986395/why-does-curl-not-work-but-wget-works
This can be fixed by using `curl -L`. I added example commands to the boxes in the Installation page. Browsers can follow the links just fine.

szarnyasg:
This should only be merged once the OS X build is fixed, which is expected happen after https://github.com/duckdb/duckdb/pull/8674

carlopi:
It would be nice to eventually have a complete script, like:

```
wget artifacts.duckdb.org/linux.zip
unzip linux.zip
unzip linux_cli.zip
chmod +x duckdb
./duckdb --version
```

szarnyasg:
My biggest discovery today is that you can save one line using `jar` to unzip:
```sh
curl -L https://artifacts.duckdb.org/latest/duckdb-binaries-linux.zip | jar x
jar xf duckdb_cli-linux-amd64.zip
chmod +x duckdb
./duckdb
```

carlopi:
At the cost of an extra dependency / black magic feel.

szarnyasg:
Sure - it's not something I'd put on the Installation page :).

Mause (changes requested):
You'll need to update the scripts that correspond to this functionality as well

szarnyasg:
Good point, I updated the JS.

szarnyasg:
This PR fixes #1030 using artifacts.duckdb.org

Mause:
> This PR fixes #1030 using artifacts.duckdb.org

Er, 1030 is this pr?

szarnyasg:
Oh no, we have fallen into a recursion loop :)

szarnyasg:
This PR fixes #281 using artifacts.duckdb.org

_includes/installation.html
```diff
@@ -201,16 +201,16 @@ <h3>Usage Example</h3>
...
 
-	<div class="master r">
+	<div class="main r">
 		{% highlight r %}install.packages('duckdb', repos=c('https://duckdb.r-universe.dev', 'https://cloud.r-project.org')){% endhighlight %}
 	</div>
 
-	<div class="master java">
-		Java master builds are available in the sonatype snapshots repository, which can be accessed with the following:
+	<div class="main java">
+		Java main builds are available in the sonatype snapshots repository, which can be accessed with the following:
```
Mause, line 213:
```suggestion
		Java dev builds are available in the sonatype snapshots repository, which can be accessed with the following:
```

_includes/installation.html
```diff
@@ -234,36 +234,44 @@ <h3>Usage Example</h3>
...
-	<div class="master cli binary linux">
-		Linux Build Artifacts are available from <a href="https://github.com/duckdb/duckdb/actions?query=branch%3Amaster+event%3Arepository_dispatch+workflow%3ALinuxRelease" target="_blank">the "LinuxRelease" CI runs</a>
-	</div>
+	<div class="main cli binary cplusplus macos">
+		MacOS binaries (universal): <a href="https://artifacts.duckdb.org/latest/duckdb-binaries-osx.zip" target="_blank">https://artifacts.duckdb.org/latest/duckdb-binaries-osx.zip</a><br/><br/>
 
-	<div class="master cli binary win">
-		Windows Build Artifacts are available from <a href="https://github.com/duckdb/duckdb/actions?query=branch%3Amaster+event%3Arepository_dispatch+workflow%3AWindows" target="_blank">the "Windows" CI runs</a>
+		wget https://artifacts.duckdb.org/latest/duckdb-binaries-osx.zip<br/>
+		curl -L https://artifacts.duckdb.org/latest/duckdb-binaries-osx.zip -o duckdb-binaries-osx.zip
```
Mause, line 262:
```suggestion
		curl -L https://artifacts.duckdb.org/latest/duckdb-binaries-osx.zip -O
```

_includes/installation.html
```diff
@@ -234,36 +234,44 @@ <h3>Usage Example</h3>
...
 	</div>
 
+	<div class="main cli binary cplusplus win odbc">
+		Win 64-bit: <a href="https://artifacts.duckdb.org/latest/duckdb-binaries-windows.zip" target="_blank">https://artifacts.duckdb.org/latest/duckdb-binaries-windows.zip</a><br/><br/>
+
+		wget https://artifacts.duckdb.org/latest/duckdb-binaries-windows.zip<br/>
+		curl -L https://artifacts.duckdb.org/latest/duckdb-binaries-windows.zip -o duckdb-binaries-windows.zip
```
Mause, line 269: Curl isn't really used on Windows, except as the alias inside powershell (maybe just use invoke web request directly instead?)

Mause (changes requested):
I'm not sure how important providing commands is? I think it's more important to have clickable links

Either way, some minor corrections

_includes/installation.html
```diff
@@ -234,36 +234,44 @@ <h3>Usage Example</h3>
...
 
-		<p>Linux Build Artifacts are available from <a href="https://github.com/duckdb/duckdb/actions?query=branch%3Amaster+event%3Arepository_dispatch+workflow%3ALinuxRelease" target="_blank">the "LinuxRelease" CI runs</a></p>
+		$ wget https://artifacts.duckdb.org/latest/duckdb-binaries-linux.zip<br/>
+		$ curl -L https://artifacts.duckdb.org/latest/duckdb-binaries-linux.zip -o duckdb-binaries-linux.zip<br/><br/>
 
-		<p>Windows Build Artifacts are available from <a href="https://github.com/duckdb/duckdb/actions?query=branch%3Amaster+event%3Arepository_dispatch+workflow%3AWindows" target="_blank">the "Windows" CI runs</a></p>
-	</div>
+		Linux arm64/aarch64: <a href="https://artifacts.duckdb.org/latest/duckdb-binaries-linux-aarch64.zip" target="_blank">https://artifacts.duckdb.org/latest/duckdb-binaries-linux-aarch64.zip</a><br/><br/>
 
-	<div class="master cli binary macos">
```
hannes, line 254: was this div deleted by accident?

szarnyasg:
> I would also not add curl commands at all, just show the URL so people can click

I removed the commands

PR #1032, opened 2023-08-23
Noting that default parameters must be named.

marhar:
Clarifying that default parameters must be named, which is different than some other languages.


szarnyasg (approved):

szarnyasg:
Thanks!

PR #1035, opened 2023-08-24
Add semicolons to SQL statements

szarnyasg:


szarnyasg:
I think this is super hard to detect reliably (without false positives), so we should defer adding a linter rule for now. It's also not a critical issue, so we can just check it manually.

PR #1041, opened 2023-08-24
Rework Substrait page

szarnyasg:
I took a pass over the Substrait page. @pdet who would be a good candidate to review this PR?

The link to https://github.com/duckdblabs/substrait is currently in a comment, shall we uncomment it?

### DIFF
--- a/docs/extensions/substrait.md
+++ b/docs/extensions/substrait.md
@@ -3,54 +3,61 @@ layout: docu
 title: Substrait
 ---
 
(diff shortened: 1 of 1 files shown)


pdet:
LGTM, I think uncommenting it is fine. Maybe we should add a note that individual support to the substrait extension has to be granted beforehand.

Sth like:
```
The substrait integration is experimental. Support is currently only available on request.
If you have not asked for permission to ask for support, [contact us prior to opening an issue](https://duckdblabs.com/contact/). If you open an issue without doing so, we will close it without further review.
```

PR #1045, opened 2023-08-25
Add 'Alter View' page and document ALTER...NULL constraints

### LINKED ISSUE #714: Document ALTER VIEW
It appears that `ALTER VIEW v RENAME TO w` is valid syntax. Browsing the source code, I can see how this can be supported already:
https://github.com/duckdb/duckdb/blob/master/src/parser/parsed_data/alter_info.cpp

Other `ALTER` commands seem available too (`ALTER_SCALAR_FUNCTION` and `ALTER_TABLE_FUNCTION`).

Yet the documentation doesn't list this syntax yet:
https://duckdb.org/docs/sql/statements/alter_table

szarnyasg:
Fixes #714

### DIFF
--- a/_data/menu_docs_dev.json
+++ b/_data/menu_docs_dev.json
@@ -534,6 +534,10 @@
                   "url": "alter_table"
                 },
                 {
+                  "page": "Alter View",
+                  "url": "alter_view"
+                },
+                {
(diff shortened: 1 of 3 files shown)


Mause:
@szarnyasg would be good to have a railroad diagram here too

PR #1052, opened 2023-08-26
Archive installation page

### LINKED ISSUE #1044: Archived installation pages point to the latest stable version
A minor thing but the archived installation pages point to the latest stable version.
E.g. currently the page https://duckdb.org/docs/archive/0.7.1/installation points to 0.8.1.

szarnyasg:
Fixes #1044

### DIFF
--- a/scripts/archive_docs.py
+++ b/scripts/archive_docs.py
@@ -138,9 +138,34 @@ def recursive_copy(source, target, version):
             recursive_copy(source_path, target_path, version)
 
 
+def archive_installation_page(version):
+    # get frontmatter of the proxy installation file (which includes the actual one)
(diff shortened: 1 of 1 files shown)


Mause (commented):
Needs formatting, but looks good

carlopi:
Just noted, installation page depends on JS to work, and that is also bound to change over time.

Unsure if that points to just having a single installation page cross version (and some way to customise via URL parameters).

Or just a single installation page cross version + some different resource (that needs to be put together) with a collection of links to the specific version (without any javascript logic).
Also the page says explicitly 'latest release'.

szarnyasg:
> Needs formatting, but looks good

Thanks, fixed

PR #1054, opened 2023-08-26
Archive single PDF document

### LINKED ISSUE #1053: Archive should include the single-PDF document
As discussed in #1054, archiving PDFs is likely best done in S3 or a similar service (and not in the documentation's repository).

szarnyasg:
Fixes #1053

### DIFF
--- a/scripts/archive_docs.py
+++ b/scripts/archive_docs.py
@@ -138,9 +138,23 @@ def recursive_copy(source, target, version):
             recursive_copy(source_path, target_path, version)
 
 
+def archive_single_pdf_document(version):
+    copy_file(
+        'single-file-document/duckdb-docs.pdf',
+        f'docs/archive/{version}/duckdb-docs-v{version}.pdf',
+        version,
+    )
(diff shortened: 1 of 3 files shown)


scripts/archive_docs.py
```diff
@@ -138,9 +138,23 @@ def recursive_copy(source, target, version):
             recursive_copy(source_path, target_path, version)
 
 
+def archive_single_pdf_document(version):
+    copy_file(
+        'single-file-document/duckdb-docs.pdf',
```
Mause, line 143: So whoever runs the archive script will have to download the file or generate it?

Mause (changes requested):

Mause:
I have mixed feelings about committing generated binary files - perhaps they're better stored in s3 somewhere?

szarnyasg:
> I have mixed feelings about committing generated binary files - perhaps they're better stored in s3 somewhere?

That's an excellent point – I think nobody likes cloning huge git repos. S3 is likely a better option.

PR #1065, opened 2023-08-31
Explain when to use JSON multi-extract

lnkuiper:
Documents duckdb/duckdb#8734

### DIFF
--- a/docs/extensions/json.md
+++ b/docs/extensions/json.md
@@ -351,6 +351,23 @@ SELECT j->'species'->>[0,1] FROM example;
 -- [duck, goose]
 ```
 
+If multiple values need to be extracted from the same JSON, it is more efficient to extract a list of paths:
+```sql
+-- The following will cause the JSON to be parsed twice,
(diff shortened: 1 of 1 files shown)


docs/extensions/json.md
```diff
@@ -351,6 +351,23 @@ SELECT j->'species'->>[0,1] FROM example;
...
 ```
 
+If multiple values need to be extracted from the same JSON, it is more efficient to extract a list of paths:
+```sql
+-- The following will cause the JSON to be parsed twice,
+-- resulting in a slower query that uses more memory
+SELECT json_extract(j, 'family') AS family,
+       json_extract(j, 'species') AS species
+FROM example;
+-- The following is faster and more memory
```
Mause, line 361:
```suggestion
-- The following is faster and more memory efficient 
```

Mause (changes requested):

izellevy:
Thanks @lnkuiper !

Issue #1068, opened 2023-08-31
alter table doesn't include "if not exists"


paultiq:
https://duckdb.org/docs/sql/statements/alter_table.html is missing syntax showing "if not exists" for alter table add column: 
```py
import duckdb
with duckdb.connect() as con:
    con.execute('create table xyz (col1 int)')
    con.execute('alter table xyz add column if not exists col2 varchar')
    df = con.execute('select * from xyz').df()
    print(df)
```

PR #1069, opened 2023-08-31
Add IF NOT EXISTS for ALTER

Mause:
Fixes https://github.com/duckdb/duckdb-web/issues/1068

### DIFF
--- a/js/statements/alter.js
+++ b/js/statements/alter.js
@@ -45,6 +45,11 @@ function GenerateAlterTable(options = {}) {
 				Sequence([
 					Keyword("ADD"),
 					Optional("COLUMN"),
+					Optional(Sequence([
+						Keyword("IF"),
+						Keyword("NOT"),
+						Keyword("EXISTS")
+					]), "skip"),
 					Expression("column-name"),
 					Expression("type-name"),
 					Expandable("column-constraints", options, "column-constraints", GenerateColumnConstraints)


js/statements/alter.js
```diff
@@ -45,6 +45,11 @@ function GenerateAlterTable(options = {}) {
 				Sequence([
 					Keyword("ADD"),
 					Optional("COLUMN"),
```
Mause, line 47: @szarnyasg not sure if COLUMN should be optional here - thoughts?

szarnyasg: It seems to be working fine:
```
⚪◗ create table t(x int);
⚪◗ alter table t add y varchar;
⚪◗ from t;
┌───────┬─────────┐
│   x   │    y    │
│ int32 │ varchar │
├─────────────────┤
│     0 rows      │
└─────────────────┘
```

Postgres also lists `COLUMN` as optional: https://www.postgresql.org/docs/current/sql-altertable.html
```sql
ADD [ COLUMN ] [ IF NOT EXISTS ] column_name data_type [ COLLATE collation ] [ column_constraint [ ... ] ]
DROP [ COLUMN ] [ IF EXISTS ] column_name [ RESTRICT | CASCADE ]
```

And the grammar definition confirms this:
https://github.com/duckdb/duckdb/blob/main/third_party/libpg_query/grammar/statements/alter_table.y#L137

szarnyasg (approved):

PR #1071, opened 2023-09-01
Add ends_with as an alias for suffix

holdenmatt:
One-line PR to add `ends_with` as an alias for `suffix`, corresponding to this PR:
https://github.com/duckdb/duckdb/pull/8763

This is my first contribution, and I wasn't sure if I should run the scripts to generate functions.json/search_data.json as part of this PR, or if that will happen automatically.


docs/sql/functions/char.md
```diff
@@ -20,6 +20,7 @@ This section describes functions and operators for examining and manipulating st
 | `concat(`*`string`*`, ...)` | Concatenate many strings together | `concat('Hello', ' ', 'World')` | `Hello World` |
 | `concat_ws(`*`separator`*`, `*`string`*`, ...)` | Concatenate strings together separated by the specified separator | `concat_ws(', ', 'Banana', 'Apple', 'Melon')` | `Banana,Apple,Melon` |
 | `contains(`*`string`*`, `*`search_string`*`)` | Return true if *search_string* is found within *string* | `contains('abc', 'a')` | `true` |
+| `ends_with(`*`string`*`, `*`search_string`*`)`| Return true if *string* ends with *search_string* | `ends_with('abc', 'c')` | `true` |
```
Mause, line 23: Can you add an alias column to the table instead and list it there instead?

Mause (changes requested):

Mause:
The autogeneration scripts run every Monday, no need to do that yourself

docs/sql/functions/char.md
holdenmatt: I was trying to follow the existing style of the docs here:
https://duckdb.org/docs/archive/0.8.1/sql/functions/char

There are a bunch of aliases listed in this way already as separate rows. I suppose one advantage of keeping separate rows vs moving to a column is that it makes it easy to find any of the functions alphabetically?

Mause: While this is true, we're trying to remove the duplication on these pages, as having one row per alias is somewhat wasteful and hard to maintan

holdenmatt: Ok, I added an Alias column and deduped most of the existing text function aliases (including this one). PTAL.

Mause (approved):

PR #1072, opened 2023-09-02
Script to check cross-links in PDF, fixes to formatting and (mostly) links

szarnyasg:
Had a long flight after VLDB, here's a script that checks for broken links with LaTeX and a bunch of fixes for the broken links found.

The only notable change is that I added Time Zones to the side menu. If it shouldn't be there, let me know.


single-file-document/list_undefined_references.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
Mause, line 1: Can we add this script to the lint Github action?

Mause (changes requested):

single-file-document/list_undefined_references.sh
szarnyasg: I find this script to be a bit too heavy-weight as it needs all the LaTeX machinery. It also still produces ~180 errors for the NodeJS and Python API docs. I will look into those later. So I'd defer adding this to the lint action now.

docs/sql/configuration.md
```diff
@@ -95,7 +95,7 @@ Below is a list of all available settings.
 | s3_url_style                             | S3 url style ('vhost' (default) or 'path')                                                                                                              | VARCHAR    | VHOST                    |
 | s3_use_ssl                               | S3 use SSL (default true)                                                                                                                               | BOOLEAN    | 1                        |
 | schema                                   | Sets the default search schema. Equivalent to setting search_path to a single value.                                                                    | VARCHAR    | MAIN                     |
-| search_path                              | Sets the default search search path as a comma-separated list of values                                                                                 | VARCHAR    |                          |
```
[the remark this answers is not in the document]
szarnyasg: Sure

PR #1106, opened 2023-09-08
Update update-docs.yml to get the latest `main` build instead

Mause:


Mause:
@carlopi would https://github.com/carlopi/duckdb-latest be safe to switch to for this yet do you think?

Issue #218, opened 2022-05-12
Add imgbot


domoritz:
imgbot automatically sends PRs to optimize images and compress them significantly. Could be useful for blog posts. 

See for example https://github.com/cmudig/cmudig.github.io/pull/160

szarnyasg:
Thanks, we added it.

PR #1111, opened 2023-09-09
Cleanup/convert some large files

szarnyasg:


Mause (changes requested):
Why convert these files? ~~Why commit the pickle files?~~ my bad, you deleted them instead

szarnyasg:
The PNGs were huge (the Jason duck PNG was ~5MB) and quality difference isn't noticeable when converted to JPG.
We could also revisit issue #218 in the near future.

Mause:
> The PNGs were huge (the Jason duck PNG was ~5MB) and quality difference isn't noticeable when converted to JPG. We could also revisit issue #218 in the near future.

Maybe we should leave these files as is to test out imgbot?

szarnyasg:
I doubt imgbot will convert from PNG to JPG (which was the trick here). I'll add imgbot in a moment but it will not resolve this.

szarnyasg:
Update: setting up imgbot requires owner permissions, which I don't have:
https://github.com/marketplace/imgbot/plan/MLP_kgDNGw8#pricing-and-setup

Mause (approved):

PR #1112, opened 2023-09-09
CI: Deploy the latest snapshot and latest stable docs as a single ZIP file

### LINKED ISSUE #794: suggest add html.zip/pdfs documents for every versions download
like
https://www.postgresql.org/docs/ did.

szarnyasg:
Fixes #794

### DIFF
--- a/.github/workflows/jekyll.yml
+++ b/.github/workflows/jekyll.yml
@@ -40,6 +40,12 @@ jobs:
         run: bundle exec jekyll build --baseurl "${{ steps.pages.outputs.base_path }}"
         env:
           JEKYLL_ENV: production
+      - name: Make zipped archive of the site available
+        run: |
+          cd _site/
+          zip -r duckdb-docs.zip * -x "docs/archive/*"
+          CURRENT_DUCKDB_VERSION=$(grep "currentduckdbversion:" _config.yml | cut -d ' ' -f 2)
+          zip -r duckdb-docs.zip docs/archive/${CURRENT_DUCKDB_VERSION}
       - name: Upload HTML artifact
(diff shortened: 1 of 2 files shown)


Mause:
I'm not sure we need both this and a PDF? People would probably need to spin up a local http server for it to be useful anyway

.github/workflows/jekyll.yml
```diff
@@ -40,6 +40,12 @@ jobs:
         run: bundle exec jekyll build --baseurl "${{ steps.pages.outputs.base_path }}"
         env:
           JEKYLL_ENV: production
+      - name: Make zipped archive of the site available
+        run: |
+          cd _site/
+          zip -r duckdb-docs.zip * -x "docs/archive/*"
```
Mause, line 46: You're tagging this file using the currentduckdbversion variable, but then excluding the files for that version as they're in the archive folder?

Mause (changes requested):

.github/workflows/jekyll.yml
szarnyasg: I am excluding the archives but then I'm adding the latest one of them back:
https://github.com/duckdb/duckdb-web/pull/1112/files#diff-60ac246cc02819dea1007076ba57241caa7c7d1e8232a5179124d4e63bb366f8R47-R48

So the zip has latest and stable, which is IMHO a good combination to have.

szarnyasg:
> I'm not sure we need both this and a PDF? People would probably need to spin up a local http server for it to be useful anyway

That's true, I was expecting index.html to work out of the box but it doesn't really work due to the use of JS and absolute URLs.

PR #1121, opened 2023-09-12
Extensions: Rework overview page

szarnyasg:
Joint work with @carlopi


docs/extensions/overview.md
```diff
@@ -2,32 +2,95 @@
...
 
-## Remote Installation
+## Overview
 
-If a given extensions is not available with your distribution, you can do the following to make it available.
+DuckDB has a flexible extension mechanism that allows for dynamically loading extension.
+These may extend DuckDB's functionality by providing support for additional file formats, introducing new types, and domain-specific functionality.
+
+> Extensions are loadable on all clients (e.g., Python and R).
+> While extensions can restrict their target platforms, the extensions distributed by the DuckDB developer team support all major platforms (Linux, MacOS, Windows).
```
Mause, line 12: While this is broadly speaking true, we don't presently support linux_arm64 or windows_amd64_mingw for example

carlopi: Extension can restrict their target platforms, for example `jemalloc` wraps an improved malloc implementation, but the original C++ code is not compatible with OSX malloc implementation.
??

Mause: Not sure jemalloc is a useful example there - it's only ever linked in, and only on linux

carlopi: Then something more factual like:
```
Extension distributed via the official repository are built and tested on MacOS (both arm64 and amd64), Windows (x and y) and Linux (a, b, c, and d).
```

carlopi: And then in the "working with extensions" page one can go more in-depth about details.

szarnyasg: I pushed a clarification

docs/extensions/working_with_extensions.md
```diff
@@ -0,0 +1,55 @@
...
+http://extensions.duckdb.org/v{release_version_number}/{platform_name}/{extension_name}.duckdb_extension.gz
+```
+
+For example:
+
+```text
+http://extensions.duckdb.org/v{{ site.currentduckdbversion }}/windows_amd64/json.duckdb_extension.gz
+```
+
+The list of supported platforms may increase over time, but the current list of platforms includes:
```
Mause, line 23: @szarnyasg are we able to check this list is up-to-date per what is in the bucket?

docs/extensions/overview.md
```diff
@@ -2,109 +2,124 @@
...
-* linux_amd64
-* linux_arm64
-* osx_amd64
-* osx_arm64
-* windows_amd64
-* windows_amd64_rtools
+This will use the `json` extension to read the JSON file.
 
-See above for a list of extension names and how to pull the latest list of extensions.
+> To make the DuckDB distribution lightweight, it only contains a few fundamental built-in extensions (e.g., `autocomplete`, `json`, `parquet`), which are loaded automatically upon startup.
```
Mause, line 48: This should probably be amended to mention it only applies to the CLI

docs/extensions/overview.md
```diff
@@ -2,109 +2,124 @@
...
-## Extension Template
+If you are using the [Python API client](../api/python/overview), you can install and load them with the `load_extension(name: str)` and `install_extension(name: str)` methods.
+
+> Autoloadable extensions can also be installed explicitly.
 
-A template for creating extensions is available in the [`extension-template` repository](https://github.com/duckdb/extension-template/). Note that this project is work-in-progress.
+### Ensuring the Integrity of Extensions
 
-## Pages in This Section
+Extensions are signed with a cryptographic key. By default, DuckDB uses its built-in public keys to verify the integrity of extension before loading them.
```
Mause, line 82:
```suggestion
Extensions are signed with a cryptographic key, which also simplifies distribution (this is why they are served over `http`, not `https`). By default, DuckDB uses its built-in public keys to verify the integrity of extension before loading them.
```

Mause (approved):
Couple of small nits, but otherwise looks good!

docs/extensions/overview.md
```diff
@@ -45,7 +45,7 @@ SELECT * FROM 'test.json';
 
 This will use the `json` extension to read the JSON file.
 
-> To make the DuckDB distribution lightweight, it only contains a few fundamental built-in extensions (e.g., `autocomplete`, `json`, `parquet`), which are loaded automatically upon startup.
+> To make the DuckDB distribution lightweight, it only contains a few fundamental built-in extensions (e.g., `autocomplete`, `json`, `parquet`), which are loaded automatically when the CLI client is started.
```
carlopi, line 48:
```
To make the DuckDB distribution lightweight, by default it only contains a few fundamental built-in extensions (e.g., `json` and `parquet`), which are loaded automatically upon startup.
```
?

docs/extensions/overview.md
carlopi: I don't agree with this change

carlopi: Actually, it's fine either way. I will make a paragraph expanding on this in "Working with extensions", but can go in a separate PR.

docs/extensions/overview.md
szarnyasg: ?

Issue #1124, opened 2023-09-12
Syntax highlighter should support Postgres, SQLite, and DuckDB keywords


szarnyasg:
I spent ~30 minutes looking into using a custom syntax highlighter that supports all the Postgres keywords (`ON CONFLICT`, ...), the SQLite keywords (`RETURNING`, ...) and also the keywords DuckDB introduced (`GROUP BY ALL`, ...).

It seems to be doable but it may take some time to do.

* Fork [Rouge](https://github.com/rouge-ruby/rouge) at version 3.x (e.g. [3.30](https://github.com/rouge-ruby/rouge/releases/tag/v3.30.0))
* Create a custom SQL lexer
* Install this Rouge package like:
  * `gem build rouge.gemspec`
  * `gem install rouge-3.3823.0.gem`
* Ensure Jekyll uses it when generating the website

There is an unfinished PR for a Postgres lexer with a brief discussion that can be useful: https://github.com/rouge-ruby/rouge/pull/1575

Mause:
I think the best route would be to create a lexer using DuckDB's own lexer, if we want to go down this route? Much of that is already exposed via the tokenize method in the Python package

szarnyasg:
I edited the existing list of keywords in Rogue's SQL lexer: https://github.com/rouge-ruby/rouge/compare/master...szarnyasg:rouge:duckdb

I used it on the website now locally:

Before:
![Screenshot 2023-09-12 at 14 50 03](https://github.com/duckdb/duckdb-web/assets/1402801/365a6045-b82a-4c4e-bd71-81a48a400c8e)

After:
![Screenshot 2023-09-12 at 14 50 20](https://github.com/duckdb/duckdb-web/assets/1402801/74530060-5b48-4379-84a0-660d5ce0dc5b)

So this only took an hour – I am already very happy with this!

PR #1125, opened 2023-09-12
Use custom Rouge Gem to include more SQL keywords

szarnyasg:
Towards #1124


Mause:
Can we use this instead? https://bundler.io/guides/git.html

Mause (changes requested):
Switch to using bundler

Mause:
:shipit:

szarnyasg:
Thanks, I will keep working on this later

PR #1130, opened 2023-09-13
design adjustments

Franz-Kafka:
- code blocks same width as tables
- Page Source button only visible in /docs/
- adjusted scroll position on hash-links
- hash-links on FAQ-page will expand the paragraph
- fixed copy button on installations page


szarnyasg:
Great, thank you!

PR #1141, opened 2023-09-14
Rename SQLType to LogicalType in C++ API documentation

### LINKED ISSUE #1140: C++ API still mention `SQLType` instead of `LogicalType`
When [`SQLType` was renamed to `LogicalType`](https://github.com/duckdb/duckdb/commit/07687c24499bc3605d77b755d3344eafb8e5df14) the change was not propagated to the [C++ API documentation](https://duckdb.org/docs/api/cpp.html).

Taepper:
Resolves #1140.

I am not sure whether my change to
> The high-level ("SQL type") types of the result set columns are in `sql_types` and the low-level data representation types are in `types`. 

is sensible.

A quick look at `MaterializedQueryResult` made me assume that the low-level data types are not exposed anymore and the high-level type vector is now called `types` instead.

### DIFF
--- a/docs/api/cpp.md
+++ b/docs/api/cpp.md
@@ -41,7 +41,7 @@ if (!result->success) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Mause:
There are still some things out-of-date on this page, but this is great progress, thanks

PR #1150, opened 2023-09-15
Add back list_intersect, list_has_any, list_has_all

maiadegraaf:
These got lost somewhere along the way.

Solves #1145


Mause:
These were probably lost in the merge I made a few weeks ago, thanks for adding them back!

PR #1159, opened 2023-09-16
Add extra space

szarnyasg:


Mause:
Can probably look at adding blacken-docs to our lint workflow to fix these kinds of issues

PR #1185, opened 2023-09-19
Silly DOC Update csv_import.md

gbrian:
Sorry for annoyin but single quotes where causing duckdb binary complain :(


szarnyasg:
Hi @gbrian, this should work with both single and double quotes. In most cases where DuckDB expectes a string, only single quotes are accepted, so it would be good to keep this with single quotes for consistency. Can you please tell us the following information?

1) DuckDB version
2) OS
3) the error message

PR #1202, opened 2023-09-21
Adjust headers for consistency: Start from h2

szarnyasg:
I wrote a script to find documents which start their headings from h3 instead of h2 (blasphemy!). Luckily, DuckDB makes this super easy. Just create a list of files with h2 headers and another with h3 headers and perform the substraction (h3\h2).

Shell:
```bash
ag --md --ignore-dir=archive/ --files-with-matches '^### ' docs/ > l3
ag --md --ignore-dir=archive/ --files-with-matches '^## '  docs/ > l2
```

DuckDB:
```sql
create table l3 as from read_csv_auto('l3');
create table l2 as from read_csv_auto('l2');
select * from l3 except from l2;
```


szarnyasg:
Order has been restored.

PR #1236, opened 2023-09-24
Eliminate bold formatting in headers

szarnyasg:
WIP

### DIFF
--- a/docs/api/python/relational_api.md
+++ b/docs/api/python/relational_api.md
@@ -59,11 +59,11 @@ When printing a relation using `show` or displaying it in the terminal, the firs
 
 Outside of SQL queries, the following methods are provided to construct relation objects from external data.
 
-* **from_arrow**
-* **from_df**
-* **read_csv**
-* **read_json**
-* **read_parquet**
+* `from_arrow`
+* `from_df`
+* `read_csv`
+* `read_json`
+* `read_parquet`
 
 ## SQL Queries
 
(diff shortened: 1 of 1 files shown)


scripts/generate_docs.py
```diff
@@ -329,7 +327,7 @@ def replace_docs_in_file(file_name, group_name, docs_string_for_this_group):
     if group_name is not current_group_name:
         if current_group_name is not None:
             total_quick_docs += quick_docs_end() + '\n'
-        total_quick_docs += '### **' + group_name + '**\n\n'
+        total_quick_docs += '### ' + group_name + '\n\n'
```
Mause, line 330: Did you forget to add the backticks here?

Mause (changes requested):

scripts/generate_docs.py
szarnyasg: I'm very much torn about the backticks. Sometimes they look really bad, e.g., yesterday I saw something like ```## Loading with the `parquet` extension``` and it looked very bad. That may be classified as a CSS problem then. So let's add them here?

Mause: Did you not add them elsewhere in this pr?

szarnyasg: Yes I did. I'm really torn! On their own, they look fine, it's where they're mixed with regular (sans serif) headings that they look bad. But I've convinced myself now that that is a CSS issue, so let's add them.

Mause (approved):

PR #1238, opened 2023-09-24
chore: Remove spaces at EOL

krlmlr:
Do we need an `.editorconfig` here?

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -1,5 +1,5 @@
 $(document).ready(function(){
-	
+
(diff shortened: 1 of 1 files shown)


krlmlr:
This will cause conflicts everywhere, also would want to convert tabs to spaces or vice versa. Let's do it when all other PRs are merged.

krlmlr:
Also need to do the `.scss` files, and perhaps more.

PR #1240, opened 2023-09-24
docs: For library binaries, indicate what compiler they have been built with

krlmlr:
Discussion at <https://github.com/duckdb/duckdb/discussions/9060>.


szarnyasg:
This is a great addition, thanks.

PR #1245, opened 2023-09-24
Revert "fix: evironment -> environment"

krlmlr:
I confirmed with `git bisect` that this is the commit that breaks the installation page. Reverting this commit works for me locally.

I don't understand what's going on here.

This reverts commit 79fbd8508516ca5cf8668d046d310b35b4915038.

### DIFF
--- a/_includes/installation.html
+++ b/_includes/installation.html
@@ -13,7 +13,7 @@
 		</ul>
 	</div>
 
-	<div class="environment select">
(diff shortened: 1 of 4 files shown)


szarnyasg:
Thanks, will look into this!

szarnyasg:
Just to confirm, merging this indeed fixed the installation page.

PR #1251, opened 2023-09-25
Document the use of the catalog search path. Fixes #1160

### LINKED ISSUE #1160: Attach docs should reference `search_path` setting
https://github.com/duckdb/duckdb-web/blob/c253b025f868e0926eef9be3526ec41fadcf6b1b/docs/sql/statements/attach.md?plain=1#L57C119-L57C119

This mentions the `catalog search path` but I think that's a little too vague still
It could use an example using either `set schema` or `set search_path` directly to drive this point home.

An example of using `search_path` would be:
```sql
D attach ':memory:' as db1;
D attach ':memory:' as db2;
D create table db1.tbl1 (i integer);
D create table db2.tbl2 (j integer);
D set search_path='db1,db2';
D select * from tbl1;
┌────────┐
│   i    │
│ int32  │
├────────┤
│ 0 rows │
└────────┘
D select * from tbl2;
┌────────┐
│   j    │
│ int32  │
├────────┤
│ 0 rows │
└────────┘
```

Both of these SELECT queries would fail if search_path was not set, because `tbl1` and `tbl2` would need to be qualified explicitly

szarnyasg:


docs/sql/statements/attach.md
```diff
@@ -91,6 +97,24 @@ CREATE TABLE new_db.tbl(i INTEGER);
...
+
+```sql
+ATTACH ':memory:' AS db1;
+ATTACH ':memory:' AS db2;
+CREATE table db1.tbl1 (i INTEGER);
+CREATE table db2.tbl2 (j INTEGER);
+-- reference the tables using their fully qualified name
+SELECT * FROM db1.tbl1;
+SELECT * FROM db2.tbl2;
+-- or set the search path and references the tables using their name
```
Tishj, line 112: reference*

I know this was my example, just catching the typo now 😅

PR #1257, opened 2023-09-25
Add page stub for Spark API

szarnyasg:
cc @hannes @Tishj

### DIFF
--- a/docs/api/python/spark_api.md
+++ b/docs/api/python/spark_api.md
@@ -1,0 +1,14 @@
+---
+layout: docu
+title: Spark API
+---
+
+DuckDB offers a PySpark-compatible API.
+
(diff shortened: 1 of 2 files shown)


docs/api/python/spark_api.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+DuckDB offers a PySpark-compatible API.
+
+> This extension is currently in an experimental state. Feel free to try it out, but be aware some things may not work as expected.
+
+## Usage
+
+```python
+import duckdb.experimental.spark
```
Mause, line 13: Can we add the example from the release post here? And maybe link to the appropriate pyspark docs?

Mause (changes requested):

PR #1258, opened 2023-09-26
Nits 2023 09 26

szarnyasg:
Spell AMD64 and ARM64 (AArch64) as they should be spelt (see https://en.wikipedia.org/wiki/AArch64, https://en.wikipedia.org/wiki/X86-64).


_includes/installation.html
```diff
@@ -64,7 +64,7 @@ <h3>Installation</h3>
 			Linux (AMD64): <a href="https://github.com/duckdb/duckdb/releases/download/v{{ site.currentduckdbversion }}/duckdb_cli-linux-amd64.zip" target="_blank">https://github.com/duckdb/duckdb/releases/download/v{{ site.currentduckdbversion }}/duckdb_cli-linux-amd64.zip</a>
 			<br/><br/>
 
-			Linux (Arm64/aarch64): <a href="https://github.com/duckdb/duckdb/releases/download/v{{ site.currentduckdbversion }}/duckdb_cli-linux-aarch64.zip" target="_blank">https://github.com/duckdb/duckdb/releases/download/v{{ site.currentduckdbversion }}/duckdb_cli-linux-aarch64.zip</a>
+			Linux (ARM64/Aarch64): <a href="https://github.com/duckdb/duckdb/releases/download/v{{ site.currentduckdbversion }}/duckdb_cli-linux-Aarch64.zip" target="_blank">https://github.com/duckdb/duckdb/releases/download/v{{ site.currentduckdbversion }}/duckdb_cli-linux-Aarch64.zip</a>
```
Mause, line 67: This isn't correct per the page you linked?

Mause (changes requested):

_includes/installation.html
szarnyasg: Very good point, thanks ... I'd spell everything differently than how they're supposed to be, so it's difficult to adjust.

szarnyasg:
Pushed a bunch of small fixes

Mause (approved):

PR #1278, opened 2023-09-27
Spin out Media section to its own page

### LINKED ISSUE #1211: Create separate Media page
The Media page is currently just [a section in the Why DuckDB page](https://duckdb.org/why_duckdb.html#media).
We have more talks and media appearances, so spinning this out into its own page (with a nicer formatting) would make sense.

szarnyasg:
Fixes #1211

### DIFF
--- a/_data/menu_side.json
+++ b/_data/menu_side.json
@@ -77,6 +77,10 @@
       "url": "/why_duckdb"
     },
     {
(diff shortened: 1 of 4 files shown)


szarnyasg:
It doesn't look great but we'll rework its layout later.

PR #1280, opened 2023-09-27
Add rules for case sensitivity

### LINKED ISSUE #1129: Document case sensitivity of identifiers in DuckDB
This is something I've seen some questions about - we should document how DuckDB works with regards to case insensitivity of identifiers (case is maintained as entered by the user, but case is insensitive so "hello" and "HelLo" are the same identifier regardless of quotes).

szarnyasg:
Fixes #1129

### DIFF
--- a/_data/menu_docs_dev.json
+++ b/_data/menu_docs_dev.json
@@ -926,6 +926,10 @@
             "url": "pragmas"
           },
           {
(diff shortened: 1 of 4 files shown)


Mause (approved):
🔥

Mytherin:
Looks great, thanks!

PR #1316, opened 2023-10-03
Add script to redirect between versions

szarnyasg:
Sketch script that would redirect from e.g., 0.9.1 to 0.9.0.

Tested locally. It works but I noticed a small issue:

from 
http://localhost:4000/docs/archive/0.9.1/guides/import/parquet_import

it redirects to
http://localhost:4000/docs/archive/0.9.0/guides/import/parquet_import.html

(note the .html)

@Mause WDYT?

### DIFF
--- a/scripts/redirect.py
+++ b/scripts/redirect.py
@@ -1,0 +1,32 @@
+import frontmatter
+import sys
+from glob import glob
+from pathlib import Path
+import re
+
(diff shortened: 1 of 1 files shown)


Mause (changes requested):
As discussed, other ways will likely work better

PR #1319, opened 2023-10-03
Simplify R docs by using library(duckdb)

szarnyasg:
I use `library(duckdb)` in the R docs which allowed me to remove `duckdb::` from a few places.
@Tmonster @hannes WDYT?


Tmonster (approved):
lgtm!

PR #1323, opened 2023-10-06
Specify exact adbc-driver-manager version for v0.9.0

Mause:
Fixes https://github.com/duckdb/duckdb/issues/9239

This is already mentioned earlier in the file, but worth mentioning here too in the actual installation step


pdet:
From the previously fixed CI, I think we had `adbc_driver_manager==0.6.0` set to 0.9

PR #1362, opened 2023-10-12
minor improvements

Franz-Kafka:
- css improvement
- make tables auto width, using <div class="narrow_table"></div> before table starts
- smaller tables will be automatically auto width
- sticky headline bar in documentation


szarnyasg:
Great, thanks!

PR #1388, opened 2023-10-16
Avoid redirect by appending "/" for docs link

smoy:
This eliminate the un-necessary redirect. 

However, it seems maintainer that have access to CloudFlare for dubckdb.org account should configure http -> https redirect to avoid folks accessing straight http protocol for the domain.


szarnyasg:
Thanks! Will look into the Cloudflare configuration.

PR #1423, opened 2023-10-20
github star counter on nav bar

Franz-Kafka:
- added star counter to the nav-bar
- small text color fix on mobile


szarnyasg:
@Franz-Kafka could you please adjust the design such that the star is also part of the link and changes color on hover?

szarnyasg:
I'll merge it now to see whether it works with the CI pipeline. We can iterate on the design later.

PR #1430, opened 2023-10-23
Revert "Increase required Python version from 3.7 to 3.8"

Mause:
Reverts duckdb/duckdb-web#1304

This was incorrectly bumped, we support >=3.7, see https://github.com/duckdb/duckdb/blob/main/tools/pythonpkg/setup.py#L351


szarnyasg:
Thanks for pinging me - let me discuss this with the team in Amsterdam

szarnyasg:
Discussed: Yes, DuckDB should work with 3.7

PR #1444, opened 2023-10-26
Apply narrow_table format to ASOF join blog post

szarnyasg:


szarnyasg:
The formatting does not work yet for blog posts but I'm adding these comments as they do not cause any regression.

PR #1456, opened 2023-10-29
Update timezone of Greenland

Flogex:
Greenland (or at least parts of it) [has changed its timezone](https://www.timeanddate.com/news/time/greenland-change-timezone.html) from UTC-3 to UTC-2. This is reflected now when using `pg_timezone_names`. This PR updates the reference list.

America/Nuuk and America/Godthab are synonymous as Godthab is the old name of Nuuk.


Mause:
@hawkfish this okay to merge you think?

PR #1468, opened 2023-11-02
Add Rust to installation page

szarnyasg:
First attempt at adding Rust to the installation page. Maybe the example is a bit too elaborate now ...

### DIFF
--- a/_includes/installation.html
+++ b/_includes/installation.html
@@ -21,6 +21,7 @@
 			<li data-id=".r">R</li>
 			<li data-id=".java">Java</li>
 			<li data-id=".js">Node.js</li>
+			<li data-id=".rust">Rust</li>
 			<li data-id=".cplusplus">C/C++</li>
 			<li data-id=".odbc">ODBC</li>
 		</ul>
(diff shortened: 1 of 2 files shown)


Mause:
Have you tested this? I'm not sure if the js requires changes for this

szarnyasg:
Yes, it works fine without adjusting the underlying JS:
Screenshot 2023-11-02 at 12 04 23

carlopi:
One thing on this page is that I am not sure whether it makes sense to point to bleeding edge at all, given for a bunch of platforms (node, rust and R) the underlying library will still be on duckdb 0.9.1, and we probably want most users to be on latest distributed release.

Maybe a lower section? Like having environment / package / platform, then Installation, usage example, Installation for bleeding edge package. Or something else

szarnyasg:
@Tmonster This PR adds Swift to the installation page based on https://github.com/duckdb/duckdb-swift#installation:
https://github.com/duckdb/duckdb-web/pull/1468/commits/929f73fcf92d995205590f872be7ecd77964f69b

Does this look good to you?

Tmonster:
It's what's been on the 'duckdb-swift' repository for a while. If I'm not mistaken there is a difference between adding duckdb to a "package" vs adding duckdb to a "project". I'll take some time this week to look into this a bit more and suggest changes

PR #1475, opened 2023-11-03
fix all -> almost every

Tmonster:
polars join at 500mb they are technically faster


szarnyasg:
big plus point for this!

PR #1478, opened 2023-11-04
Python and Jupyter updates

### LINKED ISSUE #1477: Issue found on page 'Jupyter Notebooks'
_Please describe the issue on the page and include the "Page URL" link shown below._

Page URL: <https://duckdb.org/docs/guides/python/jupyter.html>

The page says:

> To return query results into a Pandas dataframe for future usage, use << as an assignment operator. This can be used with both the %sql and %%sql Jupyter magics.

This is not true, the `<<` operator does not return a dataframe. It return a resultset that can be later converted to pandas/polars data. See the error message I got when I invoked `hist()` on the dataframe:

```
File /opt/homebrew/lib/python3.11/site-packages/sql/run/resultset.py:234, in ResultSet.__getattr__(self, attr)
    228 def __getattr__(self, attr):
    229     err_msg = (
    230         f"'{attr}' is not a valid operation, you can convert this "
    231         "into a pandas data frame by calling '.DataFrame()' or a "
    232         "polars data frame by calling '.PolarsDataFrame()'"
    233     )
--> 234     raise AttributeError(err_msg)
```

What the page means is correct only if `%config SqlMagic.autopandas = True` is set.

szarnyasg:
Fixes #1477

### DIFF
--- a/docs/guides/python/jupyter.md
+++ b/docs/guides/python/jupyter.md
@@ -105,7 +105,7 @@ ORDER BY ALL DESC
 LIMIT 5
 ```
 
-To return the query results into a Python variable, use `<<` as an assignment operator.
+To store the query results in a Python variable, use `<<` as an assignment operator.
 This can be used with both the `%sql` and `%%sql` Jupyter magics.
 ```sql
 %sql res << SELECT 'Off and flying!' AS a_duckdb_column
--- a/docs/guides/python/jupyter.md
+++ b/docs/guides/python/jupyter.md
@@ -105,11 +105,12 @@ ORDER BY ALL DESC
 LIMIT 5
 ```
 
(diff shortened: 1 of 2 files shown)


docs/guides/python/jupyter.md
```diff
@@ -105,11 +105,12 @@ ORDER BY ALL DESC
 LIMIT 5
 ```
 
-To return query results into a Pandas dataframe for future usage, use `<<` as an assignment operator.
+To return the query results into a Python variable, use `<<` as an assignment operator.
```
Mause, line 108:
```suggestion
To store the query results in a Python variable, use `<<` as an assignment operator.
```

szarnyasg:
Thanks!

PR #1490, opened 2023-11-06
Update `row` function docs

Tishj:
In <https://github.com/duckdb/duckdb/pull/8942> we changed the behavior surrounding struct casting and the `row` function.
This PR updates the docs to reflect this.


szarnyasg:
Thanks!

PR #1502, opened 2023-11-09
added url parameter to installation page

Franz-Kafka:
- added url parameter to installation page. so you can send links to exact installation instructions.


szarnyasg:
Thanks!

PR #1516, opened 2023-11-14
Clarify why DETACH is required in scripts

szarnyasg:
Added a clarification on why invoking `DETACH` may be nessary even if `close` is called.

@Tishj does the text added by this PR make sense?

### DIFF
--- a/docs/sql/statements/attach.md
+++ b/docs/sql/statements/attach.md
@@ -39,6 +39,8 @@ USE file;
 
 The `DETACH` statement allows previously attached database files to be closed and detached, releasing any locks held on the database file.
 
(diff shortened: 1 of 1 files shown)


Tishj:
Yea I think that's correct

From what I gather, the DatabaseInstance holds the DatabaseManager, which holds a CatalogSet `databases`.
AttachedDatabase is a CatalogEntry, stored in this catalog set.

An AttachedDatabase holds a StorageManager, which has a virtual LoadDatabase method, the `SingleFileStorageManager` has a BlockManager, which is a `SingleFileBlockManager` and THIS holds the filehandle of the database 😅 

I don't think we want to specify all these implementation details though

Tishj:
Though I think if the last connection to a database is closed, it should close the database.
We have some caching mechanism to prevent opening the database twice, but this holds a weak_ptr, so that shouldn't be a problem?

PR #1519, opened 2023-11-14
Rework text type page

### LINKED ISSUE #1513: [duckdb/#9668] - Add length check and space padding to CHAR type needs documentation
See https://github.com/duckdb/duckdb/issues/9668

szarnyasg:


szarnyasg:
Removing space-padded CHAR(n) and BPCHAR(n) for the time being. If we ever want to add those back, there's the text:

````
## Space-Padded Strings

The type `CHAR(n)` and its alias `BPCHAR(n)` store fixed-length character strings. If the string is shorter than length _n_, it will be padded with space characters. If it is longer, it will be truncated to length _n_.

```sql
-- returns 'abc  ' (with two trailing spaces)
SELECT 'abc '::CHAR(5) AS s;
-- returns 'abcde'
SELECT 'abcdef'::CHAR(5) AS s;
-- returns 'abcde'
SELECT 'abcde'::CHAR(5) AS s;
```
````

PR #1523, opened 2023-11-14
docs: fix grammar in `extensions/overview.md`

peter-gy:
The sentence 

> This allows users to extend the functionaly of DuckDB such to suit their domain the best.

was not very clear.

Changed it to

> This allows users to extend the functionality of DuckDB such that it suits their domain the best.


szarnyasg:
Thanks!

PR #1548, opened 2023-11-16
fix insertCSVFromPath in docs

candux:
insertCSVFromPath() is a function of the connection and not the database. Error message:
TypeError: Cannot read properties of undefined (reading 'columns')


szarnyasg:
That's right, thanks for pointing this out!

PR #1560, opened 2023-11-20
Add an example of using a macro argument as a list 

rdavis120:
Add an example of using a macro argument as a list

### DIFF
--- a/docs/sql/statements/create_macro.md
+++ b/docs/sql/statements/create_macro.md
@@ -41,7 +41,7 @@ CREATE OR REPLACE TEMP MACRO dynamic_table(col1_value, col2_value) AS TABLE
     SELECT col1_value AS column1, col2_value AS column2 
     UNION ALL 
     SELECT 'Hello' AS col1_value, 456 AS col2_value;
--- pass an argument as a list: select * from get_users([1, 5])
+-- pass an argument as a list: SELECT * FROM get_users([1, 5])
 CREATE MACRO get_users(i) AS TABLE SELECT * FROM users WHERE uid IN (select unnest(i))
 ```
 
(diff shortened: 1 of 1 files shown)


docs/sql/statements/create_macro.md
```diff
@@ -41,6 +41,8 @@ CREATE OR REPLACE TEMP MACRO dynamic_table(col1_value, col2_value) AS TABLE
     SELECT col1_value AS column1, col2_value AS column2 
     UNION ALL 
     SELECT 'Hello' AS col1_value, 456 AS col2_value;
+-- pass an argument as a list: select * from get_users([1, 5])
```
szarnyasg, line 44:
```suggestion
-- pass an argument as a list: SELECT * FROM get_users([1, 5])
```

docs/sql/statements/create_macro.md
```diff
@@ -41,6 +41,8 @@ CREATE OR REPLACE TEMP MACRO dynamic_table(col1_value, col2_value) AS TABLE
     SELECT col1_value AS column1, col2_value AS column2 
     UNION ALL 
     SELECT 'Hello' AS col1_value, 456 AS col2_value;
+-- pass an argument as a list: SELECT * FROM get_users([1, 5])
+CREATE MACRO get_users(i) AS TABLE SELECT * FROM users WHERE uid IN (select unnest(i))
```
szarnyasg, line 45:
```suggestion
CREATE MACRO get_users(i) AS TABLE SELECT * FROM users WHERE uid IN (SELECT unnest(i));
```

szarnyasg:
I made some minor adjustment but otherwise it looks good, thanks!

PR #1595, opened 2023-11-23
Syntax highlight fixes

szarnyasg:


szarnyasg:
For the meaning of Rouge's CSS classes, see `codehilite` themes, e.g.
https://github.com/develephant/mkdocs-codehilite-themes/blob/master/css/vs.css

PR #1608, opened 2023-11-25
Add note about using the list aggregate function to create lists from columns

Flogex:
After hearing a talk yesterday about DuckDB's list functions, I started playing with it. However, it took me a very long time to figure out how to create a list from a table column since I tried to somehow do it using the `list_value` function. Only by chance I stumbled across the `list` aggregate function which does the job.

To help people like myself in the future, I added a short note to the _Creating Lists_ section that refers to the `list` aggregate function.


szarnyasg:
Good point, thanks a lot!

PR #1614, opened 2023-11-25
Add description of two list functions

Flogex:
I added a verbal description for the `list_distance` and `list_dot_product` functions.


szarnyasg:
Thanks!

PR #1623, opened 2023-11-25
Add TPC-H data generator parameters

### LINKED ISSUE #1443: suggest add tpch extension document for dbgen() function
it has 7 parameters indeed
```
D load tpch;
D call dbgen(sf=1,a=1) ;
Error: Binder Error: Invalid named parameter "a" for function dbgen
Candidates:
    step UINTEGER
    sf DOUBLE
    overwrite BOOLEAN
    suffix VARCHAR
    catalog VARCHAR
    schema VARCHAR
    children UINTEGER
```

szarnyasg:
Fixes #1443

### DIFF
--- a/docs/extensions/tpcds.md
+++ b/docs/extensions/tpcds.md
@@ -40,3 +40,8 @@ PRAGMA tpcds(8);
 │ ought        │       -10994052.78 │
 └──────────────┴────────────────────┘
 ```
+
+## Limitations
+
+The `tpchds({query_id})` function runs a fixed TPC-DS query with pre-defined bind parameters (a.k.a. substitution parameters).
+It is not possible to change the query parameters using the `tpcds` extension.
--- a/docs/extensions/tpch.md
+++ b/docs/extensions/tpch.md
@@ -17,11 +17,15 @@ LOAD tpch;
 
 ## Usage
 
+### Generating Data
+
 To generate data for scale factor 1, use:
 
 ```sql
 CALL dbgen(sf = 1);
 ```
+
+### Running a Query
 
 To run a query, e.g., query 4, use:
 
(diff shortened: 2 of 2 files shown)


szarnyasg:
@pdet can you please review this?

szarnyasg:
A big thing missing in the TPC-H documentation is:
- what substitution parameters are used when we invoke PRAGMA query(x);
- is it possible to use different substitution parameters?

pdet:
I'm not sure if I would described children and step like that. In practice they are basically a way of partitioning the TPC-H data generation.

`children` is the number of partitions and `step` is which partition will be generated in that run.

Also `children` does not have a limit

For example, to generate all the sf 1 in 100 partitions we cold do:
```sql
CALL dbgen(sf=1, children =100, step = 0);
CALL dbgen(sf=1, children =100, step = 1);
CALL dbgen(sf=1, children =100, step = 2);
...
CALL dbgen(sf=1, children =100, step = 99);
```

pdet:
> A big thing missing in the TPC-H documentation is:
> 
> * what substitution parameters are used when we invoke PRAGMA query(x); ?
> * is it possible to use different substitution parameters?

what you mean with substitution parameter?

szarnyasg:
Query parameters (a.k.a. bind parameter) that we substitute into the TPC-H query template

pdet:
Ah, 
I guess you mean 
```sql
pragma tpch($query_number)
```
right?

So the tpch extension has the following functions/paramenters
dbgen:
  sf
  overwrite
  catalog
  schema
  suffix
  children
  step

tpch_queries:
query_nr
query

tpch_queries:
query_nr
query
so the query has:
```cpp
	names.emplace_back("query_nr");
	return_types.emplace_back(LogicalType::INTEGER);

	names.emplace_back("query");
	return_types.emplace_back(LogicalType::VARCHAR);
```

pdet:
Ah, 
I guess you mean 
```sql
pragma tpch($query_number)
```
right?

So the tpch extension has the following functions/arguments:
dbgen:
  sf
  overwrite
  catalog
  schema
  suffix
  children
  step

tpch:
  (unnamed - query number): integer

We also have functions that take no arguments, but return query and query-answers information on tables, the functions and their columns are: 

tpch_queries:
  query_nr
  query

tpch_answers:
  query_nr
  scale_factor
  answer

szarnyasg:
Thanks @pdet !

PR #1653, opened 2023-11-27
Decrease line height to match CLI formatting

szarnyasg:
Decreased the CSS line-height for code from 1.4 to 1.2

@jonathanauch what do you think about this change?

## Before
Screenshot 2023-11-27 at 10 16 19

## After
Screenshot 2023-11-27 at 10 16 28

### DIFF
--- a/css/docu.scss
+++ b/css/docu.scss
@@ -30,7 +30,7 @@ pre {
     border-radius: 2px;
     -moz-border-radius: 2px;
     -webkit-border-radius: 2px;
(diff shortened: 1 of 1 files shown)


jonathanauch:
never knew, that it had to be a uninterrupted line. 
this is totally fine with me.

szarnyasg:
Checking on other pages, it seems more cramped than it should.
Screenshot 2023-11-27 at 10 44 45

PR #1674, opened 2023-11-28
json_extract_string list wildcard example

eaubin:
It's not obvious which parts of jsonpath are supported from the current docs, but it turns out list wildcards are. This adds one example.


szarnyasg:
Thanks!

PR #1687, opened 2023-11-30
Document built-in extensions

### LINKED ISSUE #917: Document extensions shipped/loaded by default
An idea: document which extension is loaded by default. The challenge is that these vary from distribution (e.g. the R package doesn't ship `icu` by default).

szarnyasg:
Fixes #917

### DIFF
--- a/docs/extensions/official_extensions.md
+++ b/docs/extensions/official_extensions.md
@@ -37,15 +37,15 @@ We summarize the main distributions in the table below.
 
 | Extension Name | CLI (duckdb.org) | CLI (Homebrew) | Python | R | Java | Julia | Node.JS
 |------|------|------|---|---|---|---|---|
-| [autocomplete](autocomplete) | yes | yes | no  | no  | no  | no  | no  |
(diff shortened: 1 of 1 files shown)


docs/extensions/official_extensions.md
```diff
@@ -25,3 +27,25 @@ title: Official Extensions
...
+## Default Extensions
+
+Different DuckDB clients ship a different set of extensions.
+We summarize the main distributions in the table below.
+
+<div class="narrow_table"></div>
+
+| Extension Name | CLI (duckdb.org) | CLI (Homebrew) | Python | R | Java | Julia | Node.JS
+|------|------|------|---|---|---|---|---|
+| [autocomplete](autocomplete) | yes | yes | no  | no  | no  | no  | no  |
```
Mause, line 40: Maybe drop the `no`s, leaving the `yes`s? Would reduce the visual noise

szarnyasg: good point, did this

PR #1688, opened 2023-11-30
Rework INSERT INTO page

### LINKED ISSUE #1342: Improve documentation on upserts (INSERT OR REPLACE)
The current documentation is very shallow:

```sql
-- or update the table with the new values instead
INSERT OR REPLACE INTO tbl(i) VALUES(1);
```

szarnyasg:
Fixes #1342

One remark – The previous page said:
> If a conflict does not meet this condition, an error will be thrown instead, and the entire operation is aborted.

This seem to be incorrect. Given a sample data set:

```sql
CREATE TABLE tbl (i INT PRIMARY KEY, j INT UNIQUE, k INT);
INSERT INTO tbl VALUES (1, 20, 300);
```

All of these queries finish successfully:

```sql
INSERT INTO tbl VALUES (1, 40, 700) ON CONFLICT (i) DO UPDATE SET k = 3 * EXCLUDED.k WHERE k < 100;
INSERT INTO tbl VALUES (1, 40, 700) ON CONFLICT (i) DO UPDATE SET k = 5 * EXCLUDED.k WHERE EXCLUDED.k < 100;
INSERT INTO tbl VALUES (1, 40, 700) ON CONFLICT (i) DO UPDATE SET k = 7 * EXCLUDED.k WHERE false;
```

As part of working on this PR, I created https://github.com/duckdb/duckdb/pull/9862 into mainline.

### DIFF
--- a/docs/sql/statements/insert.md
+++ b/docs/sql/statements/insert.md
@@ -14,13 +14,13 @@ INSERT INTO tbl VALUES (1), (2), (3);
 -- insert the result of a query into a table
 INSERT INTO tbl SELECT * FROM other_tbl;
 -- insert values into the "i" column, inserting the default value into other columns
-INSERT INTO tbl(i) VALUES (1), (2), (3);
+INSERT INTO tbl (i) VALUES (1), (2), (3);
 -- explicitly insert the default value into a column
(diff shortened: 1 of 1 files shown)


szarnyasg:
Pushing this into main in the current state as I want to do some restructuring on the site

PR #1690, opened 2023-11-30
Clarifying json and parquet are always built in.

marhar:


szarnyasg:
@marhar: thanks for the PR. Which platform are you running on?

On MacOS/arm64, `json` is not part of the R package, see:
```R
library("duckdb")
con <- dbConnect(duckdb())
dbExecute(con, "LOAD json")
```
```text
Error: rapi_execute: Failed to run query
Error: IO Error: Extension ".../Library/Application Support/org.R-project.R/R/duckdb/.duckdb/extensions/v0.9.2/osx_arm64/json.duckdb_extension" not found.
Extension "json" is an existing extension.

Install it first using "INSTALL json".
```

PR #1692, opened 2023-12-01
Add link from Window clause to Window functions

rdavis120:


szarnyasg:
It's a good addition, thanks!

PR #1716, opened 2023-12-06
Add note on CSV delimiter length

szarnyasg:
Zero-length delimiter is fine but >1 delimiter strings are no longer allowed, see:
```sql
⚪◗ create table t(x int);
⚪◗ copy t to 't.csv' (delimiter '');
⚪◗ copy t to 't.csv' (delimiter 'hello');
Error: Invalid Input Error: The delimiter option cannot exceed a size of 1 byte.
```


pdet:
The same is true for quotes and escapes.

szarnyasg:
Thanks!

PR #1740, opened 2023-12-10
Add typecasting matrix

### LINKED ISSUE #836: Document what casts are implemented for each type
We should probably add something similar to what we did for the `core_functions` in that every type should likely have a `json` file describing what it can cast to, then we can leverage that to also generate a table of type -> type casts

That would also help development, to know which casts are missing

szarnyasg:
Fixes #836


mootari:
@szarnyasg Do you by any chance recall how you produced the source data for the matrix?

szarnyasg:
@mootari it was a buggy script shell. I don't think I have it anymore, and I've been wanting to rework this matrix for some time but it has been in the backlog. The image was created as a screenshot from Google Sheets.

PR #1749, opened 2023-12-11
MySQL extension is also available for v0.9.2

Mytherin:
And prefer to use `mysql` over `mysql_scanner` as the extension name


szarnyasg:
Great, thanks!

PR #1754, opened 2023-12-11
correct typo in DO UPDATE SET

rdavis120:
The columns in "do update set" does not support the parenthesis.


szarnyasg:
Good point, thanks!

PR #1756, opened 2023-12-12
Deprecate PRAGMA x=y in favor of SET x=y

### LINKED ISSUE #1753: Use SET instead of PRAGMA
We are slowly deprecating PRAGMA, this should be reflected in the documentation.

szarnyasg:
Fixes #1753

@Mytherin we discussed deprecating `PRAGMA` in the documentation where possible. This is what you had in mind?


Mytherin:
Yes, thanks!

PR #1761, opened 2023-12-13
Fix instructions for generating query graph

jhrcek:
The original instructions at the bottom of [this page](https://duckdb.org/dev/profiling.html) don't work for me.
I'm getting error saying that generate_querygraph.py doesn't exist.

I looked into my local installation of duckdb within 
/home/jhrcek/.local/lib/python3.10/site-packages/duckdb and I didn't find any file named `generate_querygraph.py `

Instead there's this (running ls in the above folder):
```
├── query_graph
│   ├── __main__.py
│   └── __pycache__
│       └── __main__.cpython-310.pyc
```

When executed using `python -m duckdb.query_graph /path/to/file.json` it works for me (generates html file and opens it in my browser).


Mause:
This was outdated yeah, thanks for the fix!

PR #1778, opened 2023-12-19
Add pybind11 to Python installation instructions (workaround)

szarnyasg:


Mause:
@szarnyasg pybind11 should be automatically installed, are you seeing otherwise?

szarnyasg:
I spun up many Ubuntu Graviton boxes in the last few days and I always had to manually install pybind11 to build the DuckDB Python client from source.

Mause:
Are you just referring to the pybind warning I added?

szarnyasg:
```bash
$ docker run -it ubuntu:22.04 /bin/bash
```
```bash
root@baca8108deca:/# apt update && apt install -y python3-pip
root@baca8108deca:/# python3
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pybind11
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'pybind11'
```

If I try to build DuckDB using setup.py, I get an error: https://gist.github.com/szarnyasg/8fa52177a069d5defdc9717904cb7259

> ModuleNotFoundError: No module named 'pybind11'

Mause:
To clarify, it's supposed to be automatically installed as part of the build process (it won't be available right after you install python). The other error may just be due to outdated setuptools or similar, but as the warning at the start says, `setup.py` is deprecated

szarnyasg:
Right, but it is not installed when I ran the `setup.py` script. I think something in the [pythonpkg README](https://github.com/duckdb/duckdb/tree/main/tools/pythonpkg) needs fixing because:

1) It does not suggest that using setup.py is deprecated

2) The first recommended option, `BUILD_PYTHON=1` does not work with `BUILD_HTTPS`, which is why [this whole section](https://duckdb.org/dev/building.html#building-the-httpfs-extension-and-python-package-on-macos) was created in the first place. I can imagine numerous workarounds such as building DuckDB in two stages: first the httpfs extension, then the Python client, then installing the extension, etc.

I'm not an expert on the Python build and installation process – Please advise on how to improve the instructions.

PR #1779, opened 2023-12-19
Update building.md with pip

Mause:
Edit (Gabor): followup of #1778


szarnyasg:
Yes, this makes perfect sense. I tried it and it worked.

PR #1810, opened 2024-01-03
Add remark on memory limit

szarnyasg:
I added a remark on the memory limit.

Screenshot 2024-01-03 at 10 55 30

### DIFF
--- a/docs/sql/pragmas.md
+++ b/docs/sql/pragmas.md
@@ -65,12 +65,18 @@ PRAGMA show('table_name');
 
 ### Memory Limit
 
-Set the memory limit:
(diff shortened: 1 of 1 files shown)


szarnyasg:
@Mytherin @lnkuiper @hawkfish do you have any comments or further text to add here?

lnkuiper:
The `list` aggregate function allocations are routed through the buffer manager, but are not spillable.

But maybe that's too much info for the average user.

szarnyasg:
Thanks for the clarification! I think we can err on the side of simplicity here, so let's leave the `list` function in the remark.

PR #1884, opened 2024-01-15
Adding the `UHUGEINT` type

nickgerrets:
Also fixes the table for `foreach` in sqllogictest/loops.md, as I noticed it was broken.


szarnyasg:
Great, thanks!

PR #1896, opened 2024-01-17
Tweak search tokenisation

### LINKED ISSUE #1865: Search function does not work reliably for terms with underscore
As observed in #1855, we have some configuration flags that are well-documented but they do not show up when using the documentation's search feature.
* `row_group_size`
* `preserve_insertion_order`

They are part of the [`data/search_data.json`](https://github.com/duckdb/duckdb-web/blob/main/data/search_data.json) file, and some other terms with underscore (e.g., `repeat_row`) show up just fine.

Mause:
Fixes #1865

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -12,12 +12,15 @@ xhr.onreadystatechange = function(event) {
 }
 xhr.send();
 
+const tokenize = (string) => string.split(/[\s-]+/); // search query tokenizer
+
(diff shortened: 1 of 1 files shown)


szarnyasg (approved):
It works great, thanks!

image

szarnyasg:
Thanks for the fix!

Mause:
May require further tweaks (the default regex it uses is much more complex) but we'll see how we go!

PR #1913, opened 2024-01-18
Adding Qualify section to the window functions docs

javier:
I think Qualify should be featured as part of the window functions section. Adding a link


szarnyasg:
@javier indeed, thanks! I pushed some minor adjustments and the PR is good to go in.

Issue #1928, opened 2024-01-23
Auto-update docs jobs is failing


szarnyasg:
https://github.com/duckdb/duckdb-web/actions/runs/7606692874/job/20712873266

```bash
  > ×	[main] command [./scripts/generate_all_docs.sh duckdb] exited with code 1. message: + '[' duckdb = '' ']'
```

Mause:
Just due to the extension build failing in the main repo:

```
>> Error: near line 3: HTTP Error: Failed to download extension "httpfs" at URL "http://extensions.duckdb.org/95e6a0369f/linux_amd64_gcc4/httpfs.duckdb_extension.gz"
    >> Extension "httpfs" is an existing extension.
    >> 
    >> Are you using a development build? In this case, extensions might not (yet) be uploaded.
    >> Error: near line 4: IO Error: Extension "/home/runner/.duckdb/extensions/95e6a0369f/linux_amd64_gcc4/httpfs.duckdb_extension" not found.
    >> Extension "httpfs" is an existing extension.
    >> 
    >> Install it first using "INSTALL httpfs".
```

szarnyasg:
Ah, alright. One thing that is confusing in the current documentation of the scripts is that some places say "use the nightly build" while others say "use the debug build" (https://github.com/duckdb/duckdb-web/blob/main/scripts/generate_all_docs.sh#L4-L11). Nightly should be sufficient, right?

Mause:
Debug isn't required no, where does it say that?

szarnyasg:
https://github.com/duckdb/duckdb-web/blob/dbb7e10f09f63af61f75d09b2af134e4c9af1da7/scripts/generate_all_docs.sh#L4-L11

Mause:
Hmm okay. Do you remember why you added that?

szarnyasg:
Probably by mistake, will edit it out now.

szarnyasg:
No, actually, reading further down the script, we run the following:
https://github.com/duckdb/duckdb-web/blob/dbb7e10f09f63af61f75d09b2af134e4c9af1da7/scripts/generate_all_docs.sh#L17

szarnyasg:
And I think this mislead me. Okay, will amend the README.

szarnyasg:
Okay, so I think it works with both the `debug` and the `release` builds. We just have to settle on one. Likely `release` is the simpler choice.

PR #1929, opened 2024-01-23
Scripts: update documentation on how to run generation scripts

szarnyasg:
Towards #1928

### DIFF
--- a/scripts/generate_all_docs.sh
+++ b/scripts/generate_all_docs.sh
@@ -3,18 +3,18 @@ set -xeuo pipefail
 
 if [ "${1-}" = "" ]
 then
-    echo >&2 "Usage: $0 duckdb_dir_with_debug_duckdb"
+    echo >&2 "Usage: $0 duckdb_dir_with_release_duckdb"
     exit 1
 fi
 
 DUCKDB=$1;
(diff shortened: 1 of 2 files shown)


scripts/generate_all_docs.sh
```diff
@@ -3,18 +3,18 @@ set -xeuo pipefail
...
-    echo >&2 "Usage: $0 duckdb_dir_with_debug_duckdb"
+    echo >&2 "Usage: $0 duckdb_dir_with_release_duckdb"
     exit 1
 fi
 
 DUCKDB=$1;
 echo "Generating docs using duckdb source in $DUCKDB"
 
-python3 ./scripts/generate_config_docs.py $DUCKDB/build/debug/duckdb
+python3 ./scripts/generate_config_docs.py $DUCKDB/build/release/duckdb
```
Mause, line 13: If you're going to change these, you need to update the workflows that populate them, else the autoupdate will break

Mause (changes requested):

scripts/generate_all_docs.sh
szarnyasg: Good point, thanks! Fix is incoming/

PR #1942, opened 2024-01-25
Postgres connection can be specified in environment variables.

marhar:
Very useful if you are running in a container or other environment where you don't want to manage the connection specification in the application.


szarnyasg:
Thanks, it's a great addition! I adjusted the formatting to fit the rest of the documentation.

PR #1959, opened 2024-01-26
Noting configuration is in duckdb_configuration().

marhar:
It is in the examples below, but not stated explicitly.


szarnyasg:
Thanks!

PR #1978, opened 2024-01-29
Add instructions for building in-tree extensions to guide

szarnyasg:


szarnyasg:
Most of this PR is fixing linter issues. The Markdown linter started complaining about some sort of issues after a version upgrade and I just decided to fix all issues then.

PR #2005, opened 2024-01-31
Document current_catalog()

elefeint:
I only backported it to 0.9.2, but I think the function got added earlier.


szarnyasg:
Thank you! Backporting 0.9.2 is fine, in general we don't do much maintenance on versions other than stable and dev.

PR #2009, opened 2024-02-01
Performance Guide: Add instructions to use CSV sniffer manually

### LINKED ISSUE #2006: Add CSV sniffer instructions to Performance Guide
@pdet's comment from LinkedIn

> For small files, the difference should be much lower ( ~30%). One of the main performance differences is that parallelism got so much better, but that only really hits for files of at least 8Mb. One manual optimization that might have a big impact on small files is turning the sniffer off. If all files have the same columns/dialect you can get the sniffer options with select prompt from sniff_csv('file_path') and use that :-)

szarnyasg:
Fixes #2006


pdet (approved):
Looks excellent!

PR #2044, opened 2024-02-06
Add tldr field to blog posts' front matter

szarnyasg:


Mause:
Shouldn't this already be handled by the excerpt?

szarnyasg:
To some extent, but I'd like to make a clear separation and put the TL;DR to the front matter.

szarnyasg:
The rationale is that we want nicer blog post pages with a TL;DR box at their beginning – see this sketch by Jonathan:

Screenshot 2024-02-06 at 16 48 00

Mause:
We should probably use the name `excerpt` instead then, so it's used in the seo description

szarnyasg:
Good point. Fix incoming ...

Issue #2045, opened 2024-02-06
Typos / Incompleteness in GROUPING_ID documentation


rpbouman:
The `GROUPING_ID()` documentation has a few typos in the second example (column is aliased to GROUPING_ID(y, q, d) but that should be GROUPING_ID(y, q, m). Also, the documentation mentions GROUPING_SETS and ROLLUP but not CUBE.

https://duckdb.org/docs/sql/query_syntax/grouping_sets.html#identifying-grouping-sets-with-grouping_id

PR #2046, opened 2024-02-06
Fix https://github.com/duckdb/duckdb-web/issues/2045

rpbouman:


szarnyasg:
Thanks for the fixes!

PR #2051, opened 2024-02-07
fix wrong output quotes + binder error in json_extract examples

OrangieLou:
Hi,

I discovered an inconsistency in the json_extract examples and ran all the example queries. I corrected a bug (one query resulted in a binder error) and the given output of one query (wrong quotation marks). I added another example to emphasise the difference between `->` and `->>`.


szarnyasg:
That's a great contribution, thanks for fixing these!

PR #2058, opened 2024-02-07
Apply headline capitalization on front page

szarnyasg:
Experimental PR, might not merge


szarnyasg:
Doesn't look good, decided against it

PR #2060, opened 2024-02-07
use raw string for R example

eitsupi:
I don't think escaping double quotes or the `paste` function are necessary here.

### DIFF
--- a/_includes/landing-page/r/sql-query.r
+++ b/_includes/landing-page/r/sql-query.r
@@ -4,11 +4,13 @@ library(duckdb)
 con <- dbConnect(duckdb())
 duckdb_register(con, "iris", iris)
 
-query <- paste(
-    "SELECT count(*) AS num_observations, ",
-    "max(\"Sepal.Width\") AS max_width, ",
-    "max(\"Petal.Length\") AS max_petal_length ",
-    "FROM iris ",
-    "WHERE \"Sepal.Length\" > 5 ",
-    "GROUP BY ALL")
+query <- r"(
+SELECT count(*) AS num_observations,
+max("Sepal.Width") AS max_width,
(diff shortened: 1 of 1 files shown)


szarnyasg:
Good point, thanks!

szarnyasg:
Hey, sorry, I ended up reverting this in #2398 as the syntax highlighter rendered the raw string incorrectly, and the colors in resulting code block quite bad IMHO. Apologies!

eitsupi:
Can we at least try to avoid double-quote escaping by enclosing the string in single quotes? The current one looks very bad.......

szarnyasg:
Sure, see #2407. I think this looks okay.

Screenshot 2024-03-02 at 15 56 08

szarnyasg:
Well, the copy button hovering over the `t` character doesn't look great, so I'll do something about that.

eitsupi:
Thank you. I can't try this right away, but if we use single quotes, can we just make it a multi-line string without using the `paste()` function since I don't need to use raw string?

```r
'select
max(...
from...
...all'
```

szarnyasg:
@eitsupi unfortunately, this does not work:

```R
> query <- 'SELECT count(*) AS num_observations
+     max("Sepal.Width") AS max_width,
+     max("Petal.Length") AS max_petal_length
+     FROM iris
+     WHERE "Sepal.Length" > 5
+     GROUP BY ALL'
> dbGetQuery(con, query)
Error: Parser Error: syntax error at or near "max"
LINE 2:     max("Sepal.Width") AS max_width,
            ^
```

eitsupi:
I see, thanks.

It is a pity that that example gives people the impression that extra function, extra quots, and extra commas are necessary.

PR #2082, opened 2024-02-09
Add ibis example

cpcloud:
This PR adds an Ibis example alongside the other Python examples. No pressure to accept, just thought I would put up a PR in case folks are amenable to it!

### DIFF
--- a/_includes/landing-page/python/ibis-integration.py
+++ b/_includes/landing-page/python/ibis-integration.py
@@ -1,0 +1,11 @@
+# Querying data with ibis using DuckDB
+import ibis
+from ibis import _
+
+t = ibis.memtable({
+    'station': ['Delft', 'Delft', 'Gouda', 'Gouda'],
+    'day': ['Mon', 'Tue', 'Mon', 'Tue'],
+    'num_services' : [22, 20, 27, 25]})
+
+# execute a query on duckdb and return a dataframe
+data = t.group_by("station").agg(_.num_services.sum()).to_pandas()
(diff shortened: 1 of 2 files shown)


_includes/landing-page/python/ibis-integration.py
```diff
@@ -0,0 +1,11 @@
...
+import ibis
+from ibis import _
+
+t = ibis.memtable({
+    'station': ['Delft', 'Delft', 'Gouda', 'Gouda'],
+    'day': ['Mon', 'Tue', 'Mon', 'Tue'],
+    'num_services' : [22, 20, 27, 25]})
+
+# execute a query on duckdb and return a dataframe
+data = t.group_by("station").agg(_.num_services.sum()).to_pandas()
```
lostmygithubaccount, line 11: might we worth adding a UDF example too? perhaps just the `count_vowels`?

cpcloud: Happy to add, but would like to wait for feedback before adding more!

szarnyasg:
Hi @cpcloud, thanks for the PR! Indeed (and unfortunately), we cannot merge this – we would like to keep our landing site lightweight and we also decided against including other libraries such as `numpy`.

cpcloud:
Ah, interesting!

numpy is a pandas dependency, so it's already in there FWIW. I guess it's something else then?

szarnyasg:
It's just a matter of having limited real estate on the landing page. We can only expect a certain amount of attention from users clicking through various examples for the languages displayed there.

cpcloud:
For sure! Thanks for the feedback!

PR #2092, opened 2024-02-11
Clarify temp_directory is a dir and not a file

davig019:
Should the wording of "By default, DuckDB uses the `.tmp` directory to spill to disk" also be updated?

The other documentation leads me to believe that this is outdated info:

1. The default on the [configuration page](https://duckdb.org/docs/sql/configuration) is null 
2. The [tuning workloads](https://duckdb.org/docs/guides/performance/how_to_tune_workloads.html#prerequisites) page implies it's set automatically IFF the database is persistent (to where?, a new file in the same location as the DB file i would guess?)
3. The [pragma](https://duckdb.org/docs/sql/pragmas#temp-directory-for-spilling-data-to-disk) documentation implies it's always set to /tmp


szarnyasg:
Hi @davig019, thanks for the PR. Indeed I agree with the observations. I'll merge this and will make a few followup adjustments in wording to clarify when the `.tmp` directory is used to spill to disk and where it's created.

PR #2093, opened 2024-02-11
Document that config options functionality exists in JDBC

davig019:
The functionality exists in the [JNI code](https://github.com/duckdb/duckdb/blob/main/tools/jdbc/src/jni/duckdb_java.cpp#L407) but is not documented.

Wording copied from the [C documentation](https://duckdb.org/docs/api/c/config)


szarnyasg:
Thanks!

PR #2215, opened 2024-02-19
Update installation page

szarnyasg:
- intro text
- ODBC link
- Java/Maven snippets
- Python pip install instruction

Followup of #2202

### DIFF
--- a/_includes/installation.html
+++ b/_includes/installation.html
@@ -1,8 +1,8 @@
-<p>You have three options for getting up and running with DuckDB:</p>
+<p>This page contains installation options for DuckDB:</p>
 <ul>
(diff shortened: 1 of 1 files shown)


_includes/installation.html
```diff
@@ -86,7 +86,7 @@ <h3>Usage Example</h3>
 	<!-- End Empty Results -->
 
 	<div class="latest python">
-		{% highlight bash %}pip install duckdb=={{ site.currentduckdbversion }}{% endhighlight %}
+		{% highlight bash %}pip install duckdb{% endhighlight %}
```
Mause, line 89: If we're going to remove the explicit version, may be worth adding the -U flag

Mause (changes requested):

_includes/installation.html
szarnyasg: Sure, see #2216

PR #2228, opened 2024-02-20
Update sql-query.js

carlopi:
Fixing errors and making so this should be now be properly runnable.

Only question is whether the URL is too long.


szarnyasg:
Thanks, should be good!

PR #2256, opened 2024-02-21
Update star count weekly

### LINKED ISSUE #2244: Fetching star count occasionally fails
See here

Screenshot 2024-02-20 at 18 03 40

szarnyasg:
Will fix #2244


Mause:
It might be worth switching to the `gh api` tool, as this might just be related to token access (who triggers the build)

szarnyasg:
Good point. Let's see, I've adjusted the main CI job first via #2261

PR #2257, opened 2024-02-21
changed main column style of documentation

Franz-Kafka:
- changed headings to new design
- adapted margins
- changed total width
- added breadcrumbs
- changed version switch
- changed mobile burger design
- bug fixes

### DIFF
--- a/_includes/breadcrumbs.html
+++ b/_includes/breadcrumbs.html
@@ -1,0 +1,11 @@
+<div id="breadcrumbs">
+{% assign crumbs = page.url | remove:'/index.html' | split: '/' %}
+
+{% for crumb in crumbs offset: 1 %}
+  {% if forloop.last %}
+    <span class="slash">/</span> <span class="current">{{ page.title }}</span>
+  {% else %}
+    <span class="slash">/</span> <span>{{ crumb | replace:'-',' ' | remove:'.html' | capitalize }}</span>
+  {% endif %}
+{% endfor %}
+</div>
--- a/_includes/breadcrumbs.html
+++ b/_includes/breadcrumbs.html
@@ -1,11 +1,69 @@
-<div id="breadcrumbs">
-{% assign crumbs = page.url | remove:'/index.html' | split: '/' %}
+{% if page.url contains '/docs/archive/' %}
+  {% assign urlVersionParts = page.url | split: "/docs/archive/" %}
+  {% assign urlVersionParts = urlVersionParts[1] | split: "/" %}
+  {% assign versionviewed = urlVersionParts[0] %}
+  {% assign startingUrl = "docs/archive/" | append: versionviewed %}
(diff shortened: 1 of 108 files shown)


szarnyasg:
Hi @Franz-Kafka, thanks! I found one issue. When deployed locally with Jekyll, visit an overview page:
http://localhost:4000/docs/data/overview

szarnyasg:
The Why DuckDB page is broken: http://localhost:4000/why_duckdb
Screenshot 2024-02-21 at 15 24 03

szarnyasg:
Regarding the breadcrumb, it currently uses the URL's slug, e.g.:

http://localhost:4000/docs/sql/query_syntax/unnest
becomes
> / Docs / Sql / Query_syntax / Unnesting

Could you please change it so that it uses the `"page"` attributes of the `menu_docs_<version>.json` file?

Franz-Kafka:
Hi @szarnyasg 
Thanks for these hints. 
1) I moved the heading "Pages in This Section" to the layout file. This helps also for consistency. 
2) I fixed the Why DuckDB page as well as other pages with the same structure
3) I think this will be quite complicated with Liquid. Maybe better to do it with JavaScript then. And for now we can go with the slugs or I will just hide the breadcrumbs. Can work on it tomorrow evening…

szarnyasg:
Indeed #3 is tricky with Jekyll, essentially we'll have to replicate the `sidenav` to some extent:
https://github.com/duckdb/duckdb-web/blob/959c231631c1e1b013ed0be5961155ad78c6b5d9/_includes/menu_sidebar.html#L25

szarnyasg:
@Franz-Kafka I can merge this now and you can continue from `main`, or I can hold this back, let you do the changes and merge it later. What would you prefer?

Franz-Kafka:
Yes, you can merge it now and I try to get the breadcrumbs done with jekyll and our menu json in the next step if this ok for you. 
Maybe Pedro or somebody else can support? I updated my current version… but I get always the same result and don't now why at the moment… https://github.com/Franz-Kafka/duckdb-web/blob/main/_includes/breadcrumbs.html

szarnyasg:
@Franz-Kafka – I found the issue and it's happening due to some really unintuitive behaviour on Jekyll's end.

Take the following code:

```liquid
{% assign a = 'x/y' %}
{% assign b = 'qwe' %}

{% if (a | replace: '/', '') == b %}
    equal
{% else %}
    not equal
{% endif %}
```

This returns `equal` for some reason! The workaround I used is to assign the left side of the comparison to a new variable:

```liquid
{% assign a = 'x/y' %}
{% assign b = 'qwe' %}

{% assign c = a | replace: '/', '' %}

{% if c == b %}
    equal
{% else %}
    not equal
{% endif %}
```

This returns `not equal` as one would expect.

szarnyasg:
Further developments are in PR #2273 and issue #2274

PR #2264, opened 2024-02-21
from.md: Add example of a table alias

moshekaplan:
Add an example of a table alias, as there does not appear to be one currently included


szarnyasg:
Thanks!

PR #2276, opened 2024-02-23
Update patternmatching.md

t829702:
fix the wrong doc; or can someone correct this?

am still new learning from the https://duckdb.org/docs/sql/functions/patternmatching#options-for-regular-expression-functions doc , but am confused:
[this section](https://duckdb.org/docs/sql/functions/patternmatching#options-for-regular-expression-functions) just said `'c' for case-sensitive matching`, then next section says use `the 's' option (case-sensitive matching)`
```
Option	Description
'c'	case-sensitive matching
```

```
To achieve best performance, the 's' option (case-sensitive matching) should be passed if applicable...
```

then what is `'s'` option: non-newline sensitive matching for? then i tried this example from live code, it tells false:
image


szarnyasg:
Thanks, we're indeed off from the official RE2 flags here. We'll take a look at the code but I'll merge this in the meantime.

PR #2280, opened 2024-02-23
RE2 updates

szarnyasg:
Followup of #2276

### DIFF
--- a/docs/sql/functions/patternmatching.md
+++ b/docs/sql/functions/patternmatching.md
@@ -6,6 +6,8 @@ railroad: expressions/like.js
 
 There are four separate approaches to pattern matching provided by DuckDB: the traditional SQL `LIKE` operator, the more recent `SIMILAR TO` operator (added in SQL:1999), a `GLOB` operator, and POSIX-style regular expressions.
 
+## The RE2 Library
+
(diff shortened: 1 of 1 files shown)


t829702:
found from Postgres, and tested some supported in DuckDB some are not, and don't find same from duckdb.org, can request another clarification in documentation DuckDB?

https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS

4.1.2. Constants [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS)
4.1.2.1. String Constants [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS)
4.1.2.2. String Constants With C-Style Escapes [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE)
4.1.2.3. String Constants With Unicode Escapes [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-UESCAPE)
4.1.2.4. Dollar-Quoted String Constants [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING)
4.1.2.5. Bit-String Constants [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-BIT-STRINGS)

PR #2291, opened 2024-02-24
Feature/azure doc update

quentingodeau:
image

image

### DIFF
--- a/docs/extensions/azure.md
+++ b/docs/extensions/azure.md
@@ -59,7 +59,7 @@ Use the following [configuration options](../sql/configuration) how the extensio
 | `azure_context_caching` | Enable/disable the caching of the underlying Azure SDK HTTP connection in the DuckDB connection context when performing queries. If you suspect that this is causing some side effect, you can try to disable it by setting it to false (not recommended). | `BOOLEAN` | `true` |
 
 > Setting `azure_transport_option_type` explicitly to `curl` with have the following effect:
-> * On Linux, this may solve certificates issue (`Error: Invalid Error: Fail to get a new connection for: https://⟨storage account name⟩.blob.core.windows.net/. Problem with the SSL CA cert (path? access rights?)`) because when specifying the extension will try to find the bundle certificate in various paths (that is not done by *curl* by default and might be wrong due to static linking see issue).
+> * On Linux, this may solve certificates issue (`Error: Invalid Error: Fail to get a new connection for: https://⟨storage account name⟩.blob.core.windows.net/. Problem with the SSL CA cert (path? access rights?)`) because when specifying the extension will try to find the bundle certificate in various paths (that is not done by *curl* by default and might be wrong due to static linking).
 > * On Windows, this replaces the default adapter (*WinHTTP*) allowing you to use all *curl* capabilities (for example using a socks proxies).
 > * On all operating systems, it will honor the following environment variables:
 >   * `CURL_CA_INFO`: Path to a PEM encoded file containing the certificate authorities sent to libcurl. Note that this option is known to only work on Linux and might throw if set on other platforms.
--- a/docs/extensions/azure.md
+++ b/docs/extensions/azure.md
@@ -21,14 +21,14 @@ Once the [authentication](#authentication) is set up, the Azure Blob Storage can
 
 ```sql
 SELECT count(*)
-FROM 'az://⟨my_container⟩/⟨my_file⟩.⟨parquet_or_csv⟩';
+FROM 'az://<my_container>@<my_storage_account>.blob.core.windows.net/<my_file>.<parquet_or_csv>';
 ```
 
 Globs are also supported:
 
 ```sql
 SELECT *
-FROM 'az://⟨my_container⟩/*.csv';
+FROM 'az://<my_container>@<my_storage_account>.blob.core.windows.net/*.csv';
 ```
 
 ## Configuration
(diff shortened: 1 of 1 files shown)


quentingodeau:
Hi here go some azure doc update with the new PRs (let me mention @samansmink to let him review too)
Regards,
Quentin

szarnyasg:
Hi @quentingodeau thanks for the contribution! I made a few minor adjustments and assigned Sam to review the PR.

docs/extensions/azure.md
```diff
@@ -20,13 +20,13 @@ LOAD azure;
 Once the [authentication](#authentication) is set up, the Azure Blob Storage can be queried as follows:
 
 ```sql
-SELECT count(*) FROM 'az://<my_container>/<my_file>.<parquet_or_csv>';
+SELECT count(*) FROM 'az://<my_container>@<my_storage_account>.blob.core.windows.net/<my_file>.<parquet_or_csv>';
```
samansmink, line 23: I think we should make clear that the `@<my_storage_account>.blob.core.windows.net/` part is optional and can be omitted, now it seems like a user always needs to specify that.

samansmink (commented):
hey @quentingodeau thanks for the PR! This looks great. I have one small comment

docs/extensions/azure.md
quentingodeau: Hum my idea here was in fact to slowly make people switch to this syntax. And then remove  from the secret part the azure_account mention. My idea is to avoid that if you have two containers with the same name in two storage account then the applied scope become a bit more tricky...
In my use case I have often a dev and prod env and the containers are the same only the storage if different (I sometimes use duck to do quick comparaison)

quentingodeau: I was planning to discuss this point with you on the initial PR but we can take time to talk about it now (if that ok with you) ;)

samansmink: Okay, but i'm still not convinced we should be optimizing for the (in my mind) edge case where users have multiple storage accounts.

I would say the following is a lot more pleasant than needing to specify the super verbose azure endpoint every time you query the file, agree?
```
create secret az1 (
    type azure,
    provider credential_chain,
    account_name 'my_acc'
)
from 'az://container/file.parquet'
```    

I like the `@<my_storage_account>.blob.core.windows.net` syntax and also your solution to do scoping on the accounts, but i think we should support both?

happy to discuss though if you disagree still

quentingodeau: Hi @samansmink,

I understand your point and I guess it also to keep a coherence between AWS and Azure. 

From my perspective I’m kind of used to seeing long paths when using Spark, often we create temporary views or use Hive. So, when I use DuckDB I do:

```sql
CREATE VIEW my_table AS 
SELECT * 
FROM 'azure://...';
```

But I understand that it may not be the case for all! I don’t mind keeping the two ways in the code, my end purpose was to try to simplify things, in code and for the end users (avoid then to wonder if they need or not to set the `ACCOUNT_NAME` in variable or secret).

But if we want to keep the two ways, does it mean that the variable `azure_account_name` shouldn’t be deprecated no? And should be use to access public storage account?

In reality I don't mind so I will let you have the final word here (please confirm your choice and I will update the documentation accordingly :)

NB: At the same time and before make the final doc update let me point [this draft PR](https://github.com/duckdb/duckdb_azure/pull/42) and validate it (or not) the changes

samansmink: The new PR looks great, lets get it in before this one is wrapped up. I think supporting both is cleanest!

Fair point on `azure_account_name`, i guess it makes sense to be able to be able to set that globally

quentingodeau: I didn't move the `azure_account_name` from the deprecated section yet, because after re-reading the code I saw that this use case if also handle by the `CONFIG` provider I think we should only keep one way to configure public storage account access to avoid any miss comprehension/usage.

Because for now if you do:

```sql
SET azure_account_name = 'storageaccount1';

-- Assuming container and data.parquet exists
SELECT * FROM 'az://container/data.parquet'; -- => OK

CREATE SECRET s1 (
    TYPE AZURE,
    PROVIDER CONFIG,
    ACCOUNT_NAME 'storageaccount2'
);

-- Now the same request will fail because the secret have precedence over the variable
SELECT * FROM 'az://container/data.parquet';
```

I think that it might confuse users. 

But we can argue that the same goes if you define two secrets^^

Once again, I will late you the final call here:
1. Let the doc as it is for now and see later
2. Move back `azure_account_name` definition
3. Move back `azure_account_name` definition and make the `CONFIG` provider only handle `CONNECTION_STRING`, in this case shouldn't we rename the provider?
4. Let the doc as it is and later remove the `azure_account_name ` from the code (making the `CONFIG` provider the only way to configure storage account names

I think I preferred the 4 because you will be able to specify `SCOPE` if you containers have different names.

Regards,
Quentin

docs/extensions/azure.md
```diff
@@ -111,7 +137,38 @@ The possible values are the following:
 [`managed_identity`](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview);
 [`env`](https://github.com/Azure/azure-sdk-for-cpp/blob/azure-identity_1.6.0/sdk/identity/azure-identity/README.md#environment-variables);
 [`default`](https://github.com/Azure/azure-sdk-for-cpp/blob/azure-identity_1.6.0/sdk/identity/azure-identity/README.md#defaultazurecredential);
-`none`. The latest will result in an exception (invalid input).
+
+If not explicit `CHAIN` is provide the default one will be [`default`](https://github.com/Azure/azure-sdk-for-cpp/blob/azure-identity_1.6.0/sdk/identity/azure-identity/README.md#defaultazurecredential)
```
samansmink, line 141:
```suggestion
If no explicit `CHAIN` is provided, the default one will be [`default`](https://github.com/Azure/azure-sdk-for-cpp/blob/azure-identity_1.6.0/sdk/identity/azure-identity/README.md#defaultazurecredential)
```

samansmink (commented):
1 small nitpick, otherwise looks good!

szarnyasg:
Perfect, thanks both!

PR #2294, opened 2024-02-24
Remove outdated note

carlopi:
This was true only on v0.9.2 for the moment the post went out. I think going forward we don't need this.


szarnyasg:
Thanks!

PR #2295, opened 2024-02-24
Remove v0.4.0 and v0.5.1 from docs

szarnyasg:


szarnyasg:
Should also drop 0.9.0 and 0.9.1, and just redirect them to 0.9.2.

PR #2307, opened 2024-02-25
Update patternmatching.md; use E'...' escape strings

t829702:
from https://github.com/duckdb/duckdb-web/pull/2280#issuecomment-1961870459 /fyi @szarnyasg


szarnyasg:
Thanks!

Issue #2308, opened 2024-02-25
request escape strings documentation DuckDB?


t829702:
> found from Postgres, and tested some supported in DuckDB some are not, and don't find same from duckdb.org, can request another clarification in documentation DuckDB?

https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS

4.1.2. Constants [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS)
4.1.2.1. String Constants [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS)
4.1.2.2. String Constants With C-Style Escapes [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE)
4.1.2.3. String Constants With Unicode Escapes [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-UESCAPE)              <<<<<<<<<<= no support of DuckDB?
4.1.2.4. Dollar-Quoted String Constants [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING)
4.1.2.5. Bit-String Constants [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-BIT-STRINGS)        <<<<<<<<<<== not clear

_Originally posted by @t829702 in https://github.com/duckdb/duckdb-web/issues/2280#issuecomment-1961892029_

szarnyasg:
@t829702 thanks for these pointers, DuckDB indeed supports `E'...'` and `$$...$$` strings. We clarified these via two PRs.

PR #2309, opened 2024-02-25
Document escape strings

szarnyasg:
Towards #2308


t829702:
the `4.1.2.4. Dollar-Quoted String Constants` seems also working, do you want to include as well?

4.1.2.4. Dollar-Quoted String Constants [#](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING)

PR #2359, opened 2024-02-28
Fix typo in COI headers example

dude0001:
Changes are happening in DuckDB-WASM that will make the COI conversation and this blog post more visible to users. If a user tries to implement the COI headers referencing the headers in the example here it would fail to enable COI and `SharedArrayBuffer` needed for DuckDB-WASM to leverage POSIX threads in the browser. The typo is subtle for a user that is new to these concepts, and they might fail to find the issue even when referring to the web.dev documentation linked.

I'm not sure what the process is for updating a block post that has been published for a while now, but it seems prudent to update this for the users who find and try to use this as a reference in the future.

Besides the web.dev article, I also verified the change after reviewing https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy.


szarnyasg:
@dude0001 Thanks for pointing this out! Indeed this is a sneaky typo.

carlopi:
Thanks also from here!

PR #2362, opened 2024-02-29
Move index generation into jekyll hook

Mause:
This removes the search index generation from the update-docs job, and ensures it is always up to date with the latest website content


szarnyasg (approved):
Great, thanks!

PR #2418, opened 2024-03-03
Fix formatting in postgres.md

cdeil:
Fix MD formatting here: https://duckdb.org/docs/extensions/postgres.html#create-schema--drop-schema

Screenshot 2024-03-03 at 14 07 35


szarnyasg:
Thanks!

PR #2427, opened 2024-03-04
new 404 page and small style fixes

Franz-Kafka:
- added new styled 404 page
- fixed some css


404.html
```diff
@@ -1,25 +1,32 @@
...
-    letter-spacing: -1px;
-  }
-</style>
-
-<div class="container">
-  <h1>404</h1>
-
-  <p><strong>Page not found :(</strong></p>
-  <p>The requested page could not be found.</p>
+<div class="pagehead bg{{ '1,2,3,4,5,6' | split: ',' | sample }}">
```
Mause, line 8: Does this randomly choose a new background at build time? Why would we want this?

Mause (changes requested):

404.html
Franz-Kafka: Jonathan would like to have different backgrounds so we thought of random backgrounds. But we could also define each page a specific background. 
I can update.

Mause: I'm just saying, they're not really random - it would only ever update when we rebuild the website?

szarnyasg:
Thanks! We can tackle the randomness question later.

PR #2428, opened 2024-03-04
Landing page: Rework R string

szarnyasg:
Following up on #2060


szarnyasg:
@eitsupi I found out that this also works

eitsupi:
Wow, that looks great, thanks!

PR #2430, opened 2024-03-04
Fix guides link

Mause:


404.html
```diff
@@ -25,7 +25,7 @@ <h1>Quick access</h1>
       </a>
       <a class="development" href="{{ site.baseurl }}/foundation">
         <div class="icon"></div>
-        <p>Development</p>
+        <p>Foundation</p>
```
Mause, line 28: Not sure what the intention was here, but we don't have a root Development page, so correcting the text

szarnyasg: The intention was to link the "Development" section, as I believed that's trivial to do. Alas, there's no root page, so let's link the build guide then.

PR #2433, opened 2024-03-04
removed google iframe for release calendar

Franz-Kafka:
- removed google calendar ifram
- added python script to fetch .ics and format json file of upcoming releases, stored in _data


szarnyasg:
Thanks, will merge this, and will add a new PR that invokes the script from CI.

PR #2456, opened 2024-03-06
Add CORS tips in s3api

douenergy:
Sometimes, even after correctly setting up all the credentials for an S3 bucket, a browser may not send requests to the S3 bucket. This issue often arises if you haven't set the correct CORS (Cross-Origin Resource Sharing) headers.

Open the [DuckDB WASM Shell](https://shell.duckdb.org/)

### With CORS Header Setting in the Bucket
```sql
FROM 'https://ducket-with-cors-header.s3.ap-northeast-1.amazonaws.com/d1.parquet';
```
### Without CORS Header Setting in the Bucket
```sql
FROM 'https://ducket-without-cors-header.s3.ap-northeast-1.amazonaws.com/d1.parquet';
```

You can download the parquet file directly from both URLs. However, only the bucket with the proper CORS header settings can be queried in the DuckDB WASM.

**We should add tips for setting the CORS header in an S3 bucket.**

截圖 2024-03-07 凌晨12 24 12

### DIFF
--- a/docs/api/wasm/data_ingestion.md
+++ b/docs/api/wasm/data_ingestion.md
@@ -150,6 +150,27 @@ await c.query(`
 `);
 ```
 
+> Tip If you encounter a Network Error `("Failed to execute 'send' on 'XMLHttpRequest')"` when you try to query files from S3, configure the S3 permission CORS header. For example:
+```
+[
+    {
+        "AllowedHeaders": [
+            "*"
+        ],
+        "AllowedMethods": [
+            "GET",
+            "HEAD"
+        ],
+        "AllowedOrigins": [
+            "*"
+        ],
+        "ExposeHeaders": [],
+        "MaxAgeSeconds": 3000
+    }
+]
+```
+
+
 ### Insert Statement
 
 ```ts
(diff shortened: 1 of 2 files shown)


douenergy:
cc @szarnyasg @carlopi  Please Take a Look !

douenergy:
I think the JS client may also encounter this issue (anything involving a browser may need proper CORS settings).

Mause:
> I think the JS client may also encounter this issue (anything involving a browser may need proper CORS settings).

The NodeJS client does not support running in a browser

douenergy:
my bad 😅. Is there any other document where we can add this?

carlopi:
Maybe you can find a way to this page: https://duckdb.org/docs/api/wasm/data_ingestion ?

It's not strictly connected to data ingestion (can happen also with extension loading) BUT it's by far the most common case of wanting to serve some files from S3 for them to be reachable.

Thanks a lot

douenergy:
Thanks @carlopi @Mause  for the reviewing.
I move the tip to [data_ingestion](https://duckdb.org/docs/api/wasm/data_ingestion) httpfs section 

截圖 2024-03-07 下午6 57 34

PR #2458, opened 2024-03-06
string -> character

### LINKED ISSUE #2457: CSV Import - `delim` says string but only accepts a single character
> Specifies the string that separates columns within each row (line) of the file.

Probably should say character, small change but helps avoid confusion

Tishj:
This PR fixes #2457 
`delim` only takes a single character


szarnyasg:
Thanks'

PR #2480, opened 2024-03-11
Correct inferred database name in `ATTACH` query

Flogex:
Small fix in the `ATTACH` docs: The inferred database name for 'file.db' is "file", not "file_db".


szarnyasg:
Thanks!

PR #2496, opened 2024-03-12
updated left sidenavigation and toc-menu

Franz-Kafka:
- new look for the left sidebar
- updated the look of the toc menu
- added another level to the toc menu


szarnyasg:
Fantastic, I tested it locally and it looks great. thanks!

PR #2499, opened 2024-03-12
case.md: Clarify switch statement example

moshekaplan:


szarnyasg:
Thanks for the clarification!

PR #2537, opened 2024-03-19
Update banner_docversion.html

Franz-Kafka:
- updated again to deliver correct url for latest version. should be good now.


szarnyasg:
Thanks!

PR #2551, opened 2024-03-20
go.md: link to repository

carlopi:
I navigated to go.md, and was expecting in some form to find the link to the repository. Unsure if that's the correct spot, or even if the phrase should be turned around, unsure.


szarnyasg:
Thanks, I formatted it to make the linter happy. There is also a link in the footer, though this may be more difficult to spot.
Screenshot 2024-03-20 at 20 47 30

PR #2561, opened 2024-03-21
Update timestamp.md

manticore-projects:
Fix the example for `epoch_us(timestamp '2021-08-03 11:59:44.123456')`


szarnyasg:
Thanks!

PR #2570, opened 2024-03-22
Update constraints.md - removed duplicate of example

bjornasm:
Seems like I had duplicated an example in my previous commit:

```sql
CREATE TABLE students (id INTEGER, name VARCHAR, PRIMARY KEY (id, name));
INSERT INTO students VALUES (1, 'Student 1');
INSERT INTO students VALUES (1, 'Student 2');
INSERT INTO students VALUES (1, 'Student 1');
-- Constraint Error: Duplicate key "id: 1, name: Student 1" violates primary key constraint
```

Removed the duplicate.


szarnyasg:
Thanks!

PR #2576, opened 2024-03-25
Fix Bash comment in JSON Data Import page

tlgs:
Just a stray SQL comment that fled to a nearby Bash block.

I took a look at every other Bash annotation in the code base and it seems that this is the only place where the `--` escaped to. For reference:

```bash
rg --multiline ' *```bash(?:.*\n)*? *```' .
```


szarnyasg:
Thanks!

PR #2594, opened 2024-03-26
Fix typo in 42.parquet post

domoritz:
Should probably say "can possibly" or just remove the word since "can" already implies that it's not guaranteed.


szarnyasg:
Good point and thanks for the fix!

PR #2603, opened 2024-03-27
Doc DuckDB have deprecated the usage of types in the sqllogictest

douenergy:
document  https://github.com/duckdb/duckdb/pull/11353

### DIFF
--- a/docs/dev/sqllogictest/result_verification.md
+++ b/docs/dev/sqllogictest/result_verification.md
@@ -15,8 +15,7 @@ SELECT 42, 84 UNION ALL SELECT 10, 20;
 10	20
 ```
 
-For legacy reasons the letters `R` and `T` are also accepted to denote columns. We have deprecated the usage of types in the SQLLogicTest. The DuckDB test runner does not use or need them internally.
-This means we only use I to denote columns.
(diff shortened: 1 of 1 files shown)


docs/dev/sqllogictest/result_verification.md
```diff
@@ -15,8 +15,7 @@ SELECT 42, 84 UNION ALL SELECT 10, 20;
 10	20
 ```
 
-For legacy reasons the letters `R` and `T` are also accepted to denote columns.
-
+For legacy reasons the letters `R` and `T` are also accepted to denote columns. We have deprecated the usage of types in the SQLLogicTest. The DuckDB test runner does not use or need them internally.This means we only use I to denote columns.
```
szarnyasg:
```suggestion
For legacy reasons the letters `R` and `T` are also accepted to denote columns.
> Deprecated DuckDB deprecated the usage of types in the SQLLogicTest. The DuckDB test runner does not use or need them internally – therefore, only `I` should be used to denote columns.

```

szarnyasg:
Thanks, I added a "Deprecated" box to highlight this.

douenergy:
Thanks @szarnyasg , looks nice.

PR #2614, opened 2024-03-29
Update window_functions.md

soerenwolfers:
Follow up on  https://github.com/duckdb/duckdb-web/pull/2612
- Mention the `list` special case (the only aggregate window function that respects nulls)
- Mention that the `FILTER` clause is available on aggregate window functions
- Mention that the `first` and `last` aggregate functions are shadowed, so `FILTER` is not available on them
- Combine the null treatment discussion of aggregate and general-purpose window functions in a joint section


szarnyasg:
Thanks!

PR #2617, opened 2024-03-29
Update window_functions.md

soerenwolfers:
I don't think the argument types add anything to justify the noise in the presentation:
Screenshot from 2024-03-29 10-52-39

(and they're not complete either since they can't specify the `typeof(expr)=typeof(default)` constraint).

Just an opinion though, free to ignore.


szarnyasg:
Thanks, that's a fair point. I'll also do the relayouting of this function tables – I simply forgot to include these functions in #2544.

PR #2618, opened 2024-03-29
Update aggregates.md

soerenwolfers:
```sql
SELECT 
mode() WITHIN GROUP (ORDER BY x),
mode(x),
mode(x ORDER BY x),
FROM 
(VALUES (2), (1)) _(x)
```
```
┌───────────────────────────────────┬─────────┬────────────────────┐
│ mode() WITHIN GROUP ( ORDER BY x) │ mode(x) │ mode(x ORDER BY x) │
│               int32               │  int32  │       int32        │
├───────────────────────────────────┼─────────┼────────────────────┤
│                                 1 │       2 │                  1 │
└───────────────────────────────────┴─────────┴────────────────────┘
```
shows that the `WITHIN GROUP` aggregate is converted to something more involved than what was previously claimed.

EDIT: Related ticket at  https://github.com/duckdb/duckdb/issues/11419, but I still think this PR makes the docs more correct than they currently are even without that ticket being resolved yet.

### DIFF
--- a/docs/sql/aggregates.md
+++ b/docs/sql/aggregates.md
@@ -142,11 +142,11 @@ as the first argument.
 
 | Function | Equivalent |
 |:---|:---|
-| `mode() WITHIN GROUP (ORDER BY column [(ASC|DESC)])` | `mode(column ORDER BY column [(ASC|DESC)])` |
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks! That's a clever use of `<code`> + `&#124;` to make it format correctly in editors.

PR #2622, opened 2024-03-29
Update char.md corrected typo

BrokenStreetlight:
I believe this is a typo in the description. It said uppercase characters are "smaller" than UPPERCASE characters. I believe its supposed to be uppercase character are "smaller" than LOWERCASE characters. 

p.s. I love DuckDB, thank you all so much for your work.


szarnyasg:
@BrokenStreetlight Thanks for the fix and also the kind words! We're glad you find DuckDB useful.

PR #2630, opened 2024-03-29
Update numeric.md

soerenwolfers:
-Replace claim that floating point is inexact by more specific examples and comparisons
-Add references


szarnyasg:
Thanks!

PR #2632, opened 2024-03-31
Update char.md

manticore-projects:
Explain `Unicode()` for empty strings returning `-1` which is different from other RDBMS (returning `0` in this case.)


szarnyasg:
Thanks!

PR #2638, opened 2024-04-01
with.md: Add Fibonacci example

moshekaplan:
Add Fibonacci example for developers unfamiliar with CTEs, so they could see CTEs being used to solve a problem they're already familiar with.


moshekaplan:
Note that this was roughly inspired by the code from https://www.codeproject.com/Tips/811924/Fibonacci-sequence-using-SQL-Server-CTE , but needed to be rewritten to be compatible with duckdb

szarnyasg:
@moshekaplan Thanks, this looks great! I added the output to the PR and it's good to go.

moshekaplan:
> @moshekaplan Thanks, this looks great! I added the output to the PR and it's good to go into `main`.

Good catch, sorry for that omission.

PR #2654, opened 2024-04-03
Remove invalid comma before IGNORE NULLS in window functions

jhrcek:
The syntax hints (or whatever you call these things like `first_value(expr[, IGNORE NULLS])`) on [this page](https://duckdb.org/docs/sql/window_functions.html#general-purpose-window-functions) suggest that there should be comma before IGNORE NULLS when calling window functions.
But this comma is not accepted by duckdb's parser.

This is either bug in duckdb parser (which should allow comma there but doesn't)
or a bug in documentation (which suggests adding a comma which shouldn't be there) :smile: 

Steps to see this in duckdb cli repl:
```bash
D create table t(a int);
D insert into t values(1),(2),(NULL);
D select last(a, IGNORE NULLS) over (order by a) from t;
Parser Error: syntax error at or near "NULLS"
LINE 1: select last(a, IGNORE NULLS) over (order by a) from t;
                              ^
D select last(a IGNORE NULLS) over (order by a) from t;
┌────────────────────────────────────────┐
│ last(a IGNORE NULLS) OVER (ORDER BY a) │
│                 int32                  │
├────────────────────────────────────────┤
│                                      1 │
│                                      2 │
│                                      2 │
└────────────────────────────────────────┘
D 
```


jhrcek:
Also a related question:

In the section https://duckdb.org/docs/sql/window_functions.html#general-purpose-window-functions
The description of nth_value says (emphasis mine):

`nth_value(expr, nth[, IGNORE NULLS])`

Returns expr evaluated at the row that is offset rows after the current row (among rows with a non-null value of expr if IGNORE NULLS is set) within the window frame; **if there is no such row, instead return default (which must be of the Same type as expr)**. Both offset and default are evaluated with respect to the current row. If omitted, offset defaults to 1 and default to NULL.

But notice that the `nth_value(expr, nth[, IGNORE NULLS])` doesn't mention `default` parameter.

it's not clear to me whether nth_value function supports passing default value or not.
Could you please clarify this (and I can update the docs accordingly).

szarnyasg:
Hi @jhrcek thanks a lot for the PR! `nth_value` does not support default arguments, this was just a copy-paste error that occurred I accidentally introduced when reworking the window function table. I fixed this and bundled up a few more small fixes with this PR.

PR #2712, opened 2024-04-12
Include OR REPLACE and TEMP in createmacro figure

fbertsch:
Found this a bit confusing when trying to create some UDFs.


szarnyasg:
Indeed these were missing. Thanks!

PR #2720, opened 2024-04-16
Update timestamp.md

manticore-projects:
Add `CURRENT_TIMESTAMP` to the list

### DIFF
--- a/docs/sql/functions/timestamp.md
+++ b/docs/sql/functions/timestamp.md
@@ -26,6 +26,7 @@ The table below shows the available scalar functions for `TIMESTAMP` values.
 | [`age(timestamp, timestamp)`](#agetimestamp-timestamp) | Subtract arguments, resulting in the time difference between the two timestamps. |
 | [`age(timestamp)`](#agetimestamp) | Subtract from current_date. |
 | [`century(timestamp)`](#centurytimestamp) | Extracts the century of a timestamp. |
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks!

hayssams:
Looks like calling current_timestamp() does not work. 
```
$ duckdb
v0.10.1 4a89d97db8
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D select current_timestamp();
Catalog Error: Scalar Function with name current_timestamp does not exist!
Did you mean "get_current_timestamp"?
LINE 1: select current_timestamp()
               ^
D select current_date() ;
┌────────────────┐
│ current_date() │
│      date      │
├────────────────┤
│ 2024-04-16     │
└────────────────┘
D 
```

Mause:
It's not a function, it's a magic value (which should probably be explicitly called out in the docs)

manticore-projects:
Thank you for quick roundturn.

```sql
-- works
select current_date=current_date();

-- fails and I think the function should be added just for consistency
select current_timestamp=current_timestamp();
```

manticore-projects:
> It's not a function, it's a magic value (for lack of a better term, which should probably be explicitly called out in the docs)

@Mause: in my limited understanding, it should be foremost a `Function()` because it returns a system value, which then can be abbreviated by a "magic value" (which I would call a Time Key Expression).

PR #2785, opened 2024-04-26
Add another example to show table functions really work as tables

endafarrell:
In hindsight it was obvious, but initially it was unclear how to get a subset of rows and columns. This should make it clearer.


szarnyasg:
Thanks!

PR #2796, opened 2024-04-29
Correct the alias for && to reference `list_has_any`

ptliu-awake:
`&&` aliases to `list_has_any` per https://github.com/duckdb/duckdb/pull/7518


szarnyasg:
Thanks!

PR #2830, opened 2024-05-06
Update setops.md

soerenwolfers:


szarnyasg:
I checked this with the devs and this is indeed expected behavior for `UNION`.

PR #2862, opened 2024-05-13
add $ before bash code blocks

Franz-Kafka:
- will add $ before all bash code blocks via css. 
- remove copy button on error code blocks.

@szarnyasg important: you need to remove the manually added "$" from the codes.


szarnyasg:
Thanks. #2866 adjusts the bash code blocks.

PR #2867, opened 2024-05-14
Fix grammar errors as reported by lint.sh and vale

corwinjoy:
When running scripts/lint.sh the vale grammar checker complains that "e.g." and "i.e." should be followed by a comma. This is probably a good idea. I have gone ahead and applied these changes to fix the reported grammar errors.


szarnyasg:
Thanks! I'm all in for improving the consistency of the docs so this definitely helps. Some of the files you touched (e.g., the API reference) are auto-generated so we'll need to change that code in mainline DuckDB to make the changes in the documentation stick.

PR #2873, opened 2024-05-16
Fix cpp example to conform to current version QueryResult class.

corwinjoy:
I tried running the C++ example in the docs, and it gives errors using the current version of the QueryResult class. I have updated the basic example to use the current version of the class and added a simple print for the output.


Mause:
Thanks!

PR #2891, opened 2024-05-23
Suggest using all threads for compilation in R

david-cortes:
The docs suggest adding a flag in the R `Makevars` file to parallelize compilation of the R package of DuckDB.

The line it suggest to add fixes it at 8 threads, but one would usually want to use as many threads as possible. This PR adds such a suggestion for linux systems using `nproc` instead.


szarnyasg:
Thanks!

Issue #2879, opened 2024-05-19
Issue found on page 'Text Types': section "Formatting Strings" 


rpbouman:
The information in the section "[Formatting Strings](https://duckdb.org/docs/sql/data_types/text#formatting-strings)" is really about string literals, which is described in detail in https://duckdb.org/docs/sql/data_types/literal_types#string-literals

I suggest to remove the section and instead, include a single line in the first general section that says something like 'Literal text type values (strings) can be created using string literals." and then have string literals be a crossreference to https://duckdb.org/docs/sql/data_types/literal_types#string-literals

Page URL: <https://duckdb.org/docs/sql/data_types/text.html>

Issue #2878, opened 2024-05-19
Issue found on page 'Keywords and Identifiers': remove sections on string literals and numeric literals


rpbouman:
This page is about keywords and identifiers, but also contains two sections about literals, https://duckdb.org/docs/sql/keywords_and_identifiers#string-literals and https://duckdb.org/docs/sql/keywords_and_identifiers#numeric-literals.

I think these sections should not be here, whatever information they have should probably be moved to https://duckdb.org/docs/sql/data_types/literal_types

Page URL: <https://duckdb.org/docs/sql/keywords_and_identifiers.html>

Issue #2877, opened 2024-05-19
Issue found on page 'Text Types': Double quotes section should not be here.


rpbouman:
The https://duckdb.org/docs/sql/data_types/text page has a section about Double Quotes. 
This section is out of place as double quotes have nothing to do with the text type (or values of that type).

I get why it's there - as so many prog languages do use double quotes for delimiting strings/text-type values it's good to say something about that. But by giving it a separate section it suggests that double quoted "strings" are still, somehow text-type values.
The sections text also doesn't explicitly say: it merely states what double quotes do do ('Double quote characters (") are used to denote table and column names. '), not what they don't do, which is relevant to this page ("Double quotes are not string delimiters"). 

My suggestion would be to remove this section. Instead, the section on text type values (which is currently called 'Formatting Strings') should have a line or para that first explains that double quotes are not string delimiters, but identifier delimiters, and this could have a crossreference to the documentation of identifiers, https://duckdb.org/docs/sql/keywords_and_identifiers#identifiers

Page URL: <https://duckdb.org/docs/sql/data_types/text#double-quote-characters>

PR #2913, opened 2024-05-27
Fixes 2877, 2878, 2879

### LINKED ISSUE #2869: Identifier case sensitivity / SQL standard
In https://duckdb.org/docs/sql/keywords_and_identifiers, it reads:

"Following the convention of the SQL standard, identifiers in DuckDB are case-insensitive."

The remark about the standard is not entirely correct. AFIK, the SQL standard treats "delimited identifiers" (quoted identifiers) as case sensitive. (I can dig up the relevant lines from the spec if required).

This is not just a thing in the standard, major products like SQL Server, Postgres, Snowflake, SAP HANA, etc also treat delimited identifiers as case-sensitive. (see for example: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS, "Quoting an identifier also makes it case-sensitive, whereas unquoted names are always folded to lower case.")

rpbouman:
Fix https://github.com/duckdb/duckdb-web/issues/2877
Fix https://github.com/duckdb/duckdb-web/issues/2878 
Fix https://github.com/duckdb/duckdb-web/issues/2879
Fix #2869

### DIFF
--- a/docs/sql/data_types/text.md
+++ b/docs/sql/data_types/text.md
@@ -28,14 +28,14 @@ CREATE TABLE strings (
 
 The `VARCHAR` field allows storage of Unicode characters. Internally, the data is encoded as UTF-8.
 
-## Text type values
+## Text Type Values
 
 Values of the text type are character strings, also known as string values or simply strings. At runtime, string values are constructed in one of the following ways:
 
 * referencing columns whose declared or implied type is the text data type
 * [string literals](literal_types#string-literals)
 * [casting](../expressions/cast#explicit-casting) expressions to a text type
-* applying a [string operator](../functions/char#text-functions-and-operators), or invoking a function that returns a text type value 
+* applying a [string operator](../functions/char#text-functions-and-operators), or invoking a function that returns a text type value
 
 ## Functions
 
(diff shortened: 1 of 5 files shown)


docs/sql/data_types/text.md
```diff
@@ -28,56 +28,14 @@ CREATE TABLE strings (
 
 The `VARCHAR` field allows storage of Unicode characters. Internally, the data is encoded as UTF-8.
 
-## Formatting Strings
+## Text type values
 
-Strings in DuckDB are surrounded by single quote (apostrophe) characters (`'`):
+Values of the text type are character strings, also known as string values or simply strings. At runtime, string values are constructed in one of the following ways:
 
-```sql
```
szarnyasg, line 35: @rpbouman why was this section and the "Strings with Special Characters" section below (explaining `chr`)  deleted?

szarnyasg:
Thanks! I pushed a few minor fixes and added a question.

szarnyasg:
This also fixes #2869, I've edited this to the root post.

PR #2919, opened 2024-05-29
Fix https://github.com/duckdb/duckdb-web/issues/2881

### LINKED ISSUE #2881: Issue found on page 'Subqueries': subquery comparisons using ANY / ALL / SOME appears not to be documented?
DuckDB supports queries like: 

SELECT .. FROM .. WHERE column > ALL|ANY|SOME (...subquery...) 

But it appears that this is not documented?

Page URL: <https://duckdb.org/docs/sql/expressions/subqueries.html>

rpbouman:
Fixes #2881


szarnyasg:
Thanks!

PR #2931, opened 2024-05-30
Fix java snapshot version

Mause:
Fixes https://github.com/duckdb/duckdb-java/issues/17


Mause:
See https://oss.sonatype.org/service/local/repositories/snapshots/content/org/duckdb/duckdb_jdbc/ for available snapshot versions as well

szarnyasg:
Thanks!

PR #2949, opened 2024-06-01
Fix formula for regr_slope

### LINKED ISSUE #2940: [duckdb/#12299] - Aggregate Regression function SLOPE and INTERCEPT seem to produce wrong results needs documentation
See https://github.com/duckdb/duckdb/issues/12299

szarnyasg:
Fixes #2940

### DIFF
--- a/docs/sql/aggregates.md
+++ b/docs/sql/aggregates.md
@@ -517,7 +517,7 @@ The table below shows the available statistical aggregate functions.
 <div class="nostroke_table"></div>
 
 | **Description** | Returns the slope of the linear regression line for non-null pairs in a group. |
-| **Formula** | `covar_pop(x, y) / var_pop(x)` |
+| **Formula** | `covar_pop(x, y) / covar_pop(x, x)` |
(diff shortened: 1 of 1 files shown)


szarnyasg:
The code comments say that the denominator is `var_pop(x)` but apparently it's `covar_pop(x, x)`.

```cpp
// REGR_SLOPE(y, x)
// Returns the slope of the linear regression line for non-null pairs in a group.
// It is computed for non-null pairs using the following formula:
// COVAR_POP(x,y) / VAR_POP(x)
```

pdet:
From the code, it seems to be using `VarPop`.

It's been a few years and I implemented many of these functions in two days, and I ended up using snowflake/presto/clickhouse as bases, should probably have documented these in the code :-).

I think this one was from snowflake:
https://docs.snowflake.com/en/sql-reference/functions/regr_slope

szarnyasg:
Thanks @soerenwolfers!

soerenwolfers:
https://github.com/duckdb/duckdb-web/pull/3048

PR #2966, opened 2024-06-04
which_secret should include a `FROM` as it's a table function

mehd-io:
Fixing a small error in the example for `which_secret`


szarnyasg:
Thanks!

PR #2970, opened 2024-06-04
rm a limitation that has been lifted since

julien-c:
IIUC this limitation has been lifted since then?


szarnyasg:
Indeed it is now auto-loading, thanks!

PR #2990, opened 2024-06-05
Fix small copy-paste error (list_position example)

OrangieLou:
Hi,

the example of `list_position` contained a copy paste error. In the example it said `list_contains` instead of `list_position`.


szarnyasg:
Thanks!

PR #2992, opened 2024-06-05
Restructure and add new JSON functions

PhictionalOne:
This PR Consists of two parts.
1. Restructure doc to follow a more "natural" reading experience such that the sections on 'Creation', 'Extraction', 'Scalar Functions' follow that scheme and not the other way round.
2. Introduce documentation for JSON Manipulation functions. Depends on whether [PR 12386](https://github.com/duckdb/duckdb/pull/12386) on the main DuckDB repo is accepted.

PR #2994, opened 2024-06-05
Restructure json documentation

PhictionalOne:
Restructure doc to follow a more "natural" reading experience such that the sections on 'Creation', 'Extraction', 'Scalar Functions' follow that scheme and not the other way round.


szarnyasg:
Thanks. This supersedes #2992

PR #3028, opened 2024-06-10
Delta extension blogpost

samansmink:


szarnyasg:
@samansmink please rename the file to `_posts/2024-06-10-delta.md` (I don't want to push it because it could conflict).

samansmink:
ah good catch, thx will do

PR #3045, opened 2024-06-11
Upgrade Jekyll from v3 to v4

szarnyasg:
Performance for full deployment is a bit better:
* Jekyll v3: 136.109 seconds
* Jekyll v4: 106.286 seconds

This needs CI fixes and thorough testing before merging.


szarnyasg:
Tested it locally and seemed fine. Let's see how it fares on CI.

PR #3046, opened 2024-06-11
Use Jekyll's link tag

szarnyasg:
To run the script, use:

```bash
ag --md --ignore-dir=archive/ -l '\]\([.a-zA-Z0-9]' docs/ | xargs -I {} python3 scripts/use-link-tags.py {}
```


szarnyasg:
This seems a bit too cumbersome as it requires absolute URLs:

> The path to the post, page, or collection is defined as the path relative to the root directory (where your config file is) to the file, not the path from your existing page to the other page.

https://jekyllrb.com/docs/liquid/tags/#link

szarnyasg:
So this would need to change a thousand plus links and it would also necessitate (smaller) changes to the archival script...

PR #3048, opened 2024-06-11
Update aggregates.md

soerenwolfers:
Fixes https://github.com/duckdb/duckdb-web/issues/2940
- fix formula for `regr_slope`
- `regr_sxx` in duckdb computes something different from snowflake. This should probably be fixed later, but for now I am fixing the documentation to state what is currently being computed. (duckdb currently computes variances instead of sums-of-squares; i.e., the sums-of-squares but divided by counts) 
- This doesn't affect things like `regr_slope = regr_sxy / regr_sxx` because the counts cancel out.
- Most formulas for the `regr...` functions were only correct in absence of nulls. I fixed some of these, but also removed some of the formulas for some of them entirely because they are pretty self-evident but difficult to write down in the table (e.g., regr_sxx does the same as `var_samp` but applied to the subset of non-NULL values, which is easier said in words than in formulas)
- Fixed some `count(*)` in some formulas by `count(x)`
- Put repeated "for non-null values" at the top of the table, since it affects every single statistical aggregate. (I did keep it in the `regr_` descriptions, because that's the whole point of the `regr_...` aggregates.) 
- Some function descriptions said "for all input values" which I deleted because it was superflous and misleading, since they too ignore NULLs.
- Remove superfluous "Returns" from start of descriptions.

### DIFF
--- a/docs/sql/aggregates.md
+++ b/docs/sql/aggregates.md
@@ -366,41 +366,42 @@ The table below shows the available approximate aggregate functions.
 
 ## Statistical Aggregates
 
-The table below shows the available statistical aggregate functions.
+The table below shows the available statistical aggregate functions. 
+They all ignore `NULL` values (in the case of a single input column `x`), or pairs where either input is `NULL` (in the case of two input columns `y` and `x`).
 
 | Function | Description |
 |:--|:--------|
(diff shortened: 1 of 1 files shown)


szarnyasg:
This is super thorough, thank you!

PR #3091, opened 2024-06-16
Interval from decimal string

soerenwolfers:
Had copied that from previous version, but it seems to have been outdated.


szarnyasg:
Thanks!

PR #3131, opened 2024-06-21
Clarifying  USING and ASOF Join

javier:
I had a hard time understanding the warning on USING and being explicit about the parameters. I reworded it and added an extra example


szarnyasg:
Hi @javier, I made some minor tweaks, and made the last example use `USING`. This returns the same result and, I think, is more in line with the intent of this patch. Feel free to ament it in a followup PR is that's not the case!

PR #3178, opened 2024-06-27
Update partitioned_writes.md

andriibratanin:
Include example of compressed hived data export with explicit file extensions set


szarnyasg:
Looks great, thanks!

PR #3247, opened 2024-07-06
is "e.g." necessary?

wibeasley:


szarnyasg:
Hi @wibeasley, yes, the e.g. is necessary. While joining on VARCHAR columns is probably the most common reason for performance hits, operations such as filtering (e.g. `x = '123'`) and aggregation with casting (`sum(x::BIGINT)`) also perform slower than their counterparts using numbers.

PR #3276, opened 2024-07-10
Fix extension links

szarnyasg:
cc @carlopi


carlopi:
Thanks, unsure why, now I have a script generating those so stuff should be aligned.

Add another PR with the current state at https://github.com/duckdb/duckdb-web/pull/3278

PR #3281, opened 2024-07-10
Configurable block size

taniabogatsch:
Fixes https://github.com/duckdb/duckdb-web/issues/3230.


szarnyasg:
Thanks! I'll merge this for the documentation's v1.1 version once it becomes available (likely _before_ DuckDB's version 1.1 is out).

taniabogatsch:
@szarnyasg, I merged the recent `main` into this PR and spotted a few irregularities in `overview.md`. Specifically, some options that are no longer there? Here is the list. Why were they removed?

```
| `allocator_background_threads` | Whether to enable the allocator background thread. | `BOOLEAN` | `false` |
| `arrow_output_list_view` | If export to arrow format should use ListView as the physical layout for LIST columns. | `BOOLEAN` | `false` |
| `produce_arrow_string_view` | If strings should be produced by DuckDB in Utf8View format instead of Utf8. | `BOOLEAN` | `false` |
```

Why were these removed?
```
| `index_scan_max_count`                       | The maximum index scan count sets a fixed threshold for index scans.                                                                              | `UBIGINT` | `2048`                                              |
 | `index_scan_percentage`                      | The index scan percentage sets a percental threshold for index scans.                                                                             | `DOUBLE`  | `0.001`                                             |
```

Intentional, this PR.
```
| `default_block_size` | The default block size for new duckdb database files (new as-in, they do not yet exist). | `UBIGINT` | `262144` |
```

taniabogatsch:
Same for the local options, why was this removed?
```
| `custom_profiling_settings` | Accepts a **JSON** enabling custom metrics | `VARCHAR` | `{"OPERATOR_TIMING": "true", "CPU_TIME": "true"}` |
```

szarnyasg:
Hi @taniabogatsch Yes, these were removed because they are not supported in v1.0.0 yet. They will be added back with the reversal of the auto-config generation step (https://github.com/duckdb/duckdb-web/pull/3251).

```sql
v1.0.0 1f98600c2c
D from duckdb_settings() where name = 'produce_arrow_string_view';
┌─────────┬─────────┬─────────────┬────────────┬─────────┐
│  name   │  value  │ description │ input_type │  scope  │
│ varchar │ varchar │   varchar   │  varchar   │ varchar │
├─────────┴─────────┴─────────────┴────────────┴─────────┤
│                         0 rows                         │
└────────────────────────────────────────────────────────┘
```

```sql
v1.0.1-dev5200 a6e32b1158
D from duckdb_settings() where name = 'produce_arrow_string_view';
┌────────────────────────┬─────────┬────────────────────────────────────────────────────────────────────────────┬────────────┬─────────┐
│          name          │  value  │                                description                                 │ input_type │  scope  │
│        varchar         │ varchar │                                  varchar                                   │  varchar   │ varchar │
├────────────────────────┼─────────┼────────────────────────────────────────────────────────────────────────────┼────────────┼─────────┤
│ produce_arrow_string…  │ false   │ If strings should be produced by DuckDB in Utf8View format instead of Utf8 │ BOOLEAN    │ GLOBAL  │
└────────────────────────┴─────────┴────────────────────────────────────────────────────────────────────────────┴────────────┴─────────┘
```

taniabogatsch:
Makes sense - thanks for the quick reply!

PR #3289, opened 2024-07-15
Replace signbit example to -1.0

Mytherin:


carlopi (commented):
I think this is autogenerated from `src/core_functions/scalar/math/functions.json` in duckdb/duckdb, should it not be changed also there?

carlopi:
Then those are independent, given the one you modified are only old docs that are not autogenerated (I think? unsure).

Mytherin:
This does not only change old docs. But yes it should also be modified upstream.

carlopi (approved):
Thanks!

PR #3292, opened 2024-07-16
Fixed typo in `-init` argument description

aksel:
`~./duckdbrc` → `~/.duckdbrc`.

The former is not a valid path (I believe):

```
$ touch ~./duckdbrc
touch: cannot touch '~./duckdbrc': No such file or directory
```


szarnyasg:
Thanks!

PR #3293, opened 2024-07-16
small rework of community extensions

samansmink:
Changed:
- Added FAQ
- Changed menu structures
- Added some build instructions

I've changed the Left hand side menu does no longer list all extensions, this will soon become unwieldy as we get more community extensions. Also this menu now no longer links to external pages which was kind of confusing.

The Logo now correctly links to the overview page of the community extension docs instead of the main DuckDB docs.

We can now add common questions we get to the FAQ, i've added a few to get us started.


carlopi (approved):
Thanks!

Clearly an improvement, thanks for writing those down / bringing in some structure.

Main comment is that I think it would make sense to have a few other extensions in the left bar. Having all of them was not ideal / sustainable, but my idea was having say 6 random ones + a link to the list of all extensions.

Idea is that if you are in one of those extension pages, I would expect to be able to navigate to other extensions, since I think this should serve two objective: documenting community-extension-related topics AND showcase extensions that one might want to look at.

Anyhow, stuff can be done later, merging as is.

PR #3295, opened 2024-07-16
Update julia.md

era127:
Remove references to appender Flush which is deprecated.

Add explanation of DuckDB.query() for queries that are executed as multiple statements.


szarnyasg:
Thanks!

PR #3297, opened 2024-07-17
More docs community

carlopi:
Minor fixes, and expanded on descriptor.yml fields and how documentation gets generated.

### DIFF
--- a/documentation.md
+++ b/documentation.md
@@ -5,7 +5,7 @@ title: Documentation
 
 ## Submit a Community Extension
 
-Extensions code are hosted as Github repositories.
+Extensions are hosted as Github repositories.
 
 Users are encouraged to check the relevant extension repository for giving feedback, look at the implementation, and report issues.
 
--- a/documentation.md
+++ b/documentation.md
@@ -48,8 +48,6 @@ For example, this is the page for [quack]({% link extensions/quack.md %}).
 
 Documentation pages are generated from the fields provided in the YAML descriptor file that is part of the [Community Extensions Repository](https://github.com/duckdb/community-extensions) and from the auto-detected changes that a given extensions introduces in DuckDB.
 
-#### Autodetected changes
-
 Process is roughly as follow:
 
 ```sql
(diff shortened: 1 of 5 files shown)


documentation.md
```diff
@@ -1,22 +1,68 @@
...
+```sql
+CREATE TABLE functions_pre AS SELECT ... FROM duckdb_functions();
+LOAD extension_name;
+CREATE TABLE functions_post AS SELECT ... FROM duckdb_functions();
+SELECT * FROM functions_pre EXCEPT (FROM functions_post) ORDER BY ...;
+```
+
+This works well for detecting new functions, functions overload, new settings, and new types.
+
+Detection of other changes to the system (e.g., whether an additional Parser or Optimizer callback has been register) are viable, but not yet implemented. Those are reccomended to be provided as part of the `docs.extended_description` field.
```
samansmink, line 64:
```suggestion
Detection of other changes to the system (e.g., whether an additional Parser or Optimizer callback has been register) are viable, but not yet implemented. Those are recommended to be provided as part of the `docs.extended_description` field.
```

documentation.md
```diff
@@ -1,22 +1,68 @@
...
+
+### Hosted Extension Documentation Page 
+
+Each Community Extension has a documentation page at: `https://community-extension.duckdb.org/extensions/$extension_name.html`.
+
+For example, this is the page for [quack]({% link extensions/quack.md %}).
+
+Documentation pages are generated from the fields provided in the YAML descriptor file that is part of the [Community Extensions Repository](https://github.com/duckdb/community-extensions) and from the auto-detected changes that a given extensions introduces in DuckDB.
+
+#### Autodetected changes
```
samansmink, line 51: I think this header is not necessary? If you want to keep it i would suggest renaming it to 'documentation generation' and start with a sentence explaining that the section is going to explain how the documentation is generated

samansmink (commented):
Looks good! added some nitpicks

documentation.md
carlopi: Thanks!

Issue #3327, opened 2024-08-03
Issue found on page 'CSV Import': "auto_type_candidates Details" section could benefit from a cross link to https://duckdb.org/docs/data/csv/auto_detection#type-detection


rpbouman:
The "auto_type_candidates Details" section in the CSV overview page would benefit from a cross link to the type detection section of the "CSV Auto detection"  page.
Page URL: <https://duckdb.org/docs/data/csv/overview#auto_type_candidates-details>

While we're at it, the https://duckdb.org/docs/data/csv/auto_detection#type-detection doesn't explicitly name the `auto_type_candidates` parameter, it would benefit from adding some text to explicitly say candidates may be specified by that parameter

I will send a PR shortly.

PR #3328, opened 2024-08-03
Fixes https://github.com/duckdb/duckdb-web/issues/3327

rpbouman:
crosslinks re. auto type detection between overview and auto detection pages.


szarnyasg:
Thanks!

Issue #3330, opened 2024-08-03
Issue found on page 'CSV Import': auto_type_candidate Details section should list all allowed data types and point out data types that aren't listed are not supported.


rpbouman:
The section on the auto_type_candidates option in the CSV overview page does not say anything about alllowed/supported data types.

I used the following script to test which data types may be appear in the list passed to auto_type_candidates:

```
set variable types = [
  -- 'BIT',
  -- 'BLOB',
  'BOOLEAN',
  'DATE',
  -- 'ENUM(''DUCK_DUCK_ENUM'')',
  -- 'INTERVAL',
  -- 'LIST', --Parser Error: LIST is not valid as a stand-alone type
  -- 'VARCHAR[]', --Binder Error: Auto Type Candidate of type LIST is not accepted as a valid input
  -- 'MAP',
  'TINYINT',
  'SMALLINT',
  'INTEGER',
  'BIGINT',
  -- 'HUGEINT',
  -- 'UTINYINT',
  -- 'USMALLINT',
  -- 'UINTEGER',
  -- 'UBIGINT',
  -- 'UHUGEINT',
  -- 'STRUCT',
  -- 'UNION',
  'DOUBLE',
  'FLOAT',
  'DECIMAL',
  'VARCHAR',
  --'VARINT',
  'TIME',
  -- 'TIMETZ',
  -- 'TIMESTAMP_NS',
  'TIMESTAMP',
  -- 'TIMESTAMP_MS',
  -- 'TIMESTAMP_S',
  -- 'TIMESTAMPTZ',
  -- 'UUID',
  'NULL'
]
;

with 
original_types as (
  select * 
  from (
    describe 
    from test_all_types()
  )
)
,
types_from_csv as (
  select * 
  from (
    describe 
    from read_csv(
      'C:\Users\Roland_Bouman\Downloads\all_types_test.csv'
    , auto_type_candidates=getvariable('types')
    )
  )
)
select    original_types.column_name
,         original_types.column_type
,         types_from_csv.column_type
,         original_types.column_type = types_from_csv.column_type is_same
from      original_types 
left join types_from_csv
on        original_types.column_name = types_from_csv.column_name
where     list_contains(getvariable('types'), original_types.column_type)
--and       is_same != true
order by  types_from_csv.column_type
;
```
The commented out elements in the array at the top of the script represent data types that are not allowed as elements for this parameter.

The list of allowed types is:
```
+-------------+
| column_type |
+-------------+
[...]

PR #3332, opened 2024-08-04
fix:https://github.com/duckdb/duckdb-web/issues/3330

rpbouman:


szarnyasg:
@pdet can you PTAL?

docs/data/csv/auto_detection.md
```diff
@@ -108,13 +108,31 @@ The type detection works by attempting to convert the values in each column to t
...
+The detected types can be individually overridden using the `types` option. This option takes either a list of type definitions (e.g., `types = ['INTEGER', 'VARCHAR', 'DATE']`) which overrides the types of the columns in-order of occurrence in the CSV file. Alternatively, `types` takes a `name` → `type` map which overrides options of individual columns (e.g., `types = {'quarter': 'INTEGER'}`).
 
-The type detection can be entirely disabled by using the `all_varchar` option. If this is set all columns will remain as `VARCHAR` (as they originally occur in the CSV file).
+The set of column types that may be specified using the `types` option is not as limited as the types available for the `auto_type_candidates` option: Any valid type definition is acceptable to the `types`-option. (To get a valid type definition, use the [`typeof()`]({% link docs/sql/functions/utility.md %}#typeofexpression) function, or use the `column_type` column  of the [`DESCRIBE`]({% link docs/guides/meta/describe.md %}) result.)
```
pdet, line 135: You can also use `select columns from sniff_csv('path/to/file.csv')` to get a struct with column names and types.

pdet:
LGTM, just have a minor comment.

docs/data/csv/auto_detection.md
szarnyasg: Thanks!

rpbouman: Hi Pedro! I checked that, and it looks to me like sniff_csv returns the same subset of types that is supported for the `types` option. They work of course, but what I was trying to convey with this piece of doc is that you can specify wildly exotic types like `STRUCT(a INTEGER[3], b VARCHAR[3])` and the CSV reader will actually swallow it and produce such values. 

I tested it like this:

```
copy (
  select *
  from   test_all_types()
)
to "C:\roland\projects\duckdb-bag-of-tricks\csv_dump_and_restore\test_all_types.csv";
```
```
set variable test_all_types_types = (
  select list(column_type)
  from   (
    describe 
    select *
    from   test_all_types()
  )
);
```

and then test whether all types are indeed reproduced:

```
select * from (describe from test_all_types())
except
select * from (describe from    read_csv(
  'C:\roland\projects\duckdb-bag-of-tricks\csv_dump_and_restore\test_all_types.csv'
  , types=getvariable("test_all_types_types")
  )
);
```
(results in 0 rows, as expected)

and then check if all values are read correctly:

```
select * exclude ("timestamp_tz","union")
from    read_csv(
          'C:\roland\projects\duckdb-bag-of-tricks\csv_dump_and_restore\test_all_types.csv'
        , types=getvariable("test_all_types_types")
        )
except
select * exclude ("timestamp_tz","union")
from test_all_types()
```

timestamp_tz and union columns do not produce the correct value but I decided to ignore that as that's probably some unrelated issue, and the principle still stands that the csv reader can reliably read arbitrary complex data types.

I was surprised by this actually and its hard to express how awesome that is! 

Anyway, I think your additional comment is also good but I'd add something to the effect that you then might lose type information whereas typeof and describe will give you a more complete definition of the type. 

Of course, your method of using the sniffer is completely appropriate when you only have CSV files, and no additional metadata. But I would then add some text that explains these two situations, because the feature of being to read arbitrary types from csv is really quite something!

rpbouman: @pdet Is my assessment correct that (save for these exceptions of union and timestamp_tz in this particular example), the CSV reader can handle deserialization of arbitrily complex types, provided the right type definition is passed to the reader? 

If so then I can amend the docs to describe this along with some type on when to use the sniffer detected types instead.

PR #3365, opened 2024-08-08
Add write_partition_columns option into statement copy page.

ykskb:
Fixes https://github.com/duckdb/duckdb-web/issues/3345. The original PR changes the default behavior of partitioned write (`COPY TO`) by not writing the partition column data into files. As I did not find any specific mention on this behavior in the doc, I thought documenting newly introduced option: `write_partition_columns` with default value: `false` would be self-explanatory and sufficient. Please let me know otherwise. Thanks!


szarnyasg:
Hi, thanks for the PR! The current `main` branch of the documentation target DuckDB v1.0, so I'm parking this PR v1.1 gets released (early September). In the long run we'll add a `latest` tag to the documentation but this was not yet created.

ykskb:
Hello @szarnyasg, Understood. Looking forward to v1.1 release :) Thanks!

PR #3385, opened 2024-08-14
Update the dbapi.md page

Tishj:
Mention named in-memory connections and reorder some of the information regarding connections and the different types of connections there are.


szarnyasg:
@Tishj the CI fails because of a broken link, can you PTAL?

Tishj:
@szarnyasg it's complaining about <https://duckdb.org/docs/api/python/reference/#duckdb.connect> but that's a valid link
I linked to it as `docs/api/python/reference/#duckdb.connect`

szarnyasg (approved):

PR #3476, opened 2024-08-26
fix typo in descriptions 'not_like_escape' and 'not_ilike_escape'

c-herrewijn:
Statements about case (in)sensitivity were reversed.


szarnyasg:
Thanks!

PR #3493, opened 2024-08-28
change `xor(x)` to `xor(x, y)`

asterix314:
`xor` takes 2 arguments. It's probably a typo to introduce it as `xor(x)`.


szarnyasg:
It's indeed a typo, thanks!

PR #3522, opened 2024-09-03
Update overview.md

soerenwolfers:
"when set" sounds like it should be possible to set to empty or null, neither of which is possible


szarnyasg:
This table is auto-generated, so I sent a PR to the main repo: https://github.com/duckdb/duckdb/pull/13724

szarnyasg:
Thanks for all 4 PRs!

Issue #3499, opened 2024-08-29
Rework Profiling Docs


maiadegraaf:
We should update the docs with all the recent changes around profiling.

Relevant PRs:
- https://github.com/duckdb/duckdb/pull/11101
- https://github.com/duckdb/duckdb/pull/12498
- https://github.com/duckdb/duckdb/pull/13234
- https://github.com/duckdb/duckdb/pull/13354
- https://github.com/duckdb/duckdb/pull/13430
- https://github.com/duckdb/duckdb/pull/13480
- https://github.com/duckdb/duckdb/pull/13497

Relevant Doc Pages
- [Profiling](https://duckdb.org/docs/dev/profiling.html)
- [`EXPLAIN`](https://duckdb.org/docs/guides/meta/explain)
- [`EXPLAIN ANALYZE`](https://duckdb.org/docs/guides/meta/explain_analyze)
- [Profiling Queries Pragmas](https://duckdb.org/docs/configuration/pragmas#profiling-queries)

@taniabogatsch

taniabogatsch:
I agree; maybe @szarnyasg can give input on how to (re?)structure our profiling/`EXPLAIN ANALYZE` documentation? And then this can go live with/around the upcoming release.

szarnyasg:
[This comment](https://github.com/duckdb/duckdb/issues/13195#issuecomment-2256260234) points out that the profiling page should mention the varios `enable_profiling` options.

Tishj:
Was <https://github.com/duckdb/duckdb/pull/13649> already covered?

Tishj:
https://github.com/duckdb/duckdb/pull/13160 might also need to be covered on the C-API page

PR #3530, opened 2024-09-04
[Metrics] Update profiling pages to reflect recent changes

maiadegraaf:
Closes: #3499


docs/dev/profiling.md
```diff
@@ -7,177 +7,371 @@ redirect_from:
...
 ```
 
-> This file is overwritten with each query that is issued. If you want to store the profile output for later it should be copied to a different file.
+The contents of the outputted file:
+```json
+{
+  "operator_timing": 0.000372,
+  "operator_cardinality": 0,
+  "extra_info": {},
+  "query_name": "SELECT name\nFROM students\nJOIN exams USING (sid)\nWHERE name LIKE 'Ma%';",
```
maiadegraaf, line 187: Potentially move these two JSON examples to a separate file? What do you think, @szarnyasg?

docs/configuration/pragmas.md
```diff
@@ -301,7 +301,7 @@ PRAGMA enable_profile;
 
 ##### Profiling Format
 
-The format of the resulting profiling information can be specified as either `json`, `query_tree`, or `query_tree_optimizer`. The default format is `query_tree`, which prints the physical operator tree together with the timings and cardinalities of each operator in the tree to the screen.
+The format of the resulting profiling information can be specified as either `json`, `query_tree`, or `query_tree_optimizer`. The default format is `query_tree`, which prints the logical query plan together with the timings and cardinalities of each operator in the tree to the screen.
```
taniabogatsch, line 304: `physical query plan`, as we print the specific operators instead of the logical operations.

docs/configuration/pragmas.md
```diff
@@ -301,7 +301,7 @@ PRAGMA enable_profile;
...
 
 To return the logical query plan as JSON:
```
taniabogatsch, line 306: `physical` query plan, no? We are also mixing it in the different examples. E.g., 
```
To return the physical query plan:
...
SET enable_profiling = 'query_tree_optimizer';
```

docs/configuration/pragmas.md
```diff
@@ -356,6 +356,24 @@ The output of this mode shows how long it takes to apply certain optimizers on t
 SET profiling_mode = 'detailed';
 ```
 
+```sql
+SET profiling_mode = 'standard';
+```
+
+#### Custom Profiling Metrics
+
+By default, all metrics are enabled, but they can be toggled on or off individually. This `PRAGMA` accepts a JSON object with the metric names as keys and a boolean value to enable or disable the metric. The metrics set by this `PRAGMA` will override the default settings.
```
taniabogatsch, line 365: Hmm, maybe `by default, most metrics are enabled`, as we do not include the detailed profiling node metrics.

docs/configuration/pragmas.md
```diff
@@ -356,6 +356,24 @@ The output of this mode shows how long it takes to apply certain optimizers on t
...
+
+> Note This only affects the metrics when the `enable_profiling` is set to `json`. The `query_tree` and `query_tree_optimizer` formats will always a default set of metrics.
```
taniabogatsch, line 367: > The `query_tree` and `query_tree_optimizer` always use a default set of metrics.

docs/dev/profiling.md
```diff
@@ -7,177 +7,371 @@ redirect_from:
 
 Profiling is important to help understand why certain queries exhibit specific performance characteristics. DuckDB contains several built-in features to enable query profiling that will be explained on this page.
 
-For the examples on this page we will use the following example data set:
+## `EXPLAIN` Statement
+
+The first step to profiling a database engine is figuring out what execution plan the engine is using. The [`EXPLAIN`]({% link docs/guides/meta/explain.md %}) statement allows you to peek into the query plan and see what is going on under the hood.
```
taniabogatsch, line 12: > A first step to profiling a duckdb can include examining the plan of a query.

docs/dev/profiling.md
```diff
@@ -7,177 +7,371 @@ redirect_from:
...
+
+## `EXPLAIN ANALYZE` Statement
+
+The query plan helps understand the performance characteristics of the system. However, often it is also necessary to look at the performance numbers of individual operators and the cardinalities that pass through them. This is where the [`EXPLAIN_ANALYZE`]({% link docs/guides/meta/explain_analyze.md %}) statement comes in, which pretty-prints the query plan and also executes the query, providing the actual run-time performance numbers.
```
taniabogatsch, line 16: I very much like this section! 👍

docs/dev/profiling.md
```diff
@@ -7,177 +7,371 @@ redirect_from:
...
+or
 ```sql
-CREATE TABLE students (sid INTEGER PRIMARY KEY, name VARCHAR);
-CREATE TABLE exams (cid INTEGER, sid INTEGER, grade INTEGER, PRIMARY KEY (cid, sid));
+PRAGMA enable_profile;
+```
+
+### Profiling Format
+
+The profiling can be output in several formats. When not specified, the default is `query_tree`, which prints the logical query plan with the timings and cardinalities of each operator in the tree to the screen.
```
taniabogatsch, line 38: > The profiling output is available in different formats.

`physical query plan`

taniabogatsch (commented):
I've added a few comments. It would be good to proof read the correct usage of logical vs. physical plan in the files. Additionally, we should at least add a note somewhere to the C API, and that it is also possible to turn the output 'off' and retrieve it via the API?

docs/configuration/pragmas.md
maiadegraaf: Maybe `by default, all metrics are enabled, excluding those enabled by detailed profiling` or too long?

docs/dev/profiling.md
```diff
@@ -7,177 +7,371 @@ redirect_from:
...
 
-INSERT INTO students VALUES (1, 'Mark'), (2, 'Hannes'), (3, 'Pedro');
-INSERT INTO exams VALUES (1, 1, 8), (1, 2, 8), (1, 3, 7), (2, 1, 9), (2, 2, 10);
+```SQL
+PRAGMA enable_profiling = 'json';
 ```
 
-## `EXPLAIN` Statement
+```sql
+# prints the physical operator tree
```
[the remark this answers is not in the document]
maiadegraaf: You mean `pragma.md` and `profiling.md`? I was wondering about this as well. I think it's useful to have all the profiling-related pragmas in this file too, but maybe we can reformat them into a table? And for more info they can go to the pragma page?

docs/dev/profiling.md
```diff
@@ -7,177 +7,371 @@ redirect_from:
...
+
+By default, the profiling output is printed to the console, but can be directed to a file using the following pragma:
+
+```SQL
+PRAGMA profiling_output = 'filename';
+```
+
+> Warning The file contents will be overwritten for every new query that is issued, hence the file will only contain the profiling information of the last query that is run.
+
+### Custom Profiling Metrics
```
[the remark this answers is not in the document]
maiadegraaf: You mean pragma.md and profiling.md? I was wondering about this as well. I think it's useful to have all the profiling-related pragmas in this file too, but maybe we can reformat them into a table? And for more info they can go to the pragma page?

docs/dev/profiling.md
```diff
@@ -7,177 +7,371 @@ redirect_from:
...
+| `OPERATOR_TIMING`       | `uint64`    |   ✅   |    ✅     | The time taken by the operator                                                     |
+| `OPERATOR_TYPE`         | `string`    |       |    ✅     | The name of the operator                                                           |
+| `QUERY_NAME`            | `string`    |   ✅   |          | The input query                                                                    |
+| `RESULT_SET_SIZE`       | `uint64`    |   ✅   |    ✅     | The size of the result in bytes                                                    |
+
+### Cumulative Metrics
+
+DuckDB also supports several cumulative metrics, which are available in all nodes. In the `QUERY_ROOT` node, these metrics are the sum of the specific metric in all the operators in the query. In the `OPERATOR` nodes, these metrics are the sum of the operator's specific metric as well as those of all its children recursively.
+
+These metrics can be used without turning on the specific metric.
```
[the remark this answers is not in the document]
maiadegraaf: Hmm yeah what I mean is that the cumulative metrics can be used without the metrics they're counting being turned on. I do think it's important to note, but maybe there is a better way of phrasing it.

docs/dev/profiling.md
```diff
@@ -7,177 +7,309 @@ redirect_from:
...
+Additionally, the following metrics are available to support the optimizer metrics:
+- `ALL_OPTIMIZERS` - Turns on all optimizer metrics, and measures the time taken by the optimizer parent node
+- `CUMMULATIVE_OPTIMIZER_TIMING` - The cumulative sum of all optimizer metrics, can be used without turning on all optimizer metrics.
 
-The output of `EXPLAIN ANALYZE` contains the estimated cardinality (`EC`), the actual cardinality, and the execution time for each operator.
+### Planner
 
-It is also possible to save the query plan to a file, e.g., in JSON format:
+The `PLANNER` is responsible for generating the logical plan. Currently, two metrics are measured in the `PLANNER`:
+- `PLANNER` - The time taken to generate the logical plan
```
taniabogatsch, line 85: `The time taken to generate the logical plan from the parsed SQL nodes`.

docs/dev/profiling.md
```diff
@@ -7,177 +7,309 @@ redirect_from:
...
+- `PLANNER_BINDING` - The time taken to bind the logical plan
+
+### Physical Planner
+
+The `PHYSICAL_PLANNER` is responsible for generating the physical plan. The following are the metrics supported in the `PHYSICAL_PLANNER`:
```
taniabogatsch, line 90: `generating the physical plan from the logical plan`

docs/guides/meta/explain.md
```diff
@@ -3,40 +3,29 @@ layout: docu
...
-
 ```sql
 EXPLAIN SELECT * FROM tbl;
 ```
 
-By default only the final physical plan is shown. In order to see the unoptimized and optimized logical plans, change the `explain_output` setting:
+The `EXPLAIN` statement displays the physical plan, i.e., the query plan that will get executed,
+and is enabled by prepending the query with `EXPLAIN`.
+The physical plan is a tree of operators that are executed in a specific order to produce the result of the query,
+and is generated by the query optimizer.
```
taniabogatsch, line 13: `and is generated by the query optimizer.`, let's remove this, the optimizer only rewrites the existing physical tree, which is generated by the planner.

taniabogatsch (changes requested):
Hey - I've mostly added nits and a few more changes.

docs/dev/profiling.md
```diff
@@ -7,177 +7,309 @@ redirect_from:
...
-                             │ Filters: name>=Ma AND name│
-                             │  <Mb AND name IS NOT NULL │
-                             │   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
-                             │           EC: 1           │
-                             └───────────────────────────┘
-```
+## Metrics
 
-Note that the query is not actually executed – therefore, we can only see the estimated cardinality (`EC`) for each operator, which is calculated by using the statistics of the base tables and applying heuristics for each operator.
+There are two types of nodes in the query tree; the `QUERY_ROOT`, and `OPERATOR` nodes.  The `QUERY_ROOT` refers exclusively to the top level node and the metrics it contains are measured over the entire query. The `OPERATOR` nodes refer to the individual operators in the query plan. Some metrics are only available for `QUERY_ROOT` nodes, while others are only available for `OPERATOR` nodes.  The table below describes each metric, as well as which nodes they are available for.
```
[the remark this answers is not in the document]
maiadegraaf: Good catch!

docs/dev/profiling.md
maiadegraaf: I think this is a good solution :)

docs/configuration/pragmas.md
```diff
@@ -356,6 +366,28 @@ The output of this mode shows how long it takes to apply certain optimizers on t
...
 ```
 
+```sql
+SET profiling_mode = 'standard';
+```
+
+#### Custom Profiling Metrics
+
+By default, all metrics are enabled except those activated by detailed profiling.
+Each metrics, including those from detailed profiling,
```
Tishj, line 376: metrics -> metric

docs/dev/profiling.md
```diff
@@ -7,177 +7,209 @@ redirect_from:
 
 Profiling is important to help understand why certain queries exhibit specific performance characteristics. DuckDB contains several built-in features to enable query profiling that will be explained on this page.
 
-For the examples on this page we will use the following example data set:
+## `EXPLAIN` Statement
 
-```sql
-CREATE TABLE students (sid INTEGER PRIMARY KEY, name VARCHAR);
-CREATE TABLE exams (cid INTEGER, sid INTEGER, grade INTEGER, PRIMARY KEY (cid, sid));
+A first step to profiling a duckdb can include examining the query plan. The [`EXPLAIN`]({% link docs/guides/meta/explain.md %}) statement allows you to peek into the query plan and see what is going on under the hood.
```
Tishj, line 12: I think this would fit better:
A first step to profiling a query

docs/configuration/pragmas.md
```diff
@@ -301,25 +301,35 @@ PRAGMA enable_profile;
...
+To return the physical query plan with optimizer and planner timings:
 
 ```sql
 SET enable_profiling = 'query_tree_optimizer';
 ```
+For more information on the profiling mode, see [profiling mode](#profiling-mode).
+
+###### Disabling Output
+
+It is also possible to disable outputting profiling information. This is specifically useful when accessing the profiling information:
```
taniabogatsch, line 328: `when accessing the profiling information through API calls`

docs/configuration/pragmas.md
```diff
@@ -337,7 +347,7 @@ PRAGMA disable_profile;
 
 ##### Profiling Output
 
-By default, profiling information is printed to the console. However, if you prefer to write the profiling information to a file the `PRAGMA` `profiling_output` can be used to write to a specified file.
+By default, profiling information is printed to the console however, if you prefer to write the profiling information to a file the `PRAGMA` `profiling_output` can be used to write to a specified file.
```
taniabogatsch, line 350: `console, however,`

docs/dev/profiling.md
```diff
@@ -7,177 +7,209 @@ redirect_from:
...
+These metrics are automatically enabled, and added to the enabled settings, when the `profiling_mode` is set to `detailed`, however, they can also be toggled individually. 
+
+### Optimizers
+
+At the `QUERY_ROOT` level, there are also metrics that measure the time taken by each [optimizer]({% link docs/internals/overview.md %}#optimizer). These metrics are only available when the specific optimizer is enabled. The available optimizations can be queried using the [`duckdb_optimizers() table function`]({% link docs/sql/meta/duckdb_table_functions.md %}#duckdb_optimizers).
+
+Each optimizer has a corresponding metric that follows the template: `OPTIMIZER_<OPTIMIZER_NAME>`. For example, the `OPTIMIZER_JOIN_ORDER` metric corresponds to the `JOIN_ORDER` optimizer.
+
+Additionally, the following metrics are available to support the optimizer metrics:
+- `ALL_OPTIMIZERS` - Turns on all optimizer metrics, and measures the time taken by the optimizer parent node
```
taniabogatsch, line 79: Missing dot. Also for the other bullet lists, not sure if we want them or not haha.

docs/guides/meta/explain.md
```diff
@@ -46,70 +34,62 @@ EXPLAIN
...
+```
+
+```sql
+# Shows only the optimized plan.
+PRAGMA explain_output = 'optimized_only';
+```
+
+```sql
+# Shows both the physical and optimized plans.
+PRAGMA explain_output = 'all';
```
taniabogatsch, line 90: I think we usually move the comment outside on top of the SQL block.

taniabogatsch (changes requested):

szarnyasg (approved):

PR #3536, opened 2024-09-05
Add section on disabling autoinstalling/autoloading

szarnyasg:
These are important security configuration flags that were so far omitted from the "Securing DuckDB" part of the docs – they were only mentioned in the blog and in the configuration flags overview.

### DIFF
--- a/docs/operations_manual/securing_duckdb/securing_extensions.md
+++ b/docs/operations_manual/securing_duckdb/securing_extensions.md
@@ -51,6 +51,22 @@ Despite this, installing and loading DuckDB extensions from the community extens
 SET allow_community_extensions = false;
 ```
 
+## Disabling Autoinstalling and Autoloading Known Extensions
+
+By default, DuckDB automatically installs and loads known extensions.
+
+To disable autoinstalling known extensions, run:
+
+```sql
+SET autoinstall_known_extensions = true;
+```
+
+To disable autoloading known extensions, run:
+
+```sql
+SET autoload_known_extensions = true;
+```
+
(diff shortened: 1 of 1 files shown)


carlopi:
I'd say those are different than other security parts, since those are not 1-way settings, so can be easily circumvented if you are running untrusted code.

I would consider something like:

----
By default, DuckDB automatically installs and loads core functionality extensions on first usage.
See [link] for a complete explanation. (do we have such a link? I guess we should, list is at https://github.com/duckdb/duckdb/blob/a6e32b115826ba543e32a733cb92f68fd0549186/src/include/duckdb/main/extension_entries.hpp#L514, and spatial is in some cases also added)

To disable this behaviour and rely on explicit SQL syntax like: `INSTALL httpfs` (one time) and `LOAD httpfs` (once per session), consider:
```sql
SET autoinstall_known_extensions = true;
```
to avoid fetching (potentially remote) extensions and moving them to be installable, but allowing loading of extensions, or
```sql
SET autoinstall_known_extensions = true;
SET autoload_known_extensions = true;
```
to disable both installing (= fetching potentially remote extensions) and loading (= adding functionality to the system).

Review other setting connected to extensions at: .... (here I would also put how to set a specific extension and secret directory)

----

I would also mention that you can lock configuration, since that would allow to make:
```
SET autoinstall_known_extensions = true;
SET autoload_known_extensions = true;
PRAGMA lock_configuration (?????)
```
irreversible.

szarnyasg:
I added the locking comment, more rework on this later.

PR #3593, opened 2024-09-09
fix link to duckdb-node-neo

archiewood:
Link was erroneously to duckdb-node repo. This PR corrects the link


szarnyasg:
Thanks!

PR #3608, opened 2024-09-11
Mention extensions can be installed from a repository

Tishj:


Tishj:
I also noticed there is no mention of `FORCE INSTALL`, I know it involves versioning but that's about it, so I don't think I'm the right person to add this information, but I think it should be added.

szarnyasg:
Thanks, looks great!

PR #3613, opened 2024-09-12
Release v1.1.1

szarnyasg:


carlopi:
I think we can also remove the note on:

> If you experience issues while running the DuckDB CLI on Windows, please install the latest <a 
> href="https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170">Microsoft Visual C++ 
>  Redistributable</a>.

and revert the custom links for Windows to be the release ones

PR #3622, opened 2024-09-12
replace necessary strings before zip offline package #3605

### LINKED ISSUE #3605: After offline deployment, got a 404 error response when accessing.
I followed the offline deployment steps provided in the [official offline deploy manual](https://duckdb.org/docs/guides/offline-copy.html) to deploy duckdb-web offline. While I can access the homepage and some other pages normally, when I use the top search icon, enter a keyword, and click on a search result, a 404 error occurs. The analysis shows that the href path in the search result does not have the .html suffix, which is necessary for static deployment. When I manually add the .html suffix in the address bar, the relevant pages are successfully loaded.

     This issue occurs not only in the navigation of the search results but also in the sidebar navigation on the left side of the page.

coloraven:
Closes #3605

### DIFF
--- a/scripts/offline_docs_replace.py
+++ b/scripts/offline_docs_replace.py
@@ -2,79 +2,36 @@ import os
 import re
 
 
-# Replace href="(/docs[^"]*[^/])" with href="$1.html" in HTML files
-def process_html_file(file_path):
-    # print("procesing html file:\t",file_path)
-    with open(file_path, "r", encoding="UTF-8") as file:
-        content = file.read()
-
-    # Replace href="(/docs[^"]*[^/])" with href="$1.html"
-    content_updated = re.sub(
-        r'href="(/docs[^"]*(?<!/)(?<!\.html))"', r'href="\1.html"', content
-    )
-
-    # Write the updated content back to the file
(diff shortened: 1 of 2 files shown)


scripts/offline_docs_replace.py
```diff
@@ -0,0 +1,80 @@
...
+        pass
+
+
+# Recursively find and apply changes to files
+def process_files(root_dir, file_ext, processing_function):
+    for subdir, dirs, files in os.walk(root_dir):
+        for file in files:
+            if file.endswith(file_ext):
+                file_path = os.path.join(subdir, file)
+                if file_path == r"./duckdb-docs/data/search_data.json":
```
szarnyasg, line 62: I think with #3617, this is `if` statement and `process_search_data_json` are no longer necessary.

szarnyasg (changes requested):

scripts/offline_docs_replace.py
coloraven: search_data.json replacing had removed now

coloraven: BTW, could you compress the HTML, JSON, CSS, JS, and other files that can be compressed to reduce the size of the `duckdb-docs.zip`?

PR #3634, opened 2024-09-14
add update time to the doc page,and add snippets site access enter

coloraven:


szarnyasg:
Hello @coloraven thanks for the PR. Let's split this into two. We currently do not directly promote 3rd-party sites such as duckdbsnippets.com – we only do so indirectly via the [awesome DuckDB repository](https://github.com/davidgasquez/awesome-duckdb).

szarnyasg:
Thanks!

PR #3635, opened 2024-09-14
Change link text to "Text functions"

shoffmeister:
The target page is named "Text functions"


szarnyasg:
That's right, thanks for the fix!

PR #3639, opened 2024-09-14
Raspberry Pi build and minor fixes

szarnyasg:
Add rpi build instructions as suggested in https://github.com/duckdb/duckdb/issues/13855#issuecomment-2341539339

### DIFF
--- a/docs/dev/building/build_instructions.md
+++ b/docs/dev/building/build_instructions.md
@@ -50,9 +50,9 @@ git clone https://github.com/duckdb/duckdb
 
 We recommend creating a full clone of the repository. Note that the directory uses approximately 1.3 GB of disk space.
 
-### Building DuckdB
+### Building DuckDB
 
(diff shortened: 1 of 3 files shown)


docs/dev/building/build_instructions.md
```diff
@@ -104,6 +104,21 @@ Once the build finishes successfully, you can find the `duckdb.exe` binary in th
 ./duckdb.exe
 ```
 
+## Raspberry Pi (32-bit)
+
+On 32-bit Raspberry Pi controllers, you need to add the [`-latomic` link flag](https://github.com/duckdb/duckdb/issues/13855#issuecomment-2341539339).
```
vgrechin, line 109: From my perspective Raspberry 3,4,5 microcomputers are so different from any family of controllers like STM32, ESP32 or Raspberry Pico to name a few. More appropriate generalization might be:
a) Raspberry Pi boards
b) Single-board computers or SBC
c) microcomputers

szarnyasg: Good point, that was sloppy writing on my end. I pushed a fix via #3642

PR #3686, opened 2024-09-18
Update gdal.md to correct typos

gregorywaynepower:
Change

"The spatial extension does not only enable _imporintg_ geospatial file formats (through the ST_Read function), it also enables _exporing_ DuckDB tables to different geospatial vector formats through a GDAL based COPY function."

to

"The spatial extension does not only enable _importing_ geospatial file formats (through the ST_Read function), it also enables _exporting_ DuckDB tables to different geospatial vector formats through a GDAL based COPY function."


szarnyasg:
Thanks!

PR #3704, opened 2024-09-23
remove BUILD_INET

flashmouse:
I want to remove the introduction of `BUILD_INET` flag because INET now is a out source tree extension and  in fact  `BUILD_INET` now is not exist.


szarnyasg:
Thanks!

PR #3714, opened 2024-09-24
Remove MSVC warning

szarnyasg:
cc @carlopi


carlopi (approved):
Thanks, I think indeed this should be solved in v1.1.1

PR #3734, opened 2024-09-27
Describes the current fact that the read_csv function and the COPY statement support different parameters.

michaelpopes:
Describes the current fact that the read_csv function and the COPY statement support different parameters.

### DIFF
--- a/docs/data/csv/overview.md
+++ b/docs/data/csv/overview.md
@@ -82,7 +82,7 @@ The DuckDB CSV reader can automatically infer which configuration flags to use b
 
 ## Parameters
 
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks, this is a good start but the new phrasing is quite vague. Which options did you encounter that are not supported by COPY?

michaelpopes:
Oh, good question!  I have noticed the following differences:
1. COPY statement accepts "DELIMITER" and "DELIM", while the read_csv function only accepts "DELIM".
2. The read_csv function accepts "ALL_VARCHAR", while the COPY statement doesn't support it.

PR #3741, opened 2024-09-27
Update aggregates.md

soerenwolfers:
Fixes https://github.com/duckdb/duckdb/issues/14144 (see discussion there for proof that this formula is now correct).


szarnyasg:
Thanks!

PR #3746, opened 2024-09-29
Fix an ambiguous anaphoric reference

tuffnatty:
The original text, "These functions supports the same two location notations as the previous functions.", seems to have outdated, as there are no JSON location notations referenced in the text above.


szarnyasg:
Thanks!

Issue #3753, opened 2024-09-30
[duckdb/#12942] New `information_schema` views need documentation


tuffnatty:
See duckdb/duckdb#12942

szarnyasg:
Fixed via #3754

PR #3754, opened 2024-09-30
Document `information_schema.constraint_column_usage` view

tuffnatty:
(towards the resolution of #3753)


szarnyasg:
This is great, thanks! I think is sufficient to fix #3753.

PR #3776, opened 2024-10-03
Revert hackery of scrolling to top via iframe include

carlopi:
Code in _include/iframe.html, looks like:
```html
<div style="position:relative; padding-bottom:60%; width: 100%">
    <iframe src="{{ include.src }}" title="{{ include.title }}" style="position: absolute; top: 0; left: 0; width: 100%; height: 100%;"></iframe>
</div>
```
This is the same used for embedding the Wasm shell.

Unsure why hackery was needed, I guess to reduce flickering?
If that's the case, this seems to work the same on my side, but would need to be double checked / possibly there are other considerations in place.


Alex-Monahan:
The odd behavior was that the page was auto-scrolling down to the iframe once it started loading. Did you happen to check if that still happened with this? I think JupyterLite calls `.focus()` or something...

PR #3782, opened 2024-10-04
array_cosine_distance = 1 - array_cosine_similarity

jgrg:
`array_cosine_distance` is not equal to one minus itself!


szarnyasg:
Indeed, thanks!

PR #3786, opened 2024-10-06
Correct statements that are no longer true in 1.1

soerenwolfers:
1.1 throws on scalar subqueries that return multiple rows.


szarnyasg:
Thanks!

PR #3787, opened 2024-10-06
Update subqueries.md

soerenwolfers:


docs/sql/expressions/subqueries.md
```diff
@@ -116,21 +116,23 @@ The `EXISTS` operator tests for the existence of any row inside the subquery. It
...
 ```sql
-SELECT EXISTS (SELECT * FROM grades WHERE course = 'History') AS history_grades_present;
+SELECT EXISTS (FROM grades WHERE course = 'History') AS history_grades_present;
 ```
 
 | history_grades_present |
 |-----------------------:|
 | false                  |
 
+> The subqueries in the examples above have made used of the fact that you can omit the `SELECT *` in duckdb, which would be required in other SQL implementations but cannot fulfil any purpose in `EXISTS` and `NOT EXISTS` subqueries.
```
soerenwolfers, line 134:
```suggestion
> The subqueries in the examples above make use of the fact that you can omit the `SELECT *` in DuckDB, which would be required in other SQL implementations but cannot fulfil any purpose in `EXISTS` and `NOT EXISTS` subqueries.
```

szarnyasg:
Thanks, changed the wording a bit, merging now.

PR #3843, opened 2024-10-16
Join order performance guide

szarnyasg:


docs/guides/performance/join_operations.md
```diff
@@ -3,9 +3,33 @@ layout: docu
...
 
-## Forcing a Join Order
+## How to Force a Join Order
 
 DuckDB has a cost-based query optimizer, which uses statistics in the base tables (stored in a DuckDB database or Parquet files) to estimate the cardinality of operations.
+
+### Turn off the Join Order Optimizer
+
+```sql
+SET disabled_optimizers = 'join_order';
```
lnkuiper, line 13: Potentially, users would also want to disable left/right swapping for joins with `'build_side_probe_side'`, then they will build a left-deep tree by chaining `JOIN` clauses. They can create a bushy plan by using parentheses.

lnkuiper (commented):
Looks good! I have one comment, which we might want to include, but is not required since it might be a bit too detailed for users.

docs/guides/performance/join_operations.md
szarnyasg: Thanks! I worked this into the PR.

Tmonster:
I know I'm late to this, but looks good to me 👍

PR #3871, opened 2024-10-18
Fix auto* instructions

### LINKED ISSUE #3869: Issue found on page 'Securing Extensions' - Confusing recommendation
Dear team,
Under [Disabling Autoinstalling and Autoloading Known Extensions](https://duckdb.org/docs/operations_manual/securing_duckdb/securing_extensions.html#disabling-autoinstalling-and-autoloading-known-extensions) it says:
To disable autoinstalling known extensions, run:
```
SET autoinstall_known_extensions = true;
```
To disable autoloading known extensions, run:
```
SET autoload_known_extensions = true;
```

I think that both of these suggestions are incorrect and they should instead be set to false, not true.

If my assumption there is incorrect I would suggest to make the configuration option more clear what it means.

Thank you for looking into this, Roy

Page URL: <https://duckdb.org/docs/operations_manual/securing_duckdb/securing_extensions.html>

szarnyasg:
Fixes #3869

### DIFF
--- a/docs/operations_manual/securing_duckdb/securing_extensions.md
+++ b/docs/operations_manual/securing_duckdb/securing_extensions.md
@@ -58,13 +58,13 @@ By default, DuckDB automatically installs and loads known extensions.
 To disable autoinstalling known extensions, run:
 
 ```sql
-SET autoinstall_known_extensions = true;
+SET autoinstall_known_extensions = false;
 ```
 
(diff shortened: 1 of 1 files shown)


carlopi (approved):

carlopi:
Unsure why this slipped, thanks

PR #3879, opened 2024-10-21
Add VSS Part 2

szarnyasg:
New blog post


_posts/2024-10-23-whats-new-in-the-vss-extension.md
```diff
@@ -0,0 +1,233 @@
...
+```
+</details>
+
+We can see that this plan is drastically simplified, but most importantly, the new `HNSW_INDEX_JOIN` operator replaces the `CROSS_PRODUCT` node that was there before and the estimated cardinality went from 5,000,000 to 50,000! Executing this query now takes about 0.15 seconds. That's an almost 66× speedup!
+
+This optimization was just recently added to the VSS extension, so if you've already installed `vss` for DuckDB v1.1.2, run the following command to get the latest version:
+
+```sql
+FORCE INSTALL vss;
+```
```
carlopi, line 229: I would go with `UPDATE EXTENSIONS` here, that is the better version of this (even tough it has side effects, those are mostly positive).

UPDATE EXTENSIONS updates all currently installed extensions but only if they are changed (check is an unexpensive GET with Etag).

UPDATE EXTENSIONS basically pays only if stuff has changed, so repeated usage have ammortized low cost, while FORCE INSTALL will still download the extension in every case.

PR #3900, opened 2024-10-21
Add extension path example to LOAD

Draiken:
Looking at the current page it's not clear that we can load extensions from a specified path. This fixes that.


szarnyasg:
Thanks!

PR #3907, opened 2024-10-22
Fix excel link in data_sources.md

yomgui:
The previous Excel link incorrectly led to the MySQL documentation page. It has now been updated to include three distinct links:
  * one for the spatial extension
  * one for importing Excel files
  * and another for exporting Excel files.


szarnyasg:
Thanks!

PR #3910, opened 2024-10-22
community_extensions hosted in duckdb.org

carlopi:
Move current community extensions markdowns into duckdb/duckdb-web.

There are still quite some work in wrapping everything properly, and I would like to add more stuff, but better get going.


szarnyasg:
Tested locally, looks great, thanks!

PR #3921, opened 2024-10-22
Preview image for community extension

carlopi:
I generated using a script like:

```
for filename in screenshot/*
do
  ext_name=${filename%.*}
  echo $ext_name
  chromium --headless --screenshot=../social_preview/preview_community_extension_$ext_name.png  "http://127.0.0.1:4000/community_extensions/screenshot/$ext_name.html" --window-size=500,240  -force-device-scale-factor=5
done;
```
starting with md files that looked like the one in `community_extensions/extensions` (I still need to PR for that to be properly up)

The results looks pleasant enough to me to get going, but if someone wants to make a pass on that it's very welcome. I will later share full repro.

Idea is taking screenshots (via chromium --headless), so that they share same styling as regular website.


szarnyasg:
Thanks!

tobilg:
@szarnyasg / @carlopi do these automatically get produced once new extensions are listed? I realized that the new dns extension has a missing OG image: #5934

PR #3937, opened 2024-10-22
Add UPDATE EXTENSIONS page

carlopi:
Informations are already available at https://duckdb.org/docs/extensions/versioning_of_extensions#updating-extensions, but I think feature might use it's own page.

To be considered adding an explanation of the various return-code, and merging both versions / linking this from INSTALL/LOAD page and possibly elsewhere.

### DIFF
--- a/docs/sql/statements/update_extensions.md
+++ b/docs/sql/statements/update_extensions.md
@@ -3,38 +3,43 @@ layout: docu
 title: UPDATE EXTENSIONS
 ---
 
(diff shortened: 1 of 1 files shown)


carlopi:
Unsure if this is handy to have is as-is, or better to iterate. Thanks

szarnyasg:
Great! I re-worded the page a bit and will merge it.

szarnyasg (approved):

PR #3941, opened 2024-10-23
[Community Extensions] Random selection of extension (decided by jekyll sample) on the bar

carlopi:
This is to be reviewed by the design team.

### DIFF
--- a/_data/menu_ce.json
+++ b/_data/menu_ce.json
@@ -24,5 +24,30 @@
       "page": "List of Community Extensions",
       "url": "/community_extensions/list_of_extensions"
     }
+  ],
+  "extensions": [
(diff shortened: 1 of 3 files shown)


szarnyasg:
The selection should be done in JS – I'll pass this on to Max.

carlopi:
Just curiosity, why JS side?

I think it's OK (better?) to have a given page have a fixed list (= refreshing will not scramble them), and that list to be determined at build time (so rebuilding the website will actually change the list).

Someone else picking this up is great, thanks!

szarnyasg:
The website builds are spaced out in time very randomly, sometimes we deploy every few minutes, sometimes there are no updates for days.

PR #3953, opened 2024-10-24
Update pivot.md

smilingthax:
Add note that FILTER clauses are currently ignored, fixes https://github.com/duckdb/duckdb/issues/11802.


szarnyasg:
Thanks!

PR #3957, opened 2024-10-24
Remove BUILD_ from the documented options

carlopi:
Connected to this PR https://github.com/duckdb/duckdb/pull/14531 (and other that might follow).

Since already 1.1.0 the `CORE_EXTENSIONS='a;b;c'` is the suggested way to build extensions, and the older BUILD_TPCD & co should be phased out going to 1.2.x. I think it's safe to just remove mentions of `BUILD_`, moving to the new methods. Possibly to be considered a warning tab saying:
> BUILD_HTTPFS=1 & co are being discontinued (although they will still might work).

### DIFF
--- a/docs/dev/benchmark.md
+++ b/docs/dev/benchmark.md
@@ -13,7 +13,7 @@ When making changes that have potential performance implications, it is importan
 To build the benchmark suite, run the following command in the [DuckDB repository](https://github.com/duckdb/duckdb):
 
 ```bash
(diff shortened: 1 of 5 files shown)


carlopi:
Note that this is not strictly needed, but I would sort out docs & co in parallel given I think this is the best direction, unsure if you have inputs @szarnyasg?

szarnyasg:
Great, thanks for keeping this up-to-date.

PR #3988, opened 2024-10-29
Iss3985

### LINKED ISSUE #3985: Document `NULL` behavior for `arg_min` and friends, and document `arg_min_null` and friends
I was recently looking for a way to get the entire row corresponding to the maximum of a column. `arg_max` should do the job! But I was surprised to see the following behavior:
```sql
create table foo (a integer, b text);
insert into foo values (1, 'a'), (2, 'b'), (3, null);
select * from foo;
```
```text
┌───────┬─────────┐
│   a   │    b    │
│ int32 │ varchar │
├───────┼─────────┤
│     1 │ a       │
│     2 │ b       │
│     3 │         │
└───────┴─────────┘
```
```sql
select arg_max(columns(*), a) from foo;
```
```text
┌───────┬─────────┐
│   a   │    b    │
│ int32 │ varchar │
├───────┼─────────┤
│     3 │ b       │
└───────┴─────────┘
```
Apparently, even though `3` is the maximum of column `a`, the "ignore `NULL`" behavior of aggregate functions applies even to the first argument of `arg_max`. So really `arg_max(arg, val)` is returning:

> The element of `arg` where `val` is maximized, among rows where both `arg` and `val` are not `NULL`.

And then the expression `arg_max(columns(*), a)` applies said behavior per-column. So really my example is shorthand for
`select arg_max(a, a), arg_max(b, a) from foo;` which is not guaranteed to choose the same row for the two `arg_max`'s. I consider this behavior to be pretty confusing, even if it is technically correct and consistent, so it warrants extra documentation. This also goes for `arg_min` and aliases `argmax`, `argmin`, `argMax`, `argMin`, `max_by`, and `min_by`.

Also, I was made aware in the Discord of the existence of `arg_max_null` and `arg_min_null` which _does_ consistently choose the same row like I originally wanted. For example:
```sql
select arg_max_null(columns(*), a) from foo;
```
```text
┌───────┬─────────┐
│   a   │    b    │
│ int32 │ varchar │
├───────┼─────────┤
│     3 │         │
└───────┴─────────┘
```
But these functions are not in the documentation yet.

szarnyasg:
Fixes #3985

### DIFF
--- a/docs/sql/functions/aggregates.md
+++ b/docs/sql/functions/aggregates.md
@@ -412,13 +412,14 @@ The table below shows the available general aggregate functions.
 | **Example** | `sum(A)` |
 | **Alias(es)** | - |
 
-### Nulls
+### Handling `NULL` Values
 
-All general aggregate functions except for `list` and `first` (and their aliases `array_agg` and `arbitrary`, respectively) ignore `NULL`s. 
-To exclude `NULL`s from `list`, you can use a [`FILTER` clause]({% link docs/sql/query_syntax/filter.md %}). To ignore `NULL`s from `first`, you can use the `any_value` aggregate. 
+All general aggregate functions except for [`list`](#listarg) and [`first`](#firstarg) (and their aliases [`array_agg`](#array_aggarg) and [`arbitrary`](#arbitraryarg), respectively) ignore `NULL`s.
+To exclude `NULL`s from `list`, you can use a [`FILTER` clause]({% link docs/sql/query_syntax/filter.md %}).
+To ignore `NULL`s from `first`, you can use the [`any_value` aggregate](#any_valuearg).
 
-All general aggregate functions except `count` return `NULL` on empty groups and groups without non-`NULL` inputs. 
-In particular, `list` does *not* return an empty list, `sum` does *not* return zero, and `string_agg` does *not* return an empty string in this case.
+All general aggregate functions except [`count`](#countarg) return `NULL` on empty groups and groups without non-`NULL` inputs.
+In particular, [`list`](#listarg) does *not* return an empty list, [`sum`](#sumarg) does *not* return zero, and [`string_agg`](#string_aggarg-sep) does *not* return an empty string in this case.
 
 ## Approximate Aggregates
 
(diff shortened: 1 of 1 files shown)


docs/sql/functions/aggregates.md
```diff
@@ -107,10 +107,12 @@ The table below shows the available general aggregate functions.
...
-| [`arg_min(arg, val, n)`](#arg_minarg-val-n) | Returns a `LIST` containing the `arg` expressions for the "bottom" `n` rows ordered by `val` ascending.  |
+| [`arg_max(arg, val)`](#arg_maxarg-val) | Finds the row with the maximum `val` and calculates the `arg` expression at that row. Rows where the value of the `arg` or `val` expression is `NULL` are ignored. This function is [affected by ordering](#order-by-clause-in-aggregate-functions). |
+| [`arg_max(arg, val, n)`](#arg_maxarg-val-n) | The generalized case of [`arg_max`](#arg_maxarg-val) for `n` values: returns a `LIST` containing the `arg` expressions for the top `n` rows ordered by `val` descending. This function is [affected by ordering](#order-by-clause-in-aggregate-functions). |
+| [`arg_max_null(arg, val)`](#arg_max_nullarg-val) | Finds the row with the maximum `val` and calculates the `arg` expression at that row. Rows where the `arg` expression evaluates to `NULL` are ignored.  This function is [affected by ordering](#order-by-clause-in-aggregate-functions). |
```
Doekeb, line 112: I think the `NULL` behavior is actually reversed. Rows where the `arg` expression evaluates to `NULL` are _not_ ignored (rows where the `val` expression evaluates to `NULL` are ignored). e.g.
```sql
D create or replace table foo (arg text, val integer);
D insert into foo values ('a', 4), ('e', null), (null, 9), ('b', 5);
D select * from foo;
┌─────────┬───────┐
│   arg   │  val  │
│ varchar │ int32 │
├─────────┼───────┤
│ a       │     4 │
│ e       │       │
│         │     9 │
│ b       │     5 │
└─────────┴───────┘
D select arg_max(arg, val) from foo;
┌───────────────────┐
│ arg_max(arg, val) │
│      varchar      │
├───────────────────┤
│ b                 │
└───────────────────┘
D select arg_max_null(arg, val) from foo;
┌────────────────────────┐
│ arg_max_null(arg, val) │
│        varchar         │
├────────────────────────┤
│                        │
└────────────────────────┘
```

PR #4024, opened 2024-11-05
speed adjustments

Franz-Kafka:
- defer javascripts
- search_data.json will be loaded only when activating search
- home typing animation faster


szarnyasg:
Thanks!

PR #4078, opened 2024-11-18
events posts, new blog design, color system preparations

Franz-Kafka:
- new blog design
- optimized blog single entry pages
- added new post type: events
- prepared post type media
- slightly updated menu
- updated thumbnails
- prepared new color system with CSS variables


szarnyasg:
Thanks'

PR #4115, opened 2024-11-20
Add column selection via list to star expression page

jzavala-gonzalez:
Mentioned how the COLUMNS expression can accept a list of column names. It is already implicitly covered when documenting the expansion of columns expressions, but having it at the top-level examples is more visible.


szarnyasg:
I agree that it's good to include this on the page. Thanks a lot!

PR #4144, opened 2024-11-25
various small improvements

Franz-Kafka:
- added sponsors with logos to DuckCon event entry
- monospace_table entries are now wrapped with code-tags so they have grey backgrounds
- fixed mobile menu


szarnyasg:
Thanks!

Issue #4035, opened 2024-11-09
Issue found on page 'Order Preservation'


soerenwolfers:
It's quite unclear to me what 

> For many operations, DuckDB preserves the insertion order of rows, similarly to data frame libraries such as Pandas. The following operations and components respect insertion order: The CSV reader

means: A user could be forgiven to conclude from this sentence, e.g., that aggregate functions are fed inputs in the "original table order" (see https://github.com/duckdb/duckdb/issues/14755), that sorting is stable (https://github.com/duckdb/duckdb/discussions/14941), or generally that arbitrarily complex and nested queries on data respects that original order (at least for the CSV reader), none of which I don't believe is the case.

I also wonder why the paragraph starts with "For many operations preserve..." and then ends with "The following operations and components respect ..: `<EXACTLY ONE COMPONENT AND NOT A SINGLE OPERATION>` The CSV reader". 

Page URL: <https://duckdb.org/docs/sql/dialect/order_preservation.html>

soerenwolfers:
@szarnyasg I think this is one of the biggest causes for  confusion about duckdb, so a few words about SQLs general approach and duckdb's actual specific enhancements would be super useful. I can't help out in this case because I genuinely don't know what duckdb's philosophy here is not what the current specifics look like.

szarnyasg:
Hi @soerenwolfers, great point. A fix is coming. Still a lot more could be said about the topic but this quick patch should clarify the topic.

PR #4159, opened 2024-11-27
Update order_preservation.md

soerenwolfers:
follow up to https://github.com/duckdb/duckdb-web/issues/4035 to explicitly answer misunderstandings exhibited in https://github.com/duckdb/duckdb/issues/14755 and https://github.com/duckdb/duckdb/issues/14941

By the way: Do you folks have recommendations for the common situation that one has a large table that's sorted by multiple columns (C1,...,Cn) and one wants to successively perform aggregations over Cn then Cn-1, ...?  In that situation, there would be a trivial O(N) solution to doing each of the aggregations by hand, simply by checking group boundaries on the fly and producing aggregate values in an order preserving manner. Feels like the following duckdb snippet
```sql
SELECT  
  agg1(x)
FROM
(
  SELECT 
    agg2(x) AS x
  FROM 
  (
     ...
  )
  GROUP BY C2
)
GROUP BY C1


 would be inefficient in comparison?


szarnyasg:
Thanks! I'll bring up the question with the team and get back to you.

PR #4164, opened 2024-11-28
Update window_functions.md

soerenwolfers:
- Clarify that restriction applies to all window functions, not just aggregate functions
- Give example of what exactly isn't possible
- Put the entire constraint warning into a box


szarnyasg:
Thanks!

PR #4169, opened 2024-11-28
Fix sampling method

### LINKED ISSUE #4166: Issue found on page 'Samples'
First line is: "Select a sample of 5 rows from tbl using reservoir sampling:"
```
SELECT *
FROM tbl
USING SAMPLE 5;
```
but the snippet here perform system sampling, that is a OK (expecially if there is no correlation between rows) but NOT the same as reservoir based samples method.

Page URL: <https://duckdb.org/docs/sql/samples.html>

szarnyasg:
Fixes #4166

### DIFF
--- a/docs/sql/samples.md
+++ b/docs/sql/samples.md
@@ -8,7 +8,7 @@ Samples are used to randomly select a subset of a dataset.
 
 ### Examples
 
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!!

PR #4175, opened 2024-11-29
Add ruby-duckdb client

gregszero:
add https://github.com/suketa/ruby-duckdb to repositories.md as Ruby client


szarnyasg:
Thanks!

Issue #3058, opened 2024-06-14
Clarify the meaning of max_temp_directory_size


jhrcek:
Hello,
could you please clarify the meaning of `max_temp_directory_size` mentioned [here](https://duckdb.org/docs/configuration/overview#global-configuration-options)?

I see the default size is 0. But it's not clear to me whether this default setting means "there's no limit" or "the limit is 0, if you want to enable offloading to disk, you have to set it to >0".

Tishj:
Thanks for opening the issue, I agree this is something we should document
<https://github.com/duckdb/duckdb/pull/10978> is the source for this option

jhrcek:
Thank you for the reference PR, the description answers most of my questions for now.

PR #4189, opened 2024-11-30
Update overview.md

soerenwolfers:
Solves https://github.com/duckdb/duckdb-web/issues/3058 --- please do check whether both `NULL` and empty string actually disable `temp_directory`.


szarnyasg:
Thanks! This table is auto-generated so I'll track this change and update the related entry in the DuckDB codebase.

Issue #4037, opened 2024-11-10
The `COPY ... TO` Options should indicate that some options aren't supported in some formats. 


kennycarruthers:
The documentation for `COPY ... TO` Options should indicate that some options might not be available depending on the `FORMAT` being used. 

https://duckdb.org/docs/sql/statements/copy.html#copy--to-options

For example, when `FORMAT = json`, attempting to include the `return_files` option generates a `BinderException` because of an "Unknown option". Including `return_files` when `FORMAT = csv` or `FORMAT = parquet` does not produce any errors. 

When possible, it would be helpful if the documentation indicated which formats support which options (as it does lower down where there are specific sections for `csv`, `json` and `parquet` specific options.

soerenwolfers:
It seems to me that all of these options should be supportable by all formats, so rather than painstakingly documenting current omissions in the matrix, I think you should rather file issues for non-supported options you run into.

PR #4202, opened 2024-11-30
Update copy.md

soerenwolfers:
Solves https://github.com/duckdb/duckdb-web/issues/4037 and https://github.com/duckdb/duckdb/issues/14927


szarnyasg:
Thanks for the correction. https://github.com/duckdb/duckdb/issues/14927 should work but it's going to work in v1.2.0 the earliest (and there are likely other flags that only work with CSV/Parquet).

Issue #4124, opened 2024-11-21
Documentation enhancement


hellower:
```
################################################################################
# duckdb
################################################################################

__STAGE__$ duckdb
v1.1.3 19864453f7
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D select version();
┌─────────────┐
│ "version"() │
│   varchar   │
├─────────────┤
│ v1.1.3      │
└─────────────┘
D select current_date, current_time, current_timestamp;
┌──────────────┬──────────────┬────────────────────────────┐
│ current_date │ current_time │     current_timestamp      │
│     date     │     time     │  timestamp with time zone  │
├──────────────┼──────────────┼────────────────────────────┤
│ 2024-11-21   │ 21:52:56.387 │ 2024-11-22 06:52:56.387+09 │
└──────────────┴──────────────┴────────────────────────────┘

local date => 2024-11-22 
current_date  => 2024-11-21 (not 22)
current_time => UTC time !!
```

It would be appropriate to specify in the documentation that current_date and current_time, unlike in PostgreSQL, always display UTC-based values and are returned without timezone information.

```
################################################################################
# postgresql
################################################################################

postgresql=# select current_date, current_time, current_timestamp;
 current_date |    current_time    |       current_timestamp       
--------------+--------------------+-------------------------------
 2024-11-22   | 07:09:30.667281+09 | 2024-11-22 07:09:30.667281+09
(1 row)

postgresql=# 
```

PR #4206, opened 2024-11-30
Update postgresql_compatibility.md

soerenwolfers:
Fixes https://github.com/duckdb/duckdb-web/issues/4124  but hoping this will be undone in https://github.com/duckdb/duckdb/issues/15069


szarnyasg:
Thanks. Why is this draft though? I think this applies to v1.1 and we can undo it for v1.2 when the eventual fix for https://github.com/duckdb/duckdb/issues/15069 lands.

soerenwolfers:
True, let's go then.

PR #4208, opened 2024-11-30
Update map.md

soerenwolfers:
Make alias handlign consistent with other duckdb docs pages and consistent within the table itself.

Long-term I'd prefer a different alias handling though: Repeating the same text snippet 6 times is tedious and error-prone.


szarnyasg:
Thanks. Indeed this is tedious. We are working on generating the function descriptions from the DuckDB codebase (see e.g., https://github.com/duckdb/duckdb/pull/14838), which will eliminate some of the redundancy.

PR #4209, opened 2024-11-30
Update timestamp.md

soerenwolfers:
Solves https://github.com/duckdb/duckdb-web/issues/4131 by moving functions to the correct place.

Performance difference shouldn't really be surprising to anyone who (1) reads that the functions are in different pages and therefore produce different types (2) has gained some understanding of the difference between TIMESTAMP and TIMESTAMPTZ from the linked blog post. 

Btw, I'd be in favor of binning (hoho) the distinction between ICU provided functions and built-in functions, given that (1) the ICU extension ships with DuckDB by default (2) I wasn't able to figure out how to even disable it.


szarnyasg:
Thanks! I agree. I think starting new DuckDB codebase that deals with datetime values without timestamps is bad practice in most of the time – except for the select few (who have special use cases / really know what they are doing / don't care), most people should use timezoned timestamps.

PR #4220, opened 2024-12-01
Update star.md

soerenwolfers:
Fixes https://github.com/duckdb/duckdb-web/issues/4002 and fixes https://github.com/duckdb/duckdb-web/issues/4004

Also:

- Remove discussion of `count(*)` from the `star expression` page -- I don't think they're related in any way, and I think the no-star `count()` aggregate is better and less confusing anyway. 
- Remove the leading list of examples. The page was already unwieldy, and with the new additions it'd just have been a mess
- Restructure the page a bit with the goal to first discuss all `SELECT * <MODIFIER>` variants and then basically discuss the `COLUMNS` expression as a super-powered version of that.
- Update the railroad to describe an optional choice between EXCLUDE/RENAME/REPLACE/(LIKE/SIMILAR TO/GLOB) modifiers. Note that previously, this offered a sequence of [optional-EXCLUDE and optional-REPLACE], but now the permitted combinations are a bit odd: no combination works EXCEPT for `EXCLUDE + <ANYOTHER>` (actually, RENAME + LIKE works but produces a wrong result, I'll open a ticket about that). Therefore, I made the railroad diagram look like *no* combination is allowed, which I think makes sense until  the support has been sanitized. Also, note that `COLUMNS(* LIKE/SIMILARTO/GLOB)` is not supported but the diagram will look like it is because it offers the same star-options for both branches, same to how it is described in text. I think this should simply be supported rather than painstakingly editing the docs to list this special behavior.

I don't know how to preview the diagram changes (github.dev doesn't help), let me know if there is an easy way!


szarnyasg:
Thanks for the changes!

> I don't know how to preview the diagram changes (github.dev doesn't help), let me know if there is an easy way!

You have to build locally. There's a Docker image for it but it is slow and the JS gets cached easily – I had open a private browsing session to see the new railroad diagram:

Screenshot 2024-12-01 at 21 07 56

It looks great!

The reason I'm parking this for a few days is because this is a v1.2 change. I'll try to set up a subpage for the nightly version, so we can proceed with merging these.

soerenwolfers:
Updated the defaults in the new bits to be the empty lines

PR #4221, opened 2024-12-01
Update aggregates.md

soerenwolfers:
Yet another stab at quantiles, this time not correcting anything but improving by adding intuition. Lemme know if you think it's too long.


szarnyasg:
This is great, thanks!

Issue #2791, opened 2024-04-28
Inconsistency in the Casting Operation Matrix?


vlulla:
How should I read the table/figure listed at <https://duckdb.org/docs/sql/data_types/typecasting#casting-operations-matrix>? It appears that reading the cell at R1C1 the rows ought to be read as `from` type and the columns ought to be read as `to` type. But that does not make sense, at least to me. For instance, typing `select '10' > 9` in duckdb will print `true` indicating that the left operand has been _implicitly typecast_ from `varchar` to `int`. So, shouldn't the `varchar` column be transposed? I.e., all values _from_ `varchar` _can be implicitly_ converted to other types! By the way, this implicit conversion of string literal to _any_ other type is mentioned explicitly at <https://duckdb.org/docs/sql/data_types/literal_types#string-literals>. I find this figure a little confusing and am wondering if it cannot be fixed up to be more precise. Regardless, thanks for all the good work on duckdb documentation!

soerenwolfers:
More casting quirks that should be fixed or documented in this page:
https://github.com/duckdb/duckdb/issues/11958
https://github.com/duckdb/duckdb/issues/13847

PR #4222, opened 2024-12-01
Update typecasting.md

soerenwolfers:
Fixes https://github.com/duckdb/duckdb/issues/13847 (sadly).

From what I can tell, `BOOL -> INTEGERTYPES` is the only example of this special type of casting?

Maybe one should get one final confirmation whether it's worth documenting this special type of cast rather than simply dropping it and gaining more [PostgreSQL compatibility]( https://duckdb.org/docs/sql/dialect/postgresql_compatibility.html) for it (just to reiterate: I'm all for  implicit casts to programatically selected parent types, I just question the value of allowing *more* implicit casts in this situation than in other implicit cast situations.) 

By the way: do you have the HTML for the typecasting matrix?  I'd like to resolve https://github.com/duckdb/duckdb-web/issues/2791 too by adding `INTEGER_LITERAL` / `STRING_LITERAL` / `NUMERIC_LITERAL` to the table (at which point I'd also add the "combination cast" from bool to int to the table)

### DIFF
--- a/docs/sql/data_types/typecasting.md
+++ b/docs/sql/data_types/typecasting.md
@@ -12,11 +12,17 @@ Explicit typecasting is performed by using a `CAST` expression. For example, `CA
 
 ## Implicit Casting
 
-In many situations, the system will add casts by itself. This is called *implicit* casting. This happens for example when a function is called with an argument that does not match the type of the function, but can be casted to the desired type.
+When a function is called with an argument that does not match the type of the function, but can be casted to the desired type, the system will add an  *implicit* cast.
+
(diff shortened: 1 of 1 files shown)


szarnyasg:
IIRC the typecasting matrix was a Google Sheets sheet. I wanted to re-do it in Markdown as three column table (from, to, what type of cast is allowed) but did not get to this so far. Would you be interested in picking this up?

soerenwolfers:
Yeah, can do. Long-format is more in the spirit of SQL too.

soerenwolfers:
Although... are you sure you want a table with 576 entries in the docs? Maybe should that table only include positive entries? That'd be around 100.

szarnyasg:
Yeah, that sounds reasonable.

szarnyasg:
Sure. I reviewed the text and it's great, merging now.

PR #4238, opened 2024-12-03
Clarify Rust shutdown documentation

ianthetechie:
I noticed a bit of funky grammar here, so I took a moment to improve it.


szarnyasg:
Thanks!

PR #4250, opened 2024-12-03
Fix LIKE query at where.md

apachaves:
The LIKE operator was written as ILIKE in the example.


apachaves:
Indeed. When I read "Select all rows that match the given case-insensitive LIKE expression:" I assumed there was only one operator (the LIKE) and that it was case-insensitive.

Also, if the idea is to showcase both operators, maybe there might be separate examples for each? Just to try to be clear.

szarnyasg:
Sure! I'll add the example in PR #4252.

PR #4253, opened 2024-12-03
small design fixes

Franz-Kafka:
- footer z-index fix
- better colors for button borders in menu bar
- link to media with trailing slash (other media link)


szarnyasg:
thanks!

PR #4270, opened 2024-12-05
Prefer user selected theme over system theme

### LINKED ISSUE #4269: Website does not remember user selected theme
I wanted to browse DuckDB docs in the light theme but it kept switching back to the dark theme when navigating to other pages.

Steps to reproduce:
- Set your computer/browser to have a dark color scheme
- Go to a DuckDB documentation page (such as [FAQ](https://duckdb.org/faq))
- Click on theme toggle in the top right corner to switch to the light theme
- Website successfully switches to light theme
- Navigate to another page in the documentation or refresh the current page
- Website switches back to dark theme

pranavmangal:
This PR fixes #4269 

It seems the core issue is this piece of code present in the various HTML templates in the [_layouts](https://github.com/duckdb/duckdb-web/tree/main/_layouts) folder:
```html
<script>
	var userPrefersDark = localStorage.getItem('mode') === 'dark';
	var systemPrefersDark = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches;
	if (userPrefersDark || systemPrefersDark) {
		document.documentElement.classList.add('darkmode');
		document.documentElement.classList.add('disable-transitions');
	}
</script>
```

It checks if the user prefers the dark theme, and even if they don't (with the light theme set explicitly), it stills ends up using the dark theme if the system is set to the dark color scheme. 

I have updated this bit of code to check the user preference, use the dark/light theme if the preference is set and only fallback to the system color scheme if it doesn't exist. I have also extracted the snippet into a separate script to avoid duplication.

### DIFF
--- a/_layouts/community_extension_doc.html
+++ b/_layouts/community_extension_doc.html
@@ -23,15 +23,7 @@
 
 	{% seo title=false %}
 	
-	<script>
-		var userPrefersDark = localStorage.getItem('mode') === 'dark';
-		var systemPrefersDark = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches;
-		if (userPrefersDark || systemPrefersDark) {
-			document.documentElement.classList.add('darkmode');
-			document.documentElement.classList.add('disable-transitions');
-		}
-	</script>
-
+	<script src="{{ site.baseurl }}/js/determine_color_scheme.js?{{ site.data['hash'] }}"></script>
 </head>
 
 <body class="documentation communityextensions{% if page.body_class %} {{ page.body_class }} {% endif %}">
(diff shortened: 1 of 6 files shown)


carlopi:
This looks like it improve the detection, thanks!
But I think selecting the mode should not require an extra resource download.

There have to be ways to fight code duplication, I guess some generated file somewhere. For now I think it's ok to duplicate this everywhere, could you make this change?

Thanks

szarnyasg:
@Franz-Kafka can you please take a look at this PR?

szarnyasg:
@pranavmangal thanks a lot for reporting this and proposing a patch. We'll see how we can integrate it to our website.

PR #4295, opened 2024-12-07
docs: replace broken `SET GLOBAL` example

ajwerner:
Search path does not support being set globally:

```
D set global search_path = 'foo';
Catalog Error: option "search_path" cannot be set globally
```

See https://github.com/duckdb/duckdb/blob/aa3787b0/src/main/config.cpp#L154.


szarnyasg:
Thanks!

PR #4296, opened 2024-12-07
Open extension list in same window if navbar item is clicked

nicosuave:
Because I keep opening too many windows

(Random nit from my end @carlopi)


szarnyasg:
Fair enough, thanks!

PR #4302, opened 2024-12-08
Update aggregates.md

soerenwolfers:
Move null handling to before the tables. 

Also, remove the _"and groups without non-`NON` inputs"_ from the sentence _"All general aggregate functions except `count` return `NULL` on empty groups and groups without non-`NULL` inputs"_ since that's not true for the two null-respecting aggregates and their aliases, and I think it's already sufficiently clear without the addition that the sentence really means: 

> If the inputs to an aggregate function are empty, either because the group is empty to begin with or because all values have been filtered out either by a FILTER clause or because the aggregate function ignores NULLs, then the aggregate function returns NULLs"

(and I don't want to actually use that very verbose sentence in the docs)

### DIFF
--- a/docs/sql/functions/aggregates.md
+++ b/docs/sql/functions/aggregates.md
@@ -99,6 +99,16 @@ FROM tbl;
 |-----------|
 | 3, 2, 1   |
 
+
+### Handling `NULL` Values
+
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks!

PR #4317, opened 2024-12-09
Update pragmas.md

soerenwolfers:
setting both here looks like they do different things (when really they are aliases)


szarnyasg:
Thanks!

PR #4322, opened 2024-12-09
Fixed s3-parquet.sql Binder Error

gropaul:
Hi dear DuckDB team,

I discovered that the s3 query on the landing page returned a Binder Error, as one column name was incorrect. I am not 100% sure if my change affects the landing page query, but I could not fine the query elsewhere. 

Best regards,
Paul


szarnyasg:
Thanks!

PR #4352, opened 2024-12-12
add documentation for using zstd for compression for writes

gregorywaynepower:
* Adds examples of file extensions for writing with ZSTD for compression
* Wraps instance of gzip with "`"'s so the formatting is consistent among examples


szarnyasg:
Thanks!

PR #4361, opened 2024-12-13
Standardize Jekyll Server Startup when using Dev Containers method

pmatogo:
Updated the post-create command in `devcontainer.json` to start Jekyll server, aligning it with our standard Docker run method.

Here I am making it [identical to the docker run method](https://github.com/duckdb/duckdb-web/blob/main/scripts/docker-serve.sh#L21-L24).

The current implementation does not work anyway.


szarnyasg:
Thanks!

PR #4368, opened 2024-12-15
Update char.md

soerenwolfers:
Explain what happens with padding functions when original string has more than `count` characters.


szarnyasg:
Thanks!

PR #4414, opened 2024-12-18
updates for node neo

jraymakers:
Updated the sql-query and web-service-integration landing page examples to use Node Neo. Note that I changed the file type to `*.mjs` so modern syntax such as `import` and top-level `await` can be used.

I also updated the quick installation command to point to Node Neo.

Additionally, I updated the stable installation command for 1.1 to point to Node Neo. But I didn't update the nightly installation command, as I don't currently release a nightly version of Node Neo. Unfortunately this feels inconsistent. What do you think we should do about the nightly command? Can/should we remove it for Node? Or should I work on releasing a nightly version of Node Neo?

I included a couple minor updates to the Node Neo documentation:
- Added a link to the @duckdb/node-api NPM package. The documentation is copied from the README of that package, where it doesn't make sense to have a self-link, but the link makes sense here.
- Updated the supported platforms to include two recent additions: Linux arm64 and Mac OS X x64 (Intel).


jraymakers:
@szarnyasg

szarnyasg:
Hi @jraymakers, thanks for the changes! Our installation page has a note option, I think adding a note option for the nightly release (stating that this is for the old driver) is a good compromise for now. I'll add the note in a second.

data/installation-data-1.1.yml
```diff
@@ -302,6 +302,7 @@
   download_method: Package manager
   architecture: universal
   installation_code: npm install duckdb@next
+  note: The nightly release of the Node.js driver installs the old Node.js driver and not DuckDB Node Neo. For the Node Neo driver, the nightly release is currently not available. If you require this driver, consider compiling it from source.
```
jraymakers, line 305: Minor comment: Compiling the Node Neo driver from source won't get you a nightly version, because the build depends on released DuckDB binaries, and currently doesn't pull in nightly binaries.

It probably wouldn't be hard to add an option to pull in nightly binaries, but that configuration doesn't exist today.

szarnyasg: Thanks for the clarification! I'll just remove this part of the note then.

PR #4419, opened 2024-12-19
Update vss.md. add INSTALL and LOAD to example.

marhar:
The examples fail if the extension is not loaded.


szarnyasg:
Good point, thanks!

PR #4421, opened 2024-12-20
quick fix for headline issue

Franz-Kafka:
- will get back again to it after talking with Jonathan. But for now a quick fix.


szarnyasg:
Thanks!

PR #4433, opened 2024-12-24
Update node-neo API doc snippet with null chunk API change

Mike-Dax:
Chunks can be null as of this commit:

https://github.com/duckdb/duckdb-node-neo/commit/09d2dddf839fe753552bda3d10012f0a7fb343f2

This PR updates a snippet on the node neo page to match the API changes.


szarnyasg:
Thanks!

PR #4447, opened 2024-12-29
Quirks

soerenwolfers:
Some of my main pet peeves with duckdb for your perusal. Just the list for now, no integration with the website, because I wouldn't know where to put it.


szarnyasg:
Thanks, this is shaping up to be a great page! It should either go to the [dialect](https://duckdb.org/docs/sql/dialect/overview) or the [operations manual](https://duckdb.org/docs/operations_manual/overview). I think dialect is better because all of these quirks can be explained in terms of SQL expressions and there is some overlap with the Postgres compatibility page, which is also under the dialect section.

szarnyasg:
Hello, I did some editing by adding a table, some example output (in one case) and headers. Merging this now but I'm sure we'll iterate on this (and also move stuff between this, the Postgres and the Numerical pages).

PR #4463, opened 2025-01-02
Explain instants and binning

hawkfish:
The blog post is very old and it makes sense to move some of it into the main documentation.

(Thanks to @soerenwolfers for lighting a fire under me...)


szarnyasg:
Thanks!

PR #4469, opened 2025-01-04
Timestamps, again

soerenwolfers:
Second stab at the timestamp documentation, post hawkfish's overhaul.  I tried to follow other systems' / people's terminology as best I could, and I apologize if I misunderstood or misrepresented anything or anybody.

**DIFFS:**

1: Change blurb from 

> A timestamp specifies a combination of a date (year, month, day) and a time (hour, minute, second, microsecond or nanosecond)

 to 

> Timestamps represent points in time.
 
The old version (a) sounds like a timestamp is a sixtuple and puts too much focus on those somewhat arbitrary bins (b) unnecssarily limits the description to two of the four available precisions.

2: Similarly, in first paragraph

- change "Timestamps represent points in absolute time, usually called instants. DuckDB represents instants as the number of microseconds (µs) (or nanoseconds, for `TIMESTAMP_NS`) since [the epoch]" to "Timestamps represent points in time" (with definition of instant deferred to later)
- change "specifies a combination of DATE and TIME" to "combine DATE and TIME information"

3: Change table entries from 
 - "timestamp with X precision (UTC)" to "naive timestamp with X precision" and
- "timestamp with X precision (time-zone aware)" to "time zone aware timestamp with X precision"

4: Change "with 'instant semantics'" to "with `WITH TIME ZONE` semantics" in parquet warning box 

5: Delete an excessive intro example (made up for by more contextually embedded examples later on, see 7.), and modify another one to exhibit more functionality 

6: Add two paragraphs after the examples in the "Timestamp types" section, to 
-  explain the "naive" vs "time zone aware distinction" from the table 
-  clarify what the data types actually store (`INT64` in both varieties) and what not (time zone identifiers in neither), which I think is important for, and surprising to, e.g., pandas & polars users
- provide possible interpretations of the `TIMESTAMP` type (while describing `TIMESTAMPTZ` as "unambiguously identifies a point, or instant, in absolute time") 
- Add forward-references to the the `#Instants` and `#Binning` sections, for a better-integrated experience 

7: Add a section on the available conversions between the various available formats. I think the examples in this new section do a better job than any text could to eliminate any remaining vagueness in reader's minds about what the different types can be used for and how they interact

8: Cleanup "Special Values" section

9: Remove the statement "databases store instants" which I don't think is true in this generality and which in any case I don't think adds much to this definition of instants

10: Minor fixes to the `#Binning` section

11: Delete section on naive timestamps, since (a) it clashes with labeling the `TIMESTAMP` type as "naive timestamp" and (b) I included all the content in earlier sections.

**THOUGHTS & REFERENCES ON TERMINOLOGY**

*Instant & absolute & Timestamp (UTC)*
I agree that an "instant" is a point in *absolute* time, as described in ISO 8601's

> point on the [time axis] = [mathematical representation of the succession in time according to the space-time model of instantaneous events along a unique axis]

but wile the docs currently seems to take the stance that every timestamp type represents an instant, I don't think that's obviously/always/unambiguously so. 

Rather than explicitly saying that `TIMESTAMP WITHOUT TIME ZONE`s are or are not instants, I propose with this PR to only ever call `TIMESTAMPTZ` instants and leave it to readers how they want to interpret `TIMESTAMP WITHOUT TIME ZONE` values. I think this agrees with terminology used in other systems.

In this vein, I also removed the `UTC` label from the `WITHOUT TIME ZONE` types. In fact, that one I believe could be quite misleading: Speaking from the perspective of a hypothetical user in `+01:00` after seeing "TIMESTAMP: timestamp with nanosecond precision (UTC)" in the current version of the docs:

>  Shouldn't `(TIMESTAMP '2020-01-01 00:00:00')::TIMESTAMPTZ` print `2020-01-01 01:00:00+01:00` then? 
I start with a UTC timestamp, `TIMESTAMP '2020-01-01 00:00:00'`, and I ask for the instant in absolute time that this timestamp identifies; clearly the answer should be '2020-01-01 01:00:00+01:00', not `2020-01-01 00:00:00+01:00'!

I guess the `(UTC)` label is meant to indicate "follows UTC manipulation rules" but I don't think that's how people will read it. The `(UTC)` label also clashes with how other systems describe / label their equivalent timestamp types. 

*Naive timestamps*

I interpret the previous paragraph "naive timestamps" as defining these as something like

> an incomplete attempt to describe a point in absolute time, missing an unambiguous offset / reference point / time zone specification 

I agree with that; however, I view the storage of such naive timestamps in the `TIMESTAMP` type as "proper" use of the `TIMESTAMP` type (to the extent that you have to deal with this situation to begin with!), whereas I believe the current docs imply this is bad practice / shoehorning / type abuse?

This is all a bit vague and open to interpretation (obviously, I don't better know how things work / were intended to be thought about than the actual code author), my point is just: I think a common (and non-disprovable) way to use and think about  `TIMESTAMP` is to treat them as sixtuple of integer fields that can be manipulated using "nominal temporal logic" without actually assigning absolute temporal meaning to the result, but rather relying  on other data, e.g., a string time zone column in the same table, to provide absolute anchoring (God this is bad vague language). I think it's therefore misleading to attach "UTC" as a label to `TIMESTAMP` and I think it's fine to call `TIMESTAMP`s naive. Since I'm only using the word "naive" in the table and in a section name, it'd not be a great loss to swap "naive" for another label, e.g., "simple" or back to the original "ignores time zone"; however, it does seem most other systems use the naive / aware pair in a similar fashion as in this PR, which is why my preference is to keep it.

**SOURCES**

- **Rust/Chrono** use `NaiveDateTime` for what I believe is their equivalent to DuckDB's `TIMESTAMP` and uses phrasing like

> Converts the local NaiveDateTime to the timezone-aware DateTime.

similar to the (naive, time zone aware) label pair proposed in this PR 
https://docs.rs/chrono/latest/chrono/trait.TimeZone.html#method.from_local_datetime https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDateTime.html

- **Polars** describes the no-time-zone-stored version of its `datetime`, which behaves like DuckDB's `TIMESTAMP` as 

> no time zone, also known as "time zone naive"

 https://docs.pola.rs/user-guide/transformations/time-series/timezones/

- **Apache Hive** describes DuckDB's (really, PostgreSQL's) `TIMESTAMP` as matching Java's `LocalDateTime` (but I already gave up on using that label for `TIMESTAMP`) and describes DuckDB's `TIMESTAMPTZ` as matching Java's `Instant` https://hive.apache.org/docs/latest/different-timestamp-types_103091503/

- **Python** uses the "naive" vs "aware" pair of labels in much the same way I intend to use it with this PR: 

> With sufficient knowledge of applicable algorithmic and political time adjustments, such as time zone and daylight saving time information, an aware object can locate itself relative to other aware objects. 

> A naive object does not contain enough information to unambiguously locate itself relative to other date/time objects. Whether a naive object represents Coordinated Universal Time (UTC), local time, or time in some other time zone is purely up to the program [...]

 https://docs.python.org/3/library/datetime.html#aware-and-naive-objects

- **Snowflake** uses the word "wallclock" (I like that, but didn't make use of it) for the description of their `NO TIME ZONE` type:

> TIMESTAMP_NTZ internally stores “wallclock” time with a specified precision. All operations are performed without taking any time zone into account.

- **PostgreSQL** wiki also uses "wallclock", although more indirectly https://wiki.postgresql.org/wiki/PostgreSQL_vs_SQL_Standard

- The **.NET PostgreSQL** bindings (npgsql) describe PostgreSQL-proper's timestamp types as follows:

> In PostgreSQL, timestamp with time zone represents a UTC timestamp, while timestamp without time zone represents a local or unspecified time zone. 
https://www.npgsql.org/doc/types/datetime.html 

- **PostgreSQL's** own type description itself is rather sparse but says 

> For timestamp with time zone, the internally stored value is always in UTC

which at least hints by omission that they don't consider  "For `timestamp withOUT time zone` the internally stored value is always in UTC" to be a true statement.  

The description of the `epoch` function says: 

> For timestamp with time zone values, the number of seconds since 1970-01-01 00:00:00 UTC (negative for timestamps before that); for date and timestamp values, the nominal number of seconds since 1970-01-01 00:00:00, without regard to timezone or daylight-savings rules; 

Note that this uses the "UTC" suffix for `timestamp with time zone` but instead uses the word "nominal" (which I also used at some point in the PR and this description) and no timezone offset for `timestamp` (which confirms my optional removal of that offset in the epoch description [here](https://github.com/duckdb/duckdb-web/pull/4469/files#diff-8cf520b606223336e61ca76e6eac343c73b8fedfeebbf20afbbd04d44ca2839aR110).

They also use the "timezone-aware" label in  "All timezone-aware dates and times are stored internally in UTC. " 

https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-INPUT-TIME-STAMPS  https://www.postgresql.org/docs/current/functions-datetime.html

- **Apache Parquet** distinguishes "Instant semantics (timestamps normalized to UTC)" from "Local semantics (timestamps not normalized to UTC)" Being a pure storage format, it is fair to assume that "semantics" here refers to what stored values represent, i.e., instants in absolute time vs local wallclocks in unspecified time zones, not the semantics of how timestamp operations operate on these values. https://github.com/apache/parquet-format/blob/master/LogicalTypes.md


docs/sql/data_types/timestamp.md
```diff
@@ -71,27 +59,59 @@ SELECT TIMESTAMPTZ '1992-09-20 11:30:00.123456789';
...
+
+The corresponding `TIMESTAMP WITHOUT TIME ZONE` stores the same single `INT64`, but arithmetic, binning, and string formatting follow the straightforward rules of UTC without offsets or time zones. Accordingly, `TIMESTAMP`s could be interpreted as UTC timestamps, but more commonly they are used to represent *local* observations of time recorded in an unspecified time zone, and operations on these types can be interpreted as mechanical manipulation of record fields.
```
soerenwolfers, line 75:
```suggestion
The corresponding `TIMESTAMP WITHOUT TIME ZONE` stores the same single `INT64`, but arithmetic, binning, and string formatting follow the straightforward rules of UTC without offsets or time zones. Accordingly, `TIMESTAMP`s could be interpreted as UTC timestamps, but more commonly they are used to represent *local* observations of time recorded in an unspecified time zone, and operations on these types can be interpreted as mere manipulation of tuple fields following nominal temporal logic.
```

szarnyasg (approved):
This is all great, thanks! I only made very minor edits. Special thanks for doing all the research on what other systems / data formats use!

PR #4483, opened 2025-01-06
union all by name blog post

Alex-Monahan:
Hi folks! Open to feedback! I tried to keep this one a bit lighter to brighten things up a bit. Happy to tone it down though! I'll also happily fix any formatting issues that CI highlights.


szarnyasg:
I added some edits. This is good to go in from my side!

Alex-Monahan:
I added some benchmarking around the case where we do not use `union_by_name`. 

I removed the t2.micro results because the small tweak to the benchmark to use union_by_name ended up making performance dramatically slower (It's still running 15+ minutes later...). 

Good to go from my side, but if you see any tweaks that are needed, feel free!

PR #4495, opened 2025-01-07
Patch 83

soerenwolfers:
1. Remove trailing `000000` since the clients don't show it either.
2. Rephrase warning box since it is currently saying that the time component is always zero no matter the contents!
3. Downgrade warning to hint since I don't see the risk in getting wrong results from this (and it couldn't be any other way, since the column type can't depend on the contents of individual rows)

### DIFF
--- a/docs/sql/data_types/interval.md
+++ b/docs/sql/data_types/interval.md
@@ -86,7 +86,7 @@ SELECT
 ;
 ```
 
-> Adding an `INTERVAL` to a `DATE` returns a `TIMESTAMP` even when the `INTERVAL` has no microseconds component. The result is the same as if the `DATE` was first cast to a `TIMESTAMP` (which sets the time component to `00:00:00`) before adding the `INTERVAL`.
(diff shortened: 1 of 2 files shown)


docs/sql/data_types/interval.md
```diff
@@ -86,7 +86,7 @@ SELECT
 ;
 ```
 
-> Warning Addition or subtraction of an `INTERVAL` value with a `DATE` value will return a `TIMESTAMP` with the time component set to `00:00:00.000000`.
+> Adding an `INTERVAL` to a `DATE` returns a `TIMESTAMP` even when the `INTERVAL` has no microseconds component. The result is the same as if the `DATE` was first cast to a `TIMESTAMP` (which sets the time component to `00:00:00`) before adding the `INTERVAL`.
```
soerenwolfers, line 89:
```suggestion
> Adding an `INTERVAL` to a `DATE` returns a `TIMESTAMP` even when the `INTERVAL` has no microseconds component. The result is the same as if the `DATE` was cast to a `TIMESTAMP` (which sets the time component to `00:00:00`) before adding the `INTERVAL`.
```

szarnyasg:
These are great, thanks!

PR #4496, opened 2025-01-07
Update with.md with correct example for directed cycles

davidvlaminck:
1, 2 and 5 doesn't form a directed cycle
1, 8 and 5 does
or 1, 5 and 3.

### DIFF
--- a/docs/sql/query_syntax/with.md
+++ b/docs/sql/query_syntax/with.md
@@ -187,7 +187,7 @@ INSERT INTO edge VALUES
     (6, 3), (6, 4), (7, 4), (8, 1), (9, 4);
 ```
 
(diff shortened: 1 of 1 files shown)


docs/sql/query_syntax/with.md
```diff
@@ -187,7 +187,7 @@ INSERT INTO edge VALUES
     (6, 3), (6, 4), (7, 4), (8, 1), (9, 4);
 ```
 
-Note that the graph contains directed cycles, e.g., between nodes 1, 2, and 5.
+Note that the graph contains directed cycles, e.g., between nodes 1, 8, and 5.
```
szarnyasg, line 190:
```suggestion
Note that the graph contains directed cycles, e.g., between nodes 1, 5 and 8.
```

PR #4500, opened 2025-01-08
Struct

szarnyasg:


docs/sql/data_types/typecasting.md
```diff
@@ -99,18 +99,46 @@ Arrays follow the same casting rules as lists. In addition, arrays can be implic
 
 ### Structs
 
-Structs can be cast to other structs as long as the names of the child elements match.
+Structs can be cast to other structs as long as they share at least one child element.
```
soerenwolfers, line 102: "Share at least one child element" sounds like this is comparing entries instead of field names. Following the entry paragraph of the struct.md ("This document refers to those entry names as keys. "), the following wording would be more accurate: 

```suggestion
Structs can be cast to other structs as long as they share at least one key.
```

docs/sql/data_types/struct.md
```diff
@@ -189,7 +189,7 @@ CREATE TABLE t2 AS SELECT row('a');
 Invalid Input Error: A table cannot be created from an unnamed struct
 ```
 
-When casting structs, the names of fields have to match. Therefore, the following query will fail:
+When casting between structs, the names of at least one field have to match. Therefore, the following query will fail:
```
soerenwolfers, line 192: Could use the same sentence as at https://github.com/duckdb/duckdb-web/pull/4500/files#r1909106875

docs/sql/data_types/typecasting.md
```diff
@@ -99,18 +99,46 @@ Arrays follow the same casting rules as lists. In addition, arrays can be implic
...
+
+> The rationale behind this requirement is to help avoid unintended errors. If two structs do not have any elements in common, then the cast was likely not intended.
```
soerenwolfers, line 104: s.a.

docs/sql/data_types/typecasting.md
```diff
@@ -99,18 +99,46 @@ Arrays follow the same casting rules as lists. In addition, arrays can be implic
...
 
 ```sql
 SELECT CAST({'a': 42} AS STRUCT(a VARCHAR));
 ```
 
+Child elements that exist in the target struct, but that do not exist in the source struct, default to `NULL`.
```
soerenwolfers, line 110: s.a.

PR #4510, opened 2025-01-08
Update securing_extensions.md: change install to load

carlopi:
Checks are performed at load time, I think it's somewhat confusing to say install.


szarnyasg:
Thanks!

PR #4516, opened 2025-01-09
Add description on DuckDB release process for community-extensions

carlopi:
Toward addressing https://github.com/duckdb/community-extensions/issues/223


carlopi:
I think does makes sense to have somewhere in the docs.

Likely needs to be iterated on, but this is a starting point.

szarnyasg:
Thanks, I did some edits. This is good to go.

PR #4528, opened 2025-01-09
Link todos.json

ruslandoga:
Similar to the CSV overview page: 
- https://duckdb.org/docs/data/csv/overview
- https://github.com/duckdb/duckdb-web/blob/3719e1c6bff5c8cffadc2ccb96ceeb5c6592e4bb/docs/data/csv/overview.md?plain=1#L10

### DIFF
--- a/docs/data/json/overview.md
+++ b/docs/data/json/overview.md
@@ -25,6 +25,8 @@ While it is not a very efficient format for tabular data, it is very commonly us
 
 ## Examples
 
+The following examples use the [`todos.json`](/data/todos.json) file.
+
 ### Loading JSON
 
 Read a JSON file from disk, auto-infer options:


ruslandoga:
Ah, no. Not all queries seem to be using that file. The

```sql
SELECT *
FROM read_json('todos.json',
               format = 'array',
               columns = {userId: 'UBIGINT',
                          id: 'UBIGINT',
                          title: 'VARCHAR',
                          completed: 'BOOLEAN'});
```
breaks with 

> Invalid Input Error: Expected top-level JSON array with format='array', but first character is '{' [...]
>
> Try setting format='auto' or format='newline_delimited'.

ruslandoga:
Ah, the file is not https://duckdb.org/data/todos.json but https://duckdb.org/data/json/todos.json :)

PR #4534, opened 2025-01-09
Use read_json_objects in read_json_objects docs

ruslandoga:
`read_json` results in a different output and is discussed later in the document


szarnyasg:
Thanks!

PR #4535, opened 2025-01-09
Add field_appearance_threshold type

ruslandoga:
Right now it's missing from the table: https://duckdb.org/docs/data/json/loading_json#parameters-1


szarnyasg:
Thanks!

PR #4596, opened 2025-01-17
Improve CSV parameter descriptions

soerenwolfers:
- Remove bloat words "Option to [...]", "Enables [...]", "This option [...]", "This option allows you to specify the [...]", "Specifies the [...]", "Whether or not to [...]", "Set the [...]", "Boolean value that specifies whether or not [...]", "If this option is enabled, [...]", "If set to true, [...]", "It is important to note that [...]"
- Fix and add links to "rejects_scan" and "rejects_table" descriptions
- Fix `filename` description to say that it returns absolute or relative paths depending on the input
- Consistently use `line` to refer to lines of a CSV file, and `row` to refer to records in a DuckDB table
- Mention that `encoding` is not supported by the `COPY` statement
- More precision in `comment` description
- Separate out some aliases that are supported by writer but not reader.
- Downgrade encoding limitation section to `Tip` box, now that DuckDB does support three encodings


szarnyasg:
This is great, thanks a lot!

szarnyasg (approved):

PR #4598, opened 2025-01-17
docs: remove the extra redundant single quote

nekomeowww:
## Summary

Simple change to remove the extra `'`.


szarnyasg:
Thanks!

PR #4669, opened 2025-01-29
Add docs for VARINT

Tishj:
This PR fixes <https://github.com/duckdb/duckdb-web/issues/4521>

I've cherry-picked some interesting information from the PR, focusing on the storage aspect, as that's the most relevant difference between the other integer types (that and the incredibly high amount of digits of course)

Because of that, I've also added `Size in Bytes` to the existing table of integer types.

### DIFF
--- a/docs/sql/data_types/numeric.md
+++ b/docs/sql/data_types/numeric.md
@@ -11,20 +11,31 @@ The types `UTINYINT`, `USMALLINT`, `UINTEGER`, `UBIGINT` and `UHUGEINT` store wh
 
 <div class="center_aligned_header_table"></div>
 
-| Name        | Aliases                          |                                      Min |                                     Max |
-| :---------- | :------------------------------- | ---------------------------------------: | --------------------------------------: |
-| `TINYINT`   | `INT1`                           |                                     -128 |                                     127 |
-| `SMALLINT`  | `INT2`, `INT16` `SHORT`          |                                   -32768 |                                   32767 |
-| `INTEGER`   | `INT4`, `INT32`, `INT`, `SIGNED` |                              -2147483648 |                              2147483647 |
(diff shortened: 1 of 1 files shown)


docs/sql/data_types/numeric.md
```diff
@@ -11,21 +11,32 @@ The types `UTINYINT`, `USMALLINT`, `UINTEGER`, `UBIGINT` and `UHUGEINT` store wh
...
+| `BIGINT`    | `INT8`, `INT64` `LONG`           |                     -9223372036854775808 |                     9223372036854775807 |                 8 |
+| `HUGEINT`   | `INT128`                         | -170141183460469231731687303715884105728 | 170141183460469231731687303715884105727 |                16 |
+| `UTINYINT`  | -                                |                                        0 |                                     255 |                 1 |
+| `USMALLINT` | -                                |                                        0 |                                   65535 |                 2 |
+| `UINTEGER`  | -                                |                                        0 |                              4294967295 |                 4 |
+| `UBIGINT`   | -                                |                                        0 |                    18446744073709551615 |                 8 |
+| `UHUGEINT`  | -                                |                                        0 | 340282366920938463463374607431768211455 |                16 |
```
pdet, line 25: I guess we could also add VARINT here?

Tishj: I don't think that's a good idea.
The minimum and maximum are both incredibly huge, and the Size in Bytes is variable, which is exactly why I've made a separate section for it.

I do understand why it would fit there, because it *just* a numeric like the others in the list, but it's giant

pdet: Well, that's a summary, on top of the page, and most likely the only thing that 99.9% of the people will read from it, so that's why I think is important to add it there. The ranges might look "confusing" at first glance, but then readers can check the actual varint description you added, if it's not immediately clear to them.

I think for bytes you can have the range 4 - 2ˆ23 bytes. Min and max can be a value * 10^...

I can also do that in a later PR.

PR #4675, opened 2025-01-30
Update installation.js

Franz-Kafka:
- adjusted download-method logic to keep valid options enabled
- moved universal-architecture check to occur after filtering by download method


szarnyasg:
Thanks!

PR #4692, opened 2025-02-04
Fixup troubleshooting redirect

carlopi:
More an issue than a PR.

I am not sure even this works, question would be, how to make:
https://duckdb.org/docs/extensions/troubleshooting/?version=061967287d&platform=osx_arm64&extension=iamnotexisting
work.


szarnyasg:
Yes, I noticed that `/some/path/` doesn't get the same treatment as `/some/path`, so the redirects are littered with cases where both of these are spelled out. We don't have an .htaccess file and don't want to use JS for this, so this seems to be the best way.

PR #4741, opened 2025-02-09
ATTACH support for table macros is available in 1.2

buremba:
> duckdb test.duckdb

```sql
CREATE MACRO static_table() AS TABLE
    SELECT 'Hello' AS column1, 'World' AS column2;
```

> duckdb

```
attach 'test.duckdb' as test;
select * from test.static_table();
```

The above works as intended.


szarnyasg:
Thanks!

PR #4743, opened 2025-02-10
Fix new functions detection query

hatamiarash7:
There is a little problem with this query in the [community extensions documentation](https://duckdb.org/community_extensions/documentation)

```sql
SELECT * FROM functions_pre EXCEPT (FROM functions_post) ORDER BY ...;
```

This query does not return new functions because the table names are misspelled. The correct query is:

```sql
SELECT * FROM functions_post EXCEPT (FROM functions_pre) ORDER BY ...;
```

Check https://github.com/duckdb/community-extensions/issues/290


carlopi:
Thanks! Documentation was off, but more relevantly this allowed to uncover a change in duckdb/duckdb that needed addressing in the function discovery code.

PR #4773, opened 2025-02-11
highlight table rows on hover

Franz-Kafka:
added css for highlighting table rows on hover in documentation like in blog posts


szarnyasg:
Thanks!

PR #4794, opened 2025-02-13
Add ST_Transform() example that uses an NTv2 grid file

tbbuck:
Hi there,

This PR adds an example use case for spatial function ST_Transform(). If someone wants to transform the x, y coordinates of British National Grid to map display WGS84, they might assume that the regular call to ST_Transform(some_geometry, 'EPSG:27700', 'EPSG:4326') would do it. And it will get them most of the way - but with noticeable error.

By highlighting the option to specify an NTv2 grid file, anyone who doesn't know that this might be an issue now has a reason to find out IF it's an issue. And anyone who's been there before will appreciate a quick copy/paste snippet.

All the best
Tom.

### DIFF
--- a/docs/extensions/spatial/functions.md
+++ b/docs/extensions/spatial/functions.md
@@ -1903,6 +1903,28 @@ SELECT ST_AsText(
 POINT (544615.0239773799 6867874.103539125)
 ```
 
+#### Example: Improve accuracy by using a local NTv2 grid file 
+```sql
+-- Transform a geometry from OSG36 British National Grid EPSG:27700 to EPSG:4326 WGS84
+-- Standard transform is often fine for the first few decimal places before being wrong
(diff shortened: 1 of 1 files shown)


szarnyasg:
@tbbuck Thanks for the contribution! This is generated code, so it should be part of the spatial repository. What do you think @Maxxen?

tbbuck:
Oh! Oh no! I can't believe I missed that, sorry 🤦‍♀️ Thanks for pointing out @szarnyasg!

@Maxxen if you think it's worth including, I'm more than happy to create a fresh pull request in the spatial repo.

szarnyasg:
@tbbuck no problem, I moved the example to the spatial repo.

tbbuck:
Many thanks @szarnyasg - much appreciated :)

PR #4834, opened 2025-02-18
Update 2025-02-10-window-catchup.md

mwhitaker:
Add missing INTERVAL in SQL query from the docs https://duckdb.org/docs/sql/functions/window_functions.html#range-framing

Using v1.2.0 5f5512b827.


szarnyasg:
Indeed, the query throws a syntax error in the absence of the `INTERVAL` terms. Thanks!

PR #4838, opened 2025-02-19
Fix CORE_EXTENSIONS in build config.

xevix:
This matches with what's in https://github.com/duckdb/duckdb-web/blob/main/docs/dev/building/building_extensions.md#building-extensions and should work as expected. I tested locally just in case.


szarnyasg:
Thanks!

PR #4846, opened 2025-02-19
Julia: Link to API docs generated on JuliaHub

frankier:
Currently the documentation for the Julia connectors is a bit limited, but there are quite good docstrings. These are available from within Julia's help system, but also build into online docs automatically on JuliaHub This PR adds a link.


szarnyasg:
This looks great, thanks!

PR #4871, opened 2025-02-22
123456 is a confusing example when formatted as hex 1e240

jgrg:
`1e240` looks like a floating point number. Switch to 654321 which is `9fbf1` in hex.


szarnyasg:
Indeed, thanks for the enhancement!

PR #4903, opened 2025-02-26
Added file loading chapter on relative paths to import overview

ojeker:
Before this modification, it was hard to find out how relative paths are expanded in duckdb.

The configuration setting `file_search_path` is a "god sent" when globbing files from a directory tree and needing to reference/harmonize the root of this tree over team members and CD/CI. The added chapter should help others to find this valuable setting more easily than me :).


szarnyasg:
This a great addition, thanks!

PR #4930, opened 2025-02-27
Restructure directories

szarnyasg:
This PR changes the documentation's directory structure.

## Unified URLs

Previously, we had two types of directories (with examples):

* stable version: `/docs/sql/indexes`
* archive versions: `/docs/archive/0.8/sql/indexes`

For the most part, this was unwieldy to handle in our scripts (Liquid templates, Python, JS). In the new structure, the stable version has the `/stable/` string in the URL, while the archive versions only have the version number.

* stable version: `/docs/stable/sql/indexes`
* archive versions: `/docs/0.8/sql/indexes`

This pretty much follows Postgres' convention (see e.g., https://www.postgresql.org/docs/curent/index.html and https://www.postgresql.org/docs/15/index.html)

## Single Installation Page

I removed old installation pages as they were difficult to maintain and a bit chaotic: if one went back to the page of, say, v0.10, the options offered were v0.10 or nightly. Going forward, it may be better to have a version selector that's part of the installation page.

## Redirects

I added redirects to keep existing links working.

## Future Work

Some scripts in the `scripts` need further maintenance.


szarnyasg:
The linter errors are expected as basically every single documentation page changed, and some don't pass the linter.

PR #4954, opened 2025-03-04
cross reference specifiers %G and %V for function strptime

c-herrewijn:
Combining specifiers %V and %Y gives unexpected results. (always January first)
```
SELECT strptime('2024-39', '%Y-%V');
->
2024-01-01 00:00:00
```
This is because `%V` should be combined with `%G` and not with `%Y`.

Python warns for this, so I guess it would be nice for us to at least explicitly mention it in the docs.
```python
from datetime import datetime

datetime.strptime('2024-37-1', '%Y-%V-%u')
# ValueError: ISO week directive '%V' is incompatible with the year directive '%Y'. Use the ISO year '%G' instead.
```


szarnyasg:
Thanks!

PR #4970, opened 2025-03-05
C++ version

szarnyasg:


dentiny:
This is how the markdown renders on my end, is it expected? I thought they should be clickable links
image

szarnyasg:
Hi @dentiny, we use Jekyll's link tags: https://jekyllrb.com/docs/liquid/tags/#links
When we built the HTML site, Jekyll checks these links, and if they exist, it substitutes them with the full links.

dentiny:
> Hi @dentiny, we use Jekyll's link tags: https://jekyllrb.com/docs/liquid/tags/#links When we built the HTML site, Jekyll checks these links, and if they exist, it substitutes them with the full links.

Thank you for the explanation!

PR #4974, opened 2025-03-06
Add SET/DROP NOT NULL to ALTER COLUMN syntax diagram

JelteF:
Ugh, 1 second before you merged it I realized I had updated the 0.8 javascript file instead of the stable one...


szarnyasg:
Ah, I see. No problem!

PR #4978, opened 2025-03-06
`sql_path` in `information_schema.schemata` applies to a feature not available in DuckDB

Mytherin:
This came up here - https://github.com/duckdb/duckdb/pull/16419

I'm not entirely sure what this field should contain, but I would say it's not the file system location of the database at least. Postgres says `Applies to a feature not available in PostgreSQL` so safe to mirror that.


szarnyasg:
Thanks!

xevix:
I tried to find it in the SQL standard, but buying this part of the ISO standard costs about 221 CHF and I'm not sure it's that important 😄 . 

I've found various answers online saying it's similar to the `search_path` in PostgreSQL, but specific to the given schema, but this definition seems questionably helpful.

Thanks for clarifying the docs 👍

PR #5011, opened 2025-03-10
fix typo in enum.md

yunyu950908:
Noticed a typo here while I'm reading the docs https://duckdb.org/docs/stable/sql/data_types/enum#enum-definition


szarnyasg:
Thanks!

PR #5024, opened 2025-03-11
Update script.js

Franz-Kafka:
Update the external link filter so that links containing images are not marked as external.


szarnyasg:
Thanks!

PR #5042, opened 2025-03-14
duckdb R package 1.2.1 released

krlmlr:
Today.

Have you considered configuring Renovate or Dependabot to track the versions?


szarnyasg:
@krlmlr not yet, but we probably should. Thanks for tracking this manually for v1.2.1!

PR #5047, opened 2025-03-15
Update create_macro.md

soerenwolfers:
Ideally, there'd be some common page for all `CREATE` statements in the future, but until then it can't harm to have this here. There is a group of people out there (including myself) that almost never create tables (because i'm using DuckDB purely for transient analytic work on dataframes in Python) but do create functions.


szarnyasg:
Thanks!

PR #5066, opened 2025-03-18
Iceberg extension: Remove allow_moved_paths argument from S3 example

floscha:
Running the example as is results in an `Invalid Input Error: Enabling allow_moved_paths is not enabled for directly scanning metadata files.`

Omitting the argument, as the error suggests, makes the example work.


szarnyasg:
Thanks!

PR #5080, opened 2025-03-20
Correct table name in query in "DuckDB Tricks - Part 2".

derekmahar:
In section "Repeated Data Transformation Steps" in DuckDB Blog entry "DuckDB Tricks - Part 2", replace table name "schedule_cleaned" with "schedule".


szarnyasg:
This was indeed off. Thanks for the PR!

Issue #3490, opened 2024-08-28
The last modified field shows the same day for all pages


szarnyasg:
The `Last modified` field in the footer shows the day of the last deployment.

szarnyasg:
The documentation is updated quite frequently with most pages updated at least once a month. So a 'last modified' field wouldn't add much value.

PR #5090, opened 2025-03-22
"Modified at" footer element

szarnyasg:


szarnyasg:
This PR does not yet resolve issue #3490 but it removes two unused Jekyll plugins. The jekyll-github-metadata plugin made the build slower and triggered GitHub's abuse detection mechanism as it flooded GH with queries.

PR #5136, opened 2025-03-24
updated colors & small fixes

Franz-Kafka:
- updated color variables
- updated warning box
- updated installation code to handle line breaks
- removed heading for normal note boxes


szarnyasg:
Thanks!

PR #5142, opened 2025-03-24
`duckdb.default_connection` changed from attribute to function in duckdb 1.2

kylebarron:
`duckdb.default_connection` was changed from attribute to function in duckdb 1.2.


szarnyasg:
Thanks!

Issue #4058, opened 2024-11-13
Issue found on page 'Python Function API'


skinkie:
The Python UDF interface is _extremely_ strong. But it does lack one very valuable example for example using functools.partial. I think you want to show the end-user the ability to _provide arguments_ directly from Python, for example, passing a context without using a global variable. The example below shows:

1. How to provide an actual function **and** call it within the UDF.
2. How to provide a predfined variable **and** use it within the UDF.
3. Notice that within, SQL can influence the UDF's output. 

```python
import functools
import duckdb
import time

def print_udf(func, arg1: str, arg2: str) -> str:
    return ' '.join([str(func()), arg1, arg2])

with duckdb.connect() as con:
    con.create_function('print', functools.partial(print_udf, time.time, 'hello'))
    con.execute("SELECT print('world');")
    print(con.fetchone()[0])
```

Maybe we need some discussion, therefore I did not create a pull request yet.

PR #5159, opened 2025-03-27
fix: add partial and null handling in Python function

acirtep:
Fixes https://github.com/duckdb/duckdb-web/issues/4154 and https://github.com/duckdb/duckdb-web/issues/4058


szarnyasg (approved):
Thanks!

PR #5217, opened 2025-04-10
Add llms.txt and llms-full.txt for /docs/stable

mehd-io:
Currently, we only have [`llms.txt`](https://duckdb.org/llms.txt) at the root, which is quite minimal.

This PR adds a Python script pipeline to keep `/docs/stable` always up to date:

- `llms.txt`: list of links
- `llms-full.txt`: full content of the linked pages

**Note:**  
@szarnyasg — I noticed you pointed to the `raw.githubusercontent.com` version of the markdown URL on the root `llms.txt`. I have mixed thoughts on this. While [llmstxt.org](https://llmstxt.org) recommends using markdown-friendly URLs, relying on the raw GitHub link could reduce traffic through duckdb.org and hurt its domain authority.

As a workaround, I’ve included both the duckdb.org link and the raw markdown URL for now. Curious to hear your thoughts, @tdoehmen.

### DIFF
--- a/.github/workflows/build-stable-documentation.yml
+++ b/.github/workflows/build-stable-documentation.yml
@@ -16,6 +16,8 @@ jobs:
           ruby-version: '3.3' # Not needed with a .ruby-version file
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
           cache-version: 0 # Increment this number if you need to re-download cached gems
+      - name: Generate llms.txt files
+        run: python scripts/generate_llms_txt.py
       - name: Setup Pages
         id: pages
         uses: actions/configure-pages@v5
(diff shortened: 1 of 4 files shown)


tdoehmen:
The llms.txt proposal is not really clear on this. I'd lean a bit more towards linking to the DuckDB website instead of DuckDB website + github.raw. The examples I've seen either link to websites or .md files, but haven't seen any doing both. If traffic reduction is a concern why not host the .md files on the duckdb domain?

szarnyasg:
Thanks, this is great. I agree that hosting the .md files on duckdb.org seems to be the right way to go. I'll implement the required changes.

PR #5228, opened 2025-04-14
general improvements

Franz-Kafka:
- fixed issue with inline highlighted text
- set max-width for videos in posts
- fixed firefox menu issues
- changed colors for "Preview" flag


szarnyasg:
Thanks!

PR #5264, opened 2025-04-18
Add page for encodings extension

### LINKED ISSUE #5263: duckdb-encodings needs documentation
cc @pdet

szarnyasg:
Fixes #5263

We need to write about which functions are supported (e.g., `read_json`?)

To be merged when the extension becomes available

cc @pdet

### DIFF
--- a/_data/menu_docs_preview.json
+++ b/_data/menu_docs_preview.json
@@ -1031,6 +1031,10 @@
               "url": "delta"
             },
             {
+              "page": "Encodings",
+              "url": "encodings"
+            },
+            {
               "page": "Excel",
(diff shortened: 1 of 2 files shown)


pdet:
This LGTM!
We currently only support encodings when reading CSV files.

I think adding them to other scanners (or connectors) should be relatively simple, but I confess I'm not sure how prevalent encodings are in other data formats.

My guess is that JSON probably doesn't really suffer from this, since it was designed in the 2000s for web stuff and has a more sane standard.

szarnyasg:
Thanks, I clarified that

PR #5266, opened 2025-04-18
Revert "Avoid overriding extension list"

carlopi:
This reverts commit 04f9d24686f6dde869591b82fc43e44ed8efdf8a.


carlopi:
I would say it's probably good, most extensions are back (I think)

szarnyasg:
excellent, thank you!

szarnyasg (approved):

PR #5283, opened 2025-04-22
small design adjustments

Franz-Kafka:
- more padding when preview or outdated banner is shown
- toc menu now auto height


szarnyasg:
Thanks!

PR #5290, opened 2025-04-23
Compact function tables

c-herrewijn:
This PR is a continuation of the functionality introduced by https://github.com/duckdb/duckdb-web/pull/5017 and https://github.com/duckdb/duckdb-web/pull/5238

### changes
- Compressed function tables.
If a function has optional arguments AND `duckdb_functions()` has the same function description for the different function_overloads there will be a single entry for these functions. Optional arguments will be in brackets: `[, ...]`
- improved examples with multiple spaces
consquitve spaces are removed by HTML rendering; symbol `␣` is used to represent a space, e.g. functions, `trim` and `bar`

### example: 
`parse_dirname(path[, separator])`
|   |   |
|---|---|
| __Description__ | Returns the top-level directory name from the given `path`. `separator` options: `system`, `both_slash` (default), `forward_slash`, `backslash`. |
| __Example__ | `parse_dirname('path/to/file.csv', 'system')` |
|__Result__ | `path` |
 
### scope
Implemented for pages `blob.md` and `char.md`, but the python script is generic and could also work for other doc pages with sql function tables.


szarnyasg:
Great, thanks!

Issue #5246, opened 2025-04-16
Add in the Sphinx generated API docs the reference to example pages


acirtep:
Where available, add in https://duckdb.org/docs/stable/clients/python/reference/ links to the example pages.

PR #5293, opened 2025-04-24
fix: add relation rst

acirtep:
Fixes: https://github.com/duckdb/duckdb-web/issues/5246

**Breaking change** for the current stable docs, since they do not have the new relational API page.


docs/preview/clients/python/reference/index.md
```diff
@@ -43,9 +17,9 @@ title: Python Client API
 </dd>
 </dl>
 
-<dl class="py exception">
```
acirtep, line 46: @szarnyasg how important is to display `exception` instead of `class`?

szarnyasg: I am fine both ways, this seems to be a minor detail

szarnyasg (approved):

PR #5298, opened 2025-04-24
Add notes on IN/NOT IN to nulls.md, and simplify prose

NickCrews:
I wanted to use simpler language than eg `semantics`.


szarnyasg:
Thanks!

PR #5303, opened 2025-04-27
Fix repeat words

szarnyasg:
@c-herrewijn in the description of the `md5` function, a repeat word "as" occurs:

```
Returns the MD5 hash of the `blob` as as a `VARCHAR`.
```

I'm not 100% where it is coming from in the DuckDB codebase, would you mind taking a look?

### DIFF
--- a/docs/preview/data/json/loading_json.md
+++ b/docs/preview/data/json/loading_json.md
@@ -348,7 +348,7 @@ CREATE TABLE numbers AS
 
 | Name | Description | Type | Default |
 |:--|:-----|:-|:-|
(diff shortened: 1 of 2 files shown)


c-herrewijn:
@szarnyasg The PR (https://github.com/duckdb/duckdb/pull/17132) is not yet merged on duckdb side. 
I just updated the PR to include this fix: https://github.com/duckdb/duckdb/pull/17132/commits/99e8be4e037b53db339514ed0aa03311601f08fd

szarnyasg:
Got it, thanks for the fix!

PR #5307, opened 2025-04-28
Fixed typo in the ST_Read example section

dbaclin:
Looks some of the keywords were truncated at some point
ELECT -> SELECT
REATE -> CREATE


szarnyasg:
@dbaclin thanks!

PR #5360, opened 2025-05-12
Preview: TPCH no longer packaged in Python

carlopi:
After https://github.com/duckdb/duckdb/pull/15923 this should no longer be an exception.

R I am not sure if also there they changed settings on extensions they ship.


szarnyasg:
Thanks!

PR #5311, opened 2025-04-28
Update lambda page with new syntax

szarnyasg:
Fixes #5294

cc @taniabogatsch 

Pending on PR https://github.com/duckdb/duckdb/pull/17235

PR #5362, opened 2025-05-13
Lambda docs with deprecation notice

taniabogatsch:
Follow-up of https://github.com/duckdb/duckdb-web/pull/5311.


szarnyasg:
Thanks!

Issue #5373, opened 2025-05-14
document parameters for functions


acirtep:
document method parameters for https://duckdb.org/docs/preview/clients/python/relational_api#functions

PR #5381, opened 2025-05-15
fix: add param desc for function Python rel API

acirtep:
Fixes https://github.com/duckdb/duckdb-web/issues/5373


szarnyasg:
Thanks!

PR #5394, opened 2025-05-19
docs: comprehensive update to Python package development docs

evertlammerts:
This commit updates the Python package development documentation to provide complete guidance on building from source, setting up development environments, and troubleshooting common issues. Includes new sections for debugging, IDEs and building with extensions, and moves snippets previously in the Python package README.

Related to duckdb/duckdb#17483

### DIFF
--- a/docs/stable/dev/building/python.md
+++ b/docs/stable/dev/building/python.md
@@ -7,14 +7,14 @@ title: Python
 
 The DuckDB Python package lives in the main [DuckDB source on Github](https://github.com/duckdb/duckdb/) under the `/tools/pythonpkg/` folder. It uses [pybind11](https://pybind11.readthedocs.io/en/stable/) to create Python bindings with DuckDB.
 
-# Prerequisites
+## Prerequisites
 
 For everything described on this page we make the following assumptions:
 
 1. You have a working copy of the duckdb source (including the git tags) and you run commands from the root of the source
 2. You have a suitable Python installation available in a dedicated virtual env
 
-## 1. DuckDB code
+### 1. DuckDB code
 
 Make sure you have checked out the [DuckDB source](https://github.com/duckdb/duckdb/) and that you are in its root. E.g.:
 
@@ -38,7 +38,7 @@ git fetch --tags upstream
 git push --tags
 ```
 
-## 2. Python Virtual Env
+### 2. Python Virtual Env
 
 For everything described here you will need a suitable Python installation. While you technically might be able to use your system Python, we **strongly** recommend you use a Python virtual environment. A virtual environment isolates dependencies and, depending on the tooling you use, gives you control over which Python interpreter you use. This way you don't pollute your system-wide Python with the different packages you need for your projects.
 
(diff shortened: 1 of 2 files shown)


docs/stable/dev/building/python.md
```diff
@@ -5,17 +5,291 @@ redirect_from:
...
+#### Cloud Storage
+
+You may need the package files to reside under the same prefix where the library is installed; e.g., when installing to cloud storage from a notebook.
+
+First, get the repository based version number and extract the source distribution.
+
+```bash
+python3 -m pip install build # required for PEP 517 compliant source dists
+cd tools/pythonpkg
+export SETUPTOOLS_SCM_PRETEND_VERSION=$(python3 -m setuptools_scm)
```
Tishj, line 100: I'm not entirely sure about this, as I haven't tested the use of `SETUPTOOLS_SCM_PRETEND_VERSION` myself.
`OVERRIDE_GIT_DESCRIBE` is more battle-tested I feel, but only by a little. The `setup.py` has both, and also `MAIN_BRANCH_VERSIONING` has impact on the generated version (v1.3.0 vs 1.2.3 for example).

Tishj: To explain this in a bit more detail:
when creating the sdist, we use the amalgamation to collect the files, then inside `src/function/table/version/pragma_version.cpp` we add defines for the:
- `DUCKDB_SOURCE_ID`
- `DUCKDB_VERSION`
- `DUCKDB_MAJOR_VERSION`
- `DUCKDB_MINOR_VERSION`
- `DUCKDB_PATCH_VERSION`

This is what's used to create the version of duckdb that gets installed.
Through the `use_scm_version` keyword argument in the `setup()` call we make sure the python package versioning matches this version.

I don't know how much of this we want to add to the docs here, but I figured I'd add some context around this

Tishj (changes requested):
Thanks, it makes a lot of sense for this to live in the docs repo rather than the `tools/pythonpkg/README.md`

I have some comments about the `SETUPTOOLS_SCM_PRETEND_VERSION` section, but that's the only thing that feels off / underspecified to me.

docs/stable/dev/building/python.md
evertlammerts: This is a section I've copied over verbatim [from the README](https://github.com/duckdb/duckdb/pull/17483/files#diff-565fa0ef6a3eabd7c23791fb3cd6dc41c0b7505d9cabd48cbf7aad44496abab1L54). It's from a 5+ year old commit (by @arjenpdevries :-)) and hasn't really been touched since then, by the looks of it. I'm not sure what exactly this installation option is and whether it's still needed... I can also just delete it from the docs if it's no longer relevant?

szarnyasg:
@evertlammerts great work! I pushed a bunch of changes and merged with `main` to make the linter happy.

I recommend using `batch` code blocks most of the time. The `bash` code block adds the prompt `$` to the line.

docs/stable/dev/building/python.md
evertlammerts: I deleted this section. It's not working anymore anyway. If we do at some point need instructions for this sort of install then we can revisit and add them.

Tishj (approved):
Thanks!

szarnyasg (approved):

PR #5404, opened 2025-05-21
Use Python standard case-style for lambda

Mytherin:


Mytherin:
This seems to be wrong in many more places actually - I will do a pass

PR #5419, opened 2025-05-22
httpfs: clarify that credentials should be made available for AWS client

simon0191:
Clarify that the credentials for GCS and R2 should be made available as AWS credentials.


szarnyasg:
Thanks!

PR #5434, opened 2025-05-24
Fix lambda syntax

szarnyasg:


szarnyasg:
Discovered while working on #5435

PR #5459, opened 2025-05-31
docs/data/csv: auto_type_candidates

Djfe:
I split both changes so 09901dd88b58b334d0edad42f2c9eb08dca29c3a can be applied later to docs folder 1.2

Both commits should be applied to folder stable as well.

I'm planning on correcting the order of https://github.com/duckdb/duckdb/blob/dcf0e1c8936d74be48fd1cc0309638117b43aa47/src/include/duckdb/execution/operator/csv_scanner/csv_reader_options.hpp#L82-L86 so it matches the actual specificity defined here: https://github.com/duckdb/duckdb/blob/dcf0e1c8936d74be48fd1cc0309638117b43aa47/src/execution/operator/csv_scanner/util/csv_reader_options.cpp#L523-L530 .

Disclaimer: This was verified using the Python API and SQL commands.
I could not find the code that maps NULL to SQLNULL. The DuckDB code base uses SQLNULL and TIMESTAMP_TZ instead of NULL and TIMESTAMPTZ like the external SQL API.

### DIFF
--- a/docs/preview/data/csv/overview.md
+++ b/docs/preview/data/csv/overview.md
@@ -130,7 +130,7 @@ Usage example:
 SELECT * FROM read_csv('csv_file.csv', auto_type_candidates = ['BIGINT', 'DATE']);
 ```
 
-The default value for the `auto_type_candidates` option is `['SQLNULL', 'BOOLEAN', 'BIGINT', 'DOUBLE', 'TIME', 'DATE', 'TIMESTAMP', 'VARCHAR']`.
+The default value for the `auto_type_candidates` option is `['NULL', 'BOOLEAN', 'TIME', 'DATE', 'TIMESTAMP', 'BIGINT', 'DOUBLE', 'VARCHAR']`.
 
 ## CSV Functions
 
--- a/docs/preview/data/csv/overview.md
+++ b/docs/preview/data/csv/overview.md
@@ -130,7 +130,7 @@ Usage example:
 SELECT * FROM read_csv('csv_file.csv', auto_type_candidates = ['BIGINT', 'DATE']);
 ```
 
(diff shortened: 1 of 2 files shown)


Djfe:
Right, I forgot to mention: `NULL` doesn't appear to make a difference to auto_type_candidates. Though adding it doesn't hurt either.
Even empty lines are of type `VARCHAR` for some reason (despite `SQLNULL` having the heighest weight of specificity).
And I cannot cast `VARCHAR` columns to type `NULL` which means I'm not even

docs/preview/data/csv/overview.md
```diff
@@ -130,7 +130,7 @@ Usage example:
 SELECT * FROM read_csv('csv_file.csv', auto_type_candidates = ['BIGINT', 'DATE']);
 ```
 
-The default value for the `auto_type_candidates` option is `['SQLNULL', 'BOOLEAN', 'BIGINT', 'DOUBLE', 'TIME', 'DATE', 'TIMESTAMP', 'VARCHAR']`.
+The default value for the `auto_type_candidates` option is `['NULL', 'BOOLEAN', 'TIME', 'DATE', 'TIMESTAMP', 'TIMESTAMPTZ', 'BIGINT', 'DOUBLE', 'VARCHAR']`.
```
pdet, line 133: The order doesn't really matter here, since it's internally sorted when the option is set. Still, this might make it more readable.

pdet (approved):
LGTM, I'm also fine with adding the info about the `DECIMAL` type.

Djfe:
How does autodetection of type `NULL` look like? Or does that rather concern single rows and not a full column?
Should we drop it from documentation or is it related to the read_csv parameter `nullstr`?

szarnyasg:
Sorry about dropping this, this should have been merged a long time ago...!

szarnyasg:
Thanks for the contribution!

PR #5477, opened 2025-06-03
small improvements

Franz-Kafka:
- updated blog post width to 750px
- updated copy button
- removed bullets from version selection


szarnyasg:
Thanks!

PR #5478, opened 2025-06-03
first version of improved logging docs

samansmink:
First version to remove outdated logging docs, this is to be updated for v1.4 which will have an improved UX to interact with the logging config


szarnyasg:
Thanks!

PR #5497, opened 2025-06-10
small enhancements

Franz-Kafka:
- removed lottie.js
- updated an image
- external links with "." at the end stay together
- open FAQ when link has hash


szarnyasg:
Thanks!

PR #5520, opened 2025-06-17
included new type system

Franz-Kafka:
- type styles are defined in mixins and variables
- updated most of the font sizes to the new system


szarnyasg:
Thanks!

PR #5534, opened 2025-06-20
explain that `use` statement uses an implicit 'main' schema if not provided explicitly

nvartolomei:
Related https://github.com/duckdb/duckdb-iceberg/issues/324


szarnyasg:
Thanks!

PR #5536, opened 2025-06-22
docs(adbc): remove confusing line and update the link to the arrow adbc docs

eitsupi:
Context: apache/arrow-adbc#2216

DuckDB actually supports ADBC API 1.1.0, but the documentation shows the old library version (different from the API version) as 0.7, which is confusing and needs to be updated.
And, the link will be updated to point to the latest version.


szarnyasg:
Great, thanks for the PR and the context!

PR #5541, opened 2025-06-24
Clarify arrow_lossless_conversion option

NickCrews:
Inspired by https://github.com/duckdb/duckdb/pull/17857#issuecomment-2999464024

In particular, the existing phrasing of "When a duckdb typedoes not have a clear native or canonical extension match in Arrow" makes it sound like this option should not affect UUID types, since there is a canonical extension type for uuid in arrow. I think the new phrasing is more accurate of when this option has an effect.

It now doesn't mention the fact that we generate our own `Duckdb.*` extension type, but I think that is fine. Before, it wasn't really enough info to actually be useful anyways. If we did want to mention this, I can adjust the PR.

The result is significantly longer than the description for the other options, but I would rather have precision and it be fully explained rather than brevity.


szarnyasg:
Thanks! This is auto-generated so I turned it into a PR to the main codebase.

szarnyasg:
@NickCrews I PR'd this to the main codebase. We try to keep the codebase self-contained, so unfortunately I could not include the link to the discussion in the pull request.

PR #5562, opened 2025-07-07
add secret type 'ducklake'

c-herrewijn:
Added Secret Type `ducklake`. 
Note that I used an external link to https://ducklake.select/docs/stable/duckdb/usage/connecting#secrets.
The [ducklake extension page](https://duckdb.org/docs/stable/core_extensions/ducklake), is not as detailed at the moment, and doesn't cover DuckLake secrets, atm


szarnyasg:
Thanks!

PR #5568, opened 2025-07-09
chore(docs): fix tyop

DaxServer:


szarnyasg:
Thanks! +1 point for the commit message!

PR #5569, opened 2025-07-09
Document USE schema.

xevix:
The `USE` command actually permits just a `schema` to be passed in as well. This seems to be [expected behavior](https://github.com/duckdb/duckdb/pull/18179#issuecomment-3054060500).


szarnyasg:
Thanks!

PR #5577, opened 2025-07-13
Add PHP client page to stable and preview

dhernandez:
PHP client basic docs


szarnyasg:
Thanks for the contribution!

PR #5579, opened 2025-07-14
Fix Typo in Known Python Issues

alanleungcn:
https://duckdb.org/docs/stable/clients/python/known_issues.html#crashes-and-errors-on-windos

Windows was mistyped as Windos.


carlopi:
Thanks!

PR #5597, opened 2025-07-21
Change .NET library link

Giorgi:
Since there are several packages, this link will point to all of them. 

@szarnyasg Or is better to link to https://duckdb.net/ ?


szarnyasg:
This is great, thanks!

PR #5604, opened 2025-07-24
CLI WSL Memory Note

Ian-Fogelman:
Hi DuckDB!

This PR addresses https://github.com/duckdb/duckdb/issues/17646, there was not a super obvious place to put this note so I simply added it to the overview page with the noted workaround solution. Unclear if its worth mentioning this caveat for the small user base using WSL and the CLI concurrently, but since the issue was still open and marked for docs I wanted to give it a shot.

### DIFF
--- a/_data/menu_docs_preview.json
+++ b/_data/menu_docs_preview.json
@@ -277,6 +277,10 @@
                 {
                   "page": "Syntax Highlighting",
                   "url": "syntax_highlighting"
+                },
+                {
+                  "page": "Known Issues",
+                  "url": "known_issues"
                 }
               ]
             },
--- a/_data/menu_docs_stable.json
+++ b/_data/menu_docs_stable.json
@@ -277,6 +277,10 @@
                 {
                   "page": "Syntax Highlighting",
                   "url": "syntax_highlighting"
+                },
+                {
+                  "page": "Known Issues",
+                  "url": "known_issues"
                 }
               ]
             },
--- a/docs/preview/clients/cli/known_issues.md
+++ b/docs/preview/clients/cli/known_issues.md
@@ -1,0 +1,26 @@
+---
+layout: docu
+title: Known Issues
+---
+
+## Incorrect Memory Values on WSL 2
+
+On Windows Subsystem for Linux 2 (WSL2), when querying the `max_memory` or `memory_limit` from the `duckdb_settings`, the values may be inaccurate on certain Ubuntu versions (e.g., 20.04 and 24.04):
+
+Example:
+
+```sql
+FROM duckdb_settings() WHERE name LIKE '%mem%';
+```
+
+The output contains values larger than 1000 PiB:
+
+```text
+┌──────────────┬────────────┬─────────────────────────────────────────────┬────────────┬─────────┐
+│     name     │   value    │                 description                 │ input_type │  scope  │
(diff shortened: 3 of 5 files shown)


szarnyasg:
Hi @Ian-Fogelman thanks for the PR. A question: is this limited to the CLI client or does this also occur on other clients such as Python?

Ian-Fogelman:
Hi @szarnyasg, I was able to test this on a previous version of WSL, version `2.5.4`. I was not able to reproduce in the DuckDB driver, but also I was not able to reproduce when using the CLI on WSL on ubuntu `22.04.05`, perhaps the issue is tied specifically to ubuntu `24.04.2` as the original reporter was using this version :

`testduckdb.py`:

```
import duckdb

# Connect to an in-memory DuckDB database
con = duckdb.connect()

# Execute the query to select settings with 'mem' in the name
result = con.execute("SELECT * FROM duckdb_settings() WHERE name LIKE '%mem%';").fetchall()

# Print the results
for row in result:
    print(row)
```

Output:

```
ian@DESKTOP-6L1B155:~$ python3 testduckdb.py
('max_memory', '15.6 GiB', 'The maximum memory of the system (e.g. 1GB)', 'VARCHAR', 'GLOBAL')
('memory_limit', '15.6 GiB', 'The maximum memory of the system (e.g. 1GB)', 'VARCHAR', 'GLOBAL')
```


DuckDB  CLI output:

```
ian@DESKTOP-6L1B155:~$ /home/ian/.duckdb/cli/latest/duckdb
DuckDB v1.3.2 (Ossivalis) 0b83e5d2f6
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D from duckdb_settings() where name like '%mem%';
┌──────────────┬──────────┬─────────────────────────────────────────────┬────────────┬─────────┐
│     name     │  value   │                 description                 │ input_type │  scope  │
│   varchar    │ varchar  │                   varchar                   │  varchar   │ varchar │
├──────────────┼──────────┼─────────────────────────────────────────────┼────────────┼─────────┤
│ max_memory   │ 15.6 GiB │ The maximum memory of the system (e.g. 1GB) │ VARCHAR    │ GLOBAL  │
│ memory_limit │ 15.6 GiB │ The maximum memory of the system (e.g. 1GB) │ VARCHAR    │ GLOBAL  │
└──────────────┴──────────┴─────────────────────────────────────────────┴────────────┴─────────┘
```
image

Ian-Fogelman:
Hi @szarnyasg,

Just an update I was able to test this on Canonical-Ubuntu-20.04-2025.07.23-0 on a remote server, and the error did occur, but only on the CLI. By the way this was on the latest duckdb CLI `v1.3.2 (Ossivalis) 0b83e5d2f6`.

This makes me think its tied to specific versions of Ubuntu, I think a followup action for me could be:

-  A github action to test the CLI and Python drivers for various distros of Ubuntu may be required to 
- Update the note in the PR with specific versions based on the test results.

Do you agree this would be worth testing to confirm and merge a PR to docs?


image

hannes:
hey, thanks for adding this. The text should probably state prominently that this is a WSL-specific issue. As I see it at the moment, it looks like this happens on all Ubuntu.

szarnyasg:
Thanks for the contribution! I simplified the wording a bit, added a note for WSL and moved it to a separate ”known issues” page.

PR #5637, opened 2025-08-08
Update editing.md with external editor section

rashkov:
Feature was implemented as part of this PR: https://github.com/duckdb/duckdb/pull/11447/

There is some current documentation of it inside of `.help` and `.help .edit` but having it in these docs seems very helpful

That PR limits this feature to linenoise compatible CLI editors. It seems like it will support windows when linenoise comes to the windows duckdb CLI.


szarnyasg:
Thanks!

PR #5644, opened 2025-08-11
feat: add ml data preprocessing post

acirtep:


_posts/2025-08-15-ml-data-preprocessing.md
```diff
@@ -0,0 +1,539 @@
...
+        "std_\0": stddev_pop(columns(column_list));
+```
+
+In the above macro definition:
+
+- any table can be provided as input parameter and queried due to [`query_table`]({% link docs/stable/guides/sql_features/query_and_query_table_functions.md %});
+- we apply aggregate functions over the list of columns provided as input parameter, by using [column expressions]({% link docs/stable/sql/expressions/star.md %}#columns-expression);
+- we generate the aggregate alias by using the original column name via the [`\0` in the alias definition]({% link docs/stable/sql/expressions/star.md %}#renaming-columns-with-regular-expressions-in-a-columns-expression).
+
+We can now calculate the standard scaling as follows:
```
guillesd, line 308: I am not sure that this is more readable than the original query, but I guess the idea here is to show that DuckDB can do this as well?

_posts/2025-08-15-ml-data-preprocessing.md
```diff
@@ -0,0 +1,539 @@
...
+    );
+```
+
+### Robust Scaling
+
+_Robust scaling_ is a data normalization technique that transforms numerical features by subtracting the median and dividing by the interquartile range (IQR). Unlike standard scaling, which uses the mean and standard deviation, robust scaling reduces the influence of outliers by focusing on the middle 50% of the data. This makes it well-suited for datasets with skewed distributions or extreme values.
+
+In DuckDB, we can calculate the quantile ranges with the [`quantile_cont` statistical aggregate]({% link docs/stable/sql/functions/aggregates.md %}#quantile_contx-pos):
+
+```sql
```
guillesd, line 379: okey maybe now I get why: the idea is to have a single table macro that provides all of the metrics for different types of scalings so that it can be reused right? Maybe not very computationally friendly since this macro will calculate also unnecessary things for different scaling methods, so maybe worth to mention this

_posts/2025-08-15-ml-data-preprocessing.md
```diff
@@ -0,0 +1,539 @@
...
+
+> The above queries are non-deterministic, therefore a sort or storing the data in a reference table might be required for incremental processing.
+
+## Feature Scaling
+
+One other common data preprocessing step in machine learning is to scale numerical features, such that the values of different features are brought to a similar range or distribution. _Scaling,_ also known as _feature normalization_ or _standardization,_ involves transforming features so they have comparable magnitudes; typically by rescaling them to a fixed range (like 0 to 1) or adjusting them to have zero mean and unit variance. This process is required because many algorithms rely on distance calculations or gradient updates, which can be skewed if features vary widely in scale.
+
+If encoding is performed on the initial raw data (due to the need of knowing the entire categorical values list), scaling requires to split the data into training and testing data sets, in order to avoid [data leakage](https://scikit-learn.org/stable/common_pitfalls.html#data-leakage). In DuckDB, we can split the data by [sampling]({% link docs/stable/sql/samples.md %}) it:
+
+```sql
```
guillesd, line 247: so here looking at the benchmark I was surprised that we were doing worse than scikit learn. Maybe this is because they just slice arrays and that is pretty quick. However I did some digging and your query pattern is slower than an anti-join. You are doing something like this:
```sql
select * from range(10_000_000) except select * from range (2_000_000);
```
which according to my profiling is at least 2x slow than this:
```sql
select range a from range(10_000_000) where a not in (select range a from range(2_000_000));
```
`explain analyze` shows that the SET operation does a hash group and a hash join whereas the anti join is just a hash join!

guillesd: Actually I see that the problem is also that you do `set thread = 1` to make it reproducible, however you do not need to do this single thread for the other 20% (the `EXCEPT` clause).

_posts/2025-08-15-ml-data-preprocessing.md
acirtep: yes, the idea is to show that common transformations can be abstracted away in macros.

_posts/2025-08-15-ml-data-preprocessing.md
acirtep: In the benchmark itself (https://github.com/duckdb/duckdb-blog-examples/blob/main/ml_data_preprocessing/benchmark_duckdb.py#L107) it is using an anti join, I will move the thread setting just before it and update the repos.

_posts/2025-08-15-ml-data-preprocessing.md
acirtep: I have checked the execution plan and only the used aggregated functions are executed, therefore I have not added a note about the computation:

```
explain
select avg_velocity_score, q75_spending_deviation_score
    from scaling_params(
        'financial_trx_training',
        ['velocity_score', 'spending_deviation_score']
    );
 ```
 
 ```
 ┌───────────────────────────┐
│    UNGROUPED_AGGREGATE    │
│    ────────────────────   │
│        Aggregates:        │
│          avg(#0)          │
│     quantile_cont(#1)     │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│    ────────────────────   │
│       velocity_score      │
│  spending_deviation_score │
│                           │
│       ~4000000 Rows       │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         SEQ_SCAN          │
│    ────────────────────   │
│           Table:          │
│   financial_trx_training  │
│                           │
│   Type: Sequential Scan   │
│                           │
│        Projections:       │
│  spending_deviation_score │
│       velocity_score      │
│                           │
│       ~4000000 Rows       │
└───────────────────────────┘
```

_posts/2025-08-15-ml-data-preprocessing.md
guillesd: Maybe also worth mentioning why you are using `reservoir` sampling (because you want an exact amount of rows), but this is the most computational intensive of the sampling techniques.

That's it though, thanks for the updates!

acirtep: I have added a note about why `reservoir` is used.
Thank you for the review!

szarnyasg:
Thanks!

Issue #5634, opened 2025-08-06
Issue found on page 'Window Functions' lag function aliases instead of example


abogdanov85:
You need to fix it like this:
Aliases  `lag(column, 3, 0)` -> Example  `lag(column, 3, 0)`

Page URL: <https://duckdb.org/docs/stable/sql/functions/window_functions#lagexpr-offset-default-order-by-ordering-ignore-nulls>

PR #5646, opened 2025-08-11
Update window_functions.md

soerenwolfers:
fixes https://github.com/duckdb/duckdb-web/issues/5634

@szarnyasg Now that there are separate stable and dev pages, is it acceptable to only change the dev version? 
I'm doing almost all of my contributions by phone (and even on a laptop applying patches to different files intra-repository is not trivial in git) and having to port back changes would severely hamper the ability of users like me to contribute in this way.  I'm not suggesting that _you_ do the work of porting changes to stable, but rather that either 
- duckdb accept that only dev has the most up-to-date and correct version of the docs or 
- duckdb stops having separate stable and dev documentation (including going back to parking some PRs where deemed necessary)
- duckdb accepts having fewer hot-patch phone contributions (tbh, I'm not sure how much you get contributed via this mechanism anyway)


szarnyasg:
Hi @soerenwolfers,

> is it acceptable to only change the dev version?
> [...]
> * duckdb accept that only dev has the most up-to-date and correct version of the docs or

Yes, it's perfectly fine to only target the dev version!

To further elaborate on this:

– I found that porting back the changes from a laptop is quite simple, it sometimes only takes one minute per PR. This is also the reason why I didn't script it yet – it would certainly be possible to get a patch out of the diffs and replay it on the other directory with some small changes, but at the current volume of PRs, I could not yet justify the development effort.

– I think that late in the release cycle – like now when we are less than a month from the next version – putting a lot of effort into stable version has limited returns. The current preview version is going to be the stable soon, after which very few users will see the outdated version.

Thanks again for all your contributions!

PR #5654, opened 2025-08-13
Update instantiation.md by fixing typo in CDN wasm example

4www:
It seems that the `!` postfix was too much, as with it there are errors in the browser console and in js linter. Without it seems to work.

I am missing anything from vite/js/ts or other things that made this `!` necessary?


carlopi:
Thanks!

PR #5676, opened 2025-08-18
refactor(go-example): Update Go example on ADBC Client docs page

### LINKED ISSUE #5660: Update Go example on ADBC Client docs page to use high-level ingest API
Recently @Mandukhai-Alimaa [added a simpler, higher-level syntax for ingestion](https://github.com/apache/arrow-adbc/pull/3150) in the Go ADBC driver manager.

Should we update the example in the [Go section of the ADBC Client page in the DuckDB docs](https://duckdb.org/docs/stable/clients/adbc.html#go) use this simpler syntax?

@Mandukhai-Alimaa can you take a look at this and see if this would be reasonable to do?

The source file is at https://github.com/duckdb/duckdb-web/blob/main/docs/preview/clients/adbc.md#go.

Mandukhai-Alimaa:
Update Go example on ADBC Client docs page to use high-level ingest API. Closes #5660

### DIFF
--- a/docs/preview/clients/adbc.md
+++ b/docs/preview/clients/adbc.md
@@ -244,23 +244,10 @@ func (r *DuckDBSQLRunner) importRecord(sr io.Reader) error {
         return fmt.Errorf("failed to create IPC reader: %w", err)
     }
     defer rdr.Release()
-    stmt, err := r.conn.NewStatement()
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks!

PR #5677, opened 2025-08-19
nit wording on cli known issues

Ian-Fogelman:
FYI we have confirmed that this issue is not relegated to WSL or WSL / Ubuntu, but instead just specific Ubuntu versions. I was able to reproduce this behavior with a ssh session to a remote Ubuntu box (`Canonical-Ubuntu-20.04-2025.07.23-0`) this is out of LTS but still serves as example. Also see https://github.com/duckdb/duckdb/issues/18663. 

I [created a GHAction](https://github.com/Ian-Fogelman/duckdb-sandbox/blob/main/.github/workflows/test_cli.yml) to test the CLI command on multiple Ubuntu versions: 

- [24-04](https://github.com/Ian-Fogelman/duckdb-sandbox/actions/runs/17056637061/job/48355445415)
- [22-04](https://github.com/Ian-Fogelman/duckdb-sandbox/actions/runs/17056637061/job/48355445403)
- [20-04](https://github.com/Ian-Fogelman/duckdb-sandbox/actions/runs/17056637061/job/48355445401)

But version 20-04 has been running for over 1 hour.


szarnyasg:
Thanks for the detailed investigation!

Ubuntu 20.04 is [end of life](https://endoflife.date/ubuntu) for 99% of users (except for the ones who purchase extended maintenance from Canonical), so the really interesting one is 24.04.

In #5604, 24.04 was shown as returning incorrect results. In GitHub Actions this does not reproduce, so this may be an interplay of WSL an Ubuntu 24.04? I'm unsure.

Ian-Fogelman:
~~Yes I believe you are right @szarnyasg , I think it is not strictly Ubuntu version, and WSL is involved in reproducing, I did spin up an additional 24.04 Ubuntu server to test in addition to the GH Action results, I think this proves that WSL is involved. Will close this PR out thanks!~~

szarnyasg:
Thanks for reopening! I pushed a fix separate via commit 9b711bb17305f0771d1124d4f18dbb6cf8da7000. I also opened a PR to resolve this for DuckDB v1.3.

szarnyasg:
The PR landed in `main`, so the fix will be out for DuckDB v1.5

PR #5687, opened 2025-08-21
add merge into docs in preview

guillesd:


szarnyasg:
Great work! A few comments:

0) I added the keywords to our SQL highlighter (a fork of Rouge) at https://github.com/duckdb/rouge/blob/duckdb/lib/rouge/lexers/sql.rb, and bumped the dependency.

1) I made a few small tweaks and added extra ordering to ensure the results are deterministic.

2) If I run the queries, in sequence, I get one mismatch from the output presented in the PR:

```
MERGE INTO people
    USING (
        SELECT
            unnest([3, 2]) AS id, 
    ) AS deletes
    ON (deletes.id = people.id)
    WHEN MATCHED AND people.salary > 100_000.0 THEN DELETE;

FROM people
ORDER BY id;
```

returns

```
┌───────┬─────────┬──────────┐
│  id   │  name   │  salary  │
│ int32 │ varchar │  float   │
├───────┼─────────┼──────────┤
│     2 │ Anna    │ 100000.0 │
│     3 │ Sarah   │  95000.0 │
└───────┴─────────┴──────────┘
```

I would also add the output for the `UPDATE` and `DELETE` conditions example.

guillesd:
@szarnyasg I fixed some stuff in the examples and added the table for the final query!

PR #5701, opened 2025-08-22
fix: add description of completion time eta

rustyconover:
Hi @szarnyasg,

As requested here is a first start at describing the query completion ETA feature.

Cheers,

Rusty


rustyconover:
Should address #5673

szarnyasg:
Looks great, thanks!

Issue #5704, opened 2025-08-22
Issue found on page 'query and query_table Functions'


soerenwolfers:
doesn't describe `query` at all, only `query_table`

rpbouman:
(For completeness) 
The page is: 

https://duckdb.org/docs/stable/guides/sql_features/query_and_query_table_functions

Issue #5702, opened 2025-08-22
Issue found on page 'Utility Functions': query( query_string_literal ) can also be passed string expressions


rpbouman:
The documentation for [query(query_string_literal)](https://duckdb.org/docs/stable/sql/functions/utility#queryquery_string_literal) reads:

> Table function that parses and executes the query defined in query_string_literal. Only literal strings are allowed. Warning: this function allows invoking arbitrary queries, potentially altering the database state.

On v1.3.2 I found it also allows string expressions:

```
from query('select '||'1');
┌───────┐
│   1   │
│ int32 │
├───────┤
│   1   │
└───────┘
```

And functions returning strings:

```
set variable myquery = 'select 1';
from query( getvariable('myquery') )
┌───────┐
│   1   │
│ int32 │
├───────┤
│   1   │
└───────┘
```

Even user-defined functions returning strings:

```
create function q_count(str) as 'select count(*) from '||str;
select * from query( q_count('duckdb_settings()') );
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│     126      │
└──────────────┘
```

And, the kicker:

`select * from query( current_query() );`
(crashes)

TBC I think it's incredibly useful that query() accepts string expressions, in particular user-defined function calls, and getvariable(). Absolutely desired behavior! 
The documentation should be modified to explain this feature. It should probably also explain that it doesn't work when the query string is generated by a query:

```
select *
from query( (select 'select 1') );

Binder Error:
Table function cannot contain subqueries
```

but, that people my be able to work around it using variables:

```
set variable my_query = (select 'select 1');
select *
from query( getvariable('my_query') );
```

(As for passing current_query() as argument to query(), I guess it would be cool if that could be caught, but that is not a documentation issue. Interestingly, a similar scenario with query_table() results in infinite recursion being detected:

```
DuckDb> create or replace view my_view as select 1;
Run Time (s): real 0.001 user 0.000000 sys 0.000000
changes:   0   total_changes: 0
DuckDb> create or replace view my_view as select * from query_table('my_view');
Run Time (s): real 0.000 user 0.000000 sys 0.000000
changes:   0   total_changes: 0
DuckDb> select * from my_view;
Run Time (s): real 0.001 user 0.000000 sys 0.000000
Binder Error:
infinite recursion detected: attempting to recursively bind view "my_view"

```
perhaps that would also be possible for query())

[...]

PR #5708, opened 2025-08-24
Update query_and_query_table_functions.md

soerenwolfers:
fixes https://github.com/duckdb/duckdb-web/issues/5704 
fixes https://github.com/duckdb/duckdb-web/issues/5702

- include `query` description in page "`query` and `query_table` functions"
- change "string literal" to "constant string". One could be more technical by saying somethign like "bind time" constant, but I haven't found that level of detail elsewhere in the docs where there is talk about "constant parameters"


szarnyasg:
Thanks! A small change from me: I adjusted the function argument everywhere.

PR #5709, opened 2025-08-24
Update aggregates.md

soerenwolfers:
@szarnyasg While fixing this, I noticed that the alias handling here is a very inconsistent:

- There are tons of aliases that are in neither the short nor the long form table
- There are some aliases that are in the short but not the long form table
- Ther are tons of aliases where all alises are in both the short and the long form table


szarnyasg:
Thanks! I spun out an issue for the aliases.

PR #5710, opened 2025-08-24
Update list.md

soerenwolfers:
fixes https://github.com/duckdb/duckdb/discussions/18696

Please fact check the claim that `list_aggr` applies aggregaes in the order of the list with the team. I've based my claims on the experimental

```sql
CREATE OR REPLACE TABLE df1 AS (SELECT random() AS x FROM range(10_000_000));
CREATE OR REPLACE TABLE df2 AS (SELECT array_agg(x ORDER BY rowid) as x FROM df1);
SELECT
   list_aggr(x, 'sum') -- always the same result
FROM df2;
SELECT
    sum(x) -- different results on each execution
FROM df1;
```

### DIFF
--- a/docs/preview/sql/functions/list.md
+++ b/docs/preview/sql/functions/list.md
@@ -1085,7 +1085,8 @@ The function [`list_aggregate`](#list_aggregatelist-name) allows the execution o
 
 `list_aggregate` accepts additional arguments after the aggregate function name. These extra arguments are passed directly to the aggregate function, which serves as the second argument of `list_aggregate`.
 
-Order sensitive aggregate functions are applied in the order of the list. No `ORDER BY` / `DISTINCT` / `FILTER` clauses are supported by `list_aggregate`. They may instead be emulated using `list_sort` / `list_grade_up` / `list_select` / `list_distinct` / `list_filter`. 
+Order-sensitive aggregate functions are applied in the order of the list. The `ORDER BY`, `DISTINCT` and `FILTER` clauses are not supported by `list_aggregate`.
+They may instead be emulated using `list_sort`, `list_grade_up`, `list_select`, `list_distinct` and `list_filter`.
 
 ```sql
 SELECT list_aggregate([1, 2, -4, NULL], 'min');
--- a/_posts/2021-11-12-moving-holistic.md
+++ b/_posts/2021-11-12-moving-holistic.md
@@ -31,7 +31,7 @@ These are called _holistic_ aggregates, and they require more care when implemen
 For some aggregates (like `STRING_AGG`) the order of the values can change the result.
 This is not a problem for windowing because `OVER` clauses can specify an ordering,
 but in a `GROUP BY` clause, the values are unordered.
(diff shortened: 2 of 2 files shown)


taniabogatsch:
Taken from `extension/core_functions/scalar/list/list_aggregates.cpp`, `ListAggregatesFunction`.
Indeed, we call the aggregate's `update` function once we reach `STANDARD_VECTOR_SIZE` new inputs, and we iterate/find these inputs in the order of the elements of the incoming lists. That order is dependent on how we receive the incoming chunk from the previous operator.

```cpp
for (idx_t i = 0; i < count; i++) {
		// ...
		for (idx_t child_idx = 0; child_idx < list_entry.length; child_idx++) {
			// states vector is full, update
			if (states_idx == STANDARD_VECTOR_SIZE) {
				// update the aggregate state(s)
				Vector slice(child_vector, sel_vector, states_idx);
				aggr.function.update(&slice, aggr_input_data, 1, state_vector_update, states_idx);

				// reset values
				states_idx = 0;
			}
		}
	}
```

taniabogatsch:
Maybe I am missing something, but how are we seeing different results on each execution of a `SUM` over the entire table?
```sql
SELECT
    sum(x) -- different results on each execution
FROM df1;
```

soerenwolfers:
> Maybe I am missing something, but how are we seeing different results on each execution of a `SUM` over the entire table?
> 
> ```sql
> SELECT
>     sum(x) -- different results on each execution
> FROM df1;
> ```

Floating point summation is not associative.

soerenwolfers:
> That order is dependent on how we receive the incoming chunk from the previous operator.

Could you confirm whether that means what's in this PR is right or not?

docs/preview/sql/functions/list.md
```diff
@@ -1085,6 +1085,9 @@ The function [`list_aggregate`](#list_aggregatelist-name) allows the execution o
 
 `list_aggregate` accepts additional arguments after the aggregate function name. These extra arguments are passed directly to the aggregate function, which serves as the second argument of `list_aggregate`.
 
+Order-sensitive aggregate functions are applied in the order of the list. The `ORDER BY`, `DISTINCT` and `FILTER` clauses are not supported by `list_aggregate`.
```
taniabogatsch, line 1088: in the order of the list, as emitted from the previous operator in the execution pipeline.

soerenwolfers: Most of the docs are at a level of abstraction that doesn't involve physical operators or execution pipelines. If there are cases where the aggregate could be applied in an order that's different to the order of the list I think that should be simplified to "may be applied in an order different to that of the list itself". If not, I think the current phrasing in the PR is fine. Since I myself don't understand duckdb beyond the level of abstraction that the docs offer, I'm actually not sure which it is.

taniabogatsch: Fair, let's keep it at that phrasing, then. :)

szarnyasg:
Thank you both!

PR #5711, opened 2025-08-24
Adding Scope to set_variable.md page

Ian-Fogelman:
- Adding the h3 `Variable Scope` section to `docs/preview/sql/statements/set_variable.md`
- Moving the single set global code example to its own section on `docs/stable/sql/statements/set.md`.
- Adding an additional example of setting the global variable `threads` value.

### DIFF
--- a/docs/stable/sql/statements/set.md
+++ b/docs/stable/sql/statements/set.md
@@ -38,7 +38,16 @@ Retrieve configuration value:
 
 ```sql
 SELECT current_setting('threads');
+
 ```
+
+Set the default collation for the session:
+
+```sql
+SET SESSION default_collation = 'nocase';
+```
+
+#### Set a Global Variable
 
 Set the default sort order globally:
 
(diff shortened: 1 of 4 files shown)


docs/stable/sql/statements/set.md
```diff
@@ -38,18 +38,27 @@ Retrieve configuration value:
...
+
+#### Set a Global Variable
+
 Set the default sort order globally:
 
 ```sql
 SET GLOBAL sort_order = 'desc';
 ```
 
-Set the default collation for the session:
```
Ian-Fogelman, line 49: This is still on the page, the diff just makes it looks like it was removed from the page, but its still there.

Ian-Fogelman:
Hi @szarnyasg, here is my first draft on this, this is ready for review, cheers!

docs/stable/sql/statements/set.md
szarnyasg: Thanks for pointing this out!

Issue #5685, opened 2025-08-20
ART index limitation docs


rynoV:
Following the change from https://github.com/duckdb/duckdb/pull/15092, the [ART index limitation doc](https://duckdb.org/docs/stable/sql/indexes#limitations-of-art-indexes) was updated in https://github.com/duckdb/duckdb-web/pull/4358 to remove the notice:

> Due to the presence of transactions, data can only be removed from the index after (1) the transaction that performed the delete is committed, and (2) no further transactions exist that refer to the old entry still present in the index. As a result of this – transactions that perform *deletions followed by insertions* may trigger unexpected unique constraint violations, as the deleted tuple has not actually been removed from the index yet.

I wonder if the item "(2) no further transactions exist that refer to the old entry still present in the index" should be added back somewhere?

As far as I can tell, https://github.com/duckdb/duckdb/pull/15092 doesn't address the limitations around concurrent transactions referencing the same record in an index.

Specifically, in my code I believe I'm running into this limitation: on a table with a primary key I have one long-running query which references a record, and concurrently a transaction which deletes that record followed by a transaction which re-inserts the record, and the re-insert fails with a unique constraint violation.

rynoV:
Closed by #5724

PR #5724, opened 2025-08-27
Document ART index limitation with concurrent transactions and deletes

rynoV:
Addresses #5685

### DIFF
--- a/docs/1.2/sql/indexes.md
+++ b/docs/1.2/sql/indexes.md
@@ -109,3 +109,24 @@ Violates foreign key constraint because key "id: 1" is still referenced by a for
 
 The reason for this is because DuckDB does not yet support “looking ahead”.
 During the `INSERT`, it is unaware it will reinsert the foreign key value as part of the `UPDATE` rewrite.
+
+### Constraint Checking After Delete With Concurrent Transactions
+
+When a delete is committed on a table with an ART index, data can only be removed from the index when no further transactions exist that refer to the deleted entry. This means if you perform a delete transaction, a subsequent transaction which inserts a record with the same key as the deleted record can fail with a constraint error if there is a concurrent transaction referencing the deleted record. Pseudocode to demonstrate:
+
+```
+// Assume "someTable" is a table with an ART index preventing duplicates
+tx1 = duckdbTxStart()
+someRecord = duckdb(tx1, "select * from someTable using sample 1 rows")
+
+tx2 = duckdbTxStart()
+duckdbDelete(tx2, someRecord)
+duckdbTxCommit(tx2)
+
+// At this point someRecord is deleted, but the ART index is not updated, so the following would fail with a constraint error:
+// tx3 = duckdbTxStart()
+// duckdbInsert(tx3, someRecord)
+// duckdbTxCommit(tx3)
+
+duckdbTxCommit(tx1) // Following this, the above insert would succeed because the ART index was allowed to update
+```


docs/1.2/sql/indexes.md
```diff
@@ -108,4 +108,25 @@ Violates foreign key constraint because key "id: 1" is still referenced by a for
...
 The reason for this is because DuckDB does not yet support “looking ahead”.
-During the `INSERT`, it is unaware it will reinsert the foreign key value as part of the `UPDATE` rewrite.
\ No newline at end of file
+During the `INSERT`, it is unaware it will reinsert the foreign key value as part of the `UPDATE` rewrite.
+
+### Constraint Checking After Delete With Concurrent Transactions
+
+When a delete is committed on a table with an ART index, data can only be removed from the index when no further transactions exist that refer to the deleted entry. This means if you perform a delete transaction, a subsequent transaction which inserts a record with the same key as the deleted record can fail with a constraint error if there is a concurrent transaction referencing the deleted record. Pseudocode to demonstrate:
+
+```
```
szarnyasg:
```suggestion
```cpp
```

docs/1.2/sql/indexes.md
```diff
@@ -108,4 +108,25 @@ Violates foreign key constraint because key "id: 1" is still referenced by a for
 ```
 
 The reason for this is because DuckDB does not yet support “looking ahead”.
-During the `INSERT`, it is unaware it will reinsert the foreign key value as part of the `UPDATE` rewrite.
\ No newline at end of file
+During the `INSERT`, it is unaware it will reinsert the foreign key value as part of the `UPDATE` rewrite.
+
+### Constraint Checking After Delete With Concurrent Transactions
+
+When a delete is committed on a table with an ART index, data can only be removed from the index when no further transactions exist that refer to the deleted entry. This means if you perform a delete transaction, a subsequent transaction which inserts a record with the same key as the deleted record can fail with a constraint error if there is a concurrent transaction referencing the deleted record. Pseudocode to demonstrate:
```
taniabogatsch: `committed on a table with an index` (happens for all our indexes, also those in extensions). There, we just don't notice, as we don't use them for constraint checking.

taniabogatsch: `referencing the deleted record. Note that constraint violations are only relevant for PK, FK, and UNIQUE indexes.` Maybe expand on this? Not sure.

docs/1.2/sql/indexes.md
```diff
@@ -108,4 +108,25 @@ Violates foreign key constraint because key "id: 1" is still referenced by a for
...
+
+```cpp
+// Assume "someTable" is a table with an ART index preventing duplicates
```
taniabogatsch: `with an index enforcing a constraint violation` - I think this can also happen for FKs, no (not only duplicate checks)? Oh, can we actually enter an inconsistent state here? I.e., I delete from the PK table, but keep the value alive. Then, I insert into the FK table, and now the delete goes through. ??

rynoV: I just checked this using the .NET SDK and it does seem to get into an inconsistent state. This is the pseudocode: 

```cpp
// Setup: Create a primary table with UUID primary key and a secondary table with foreign key reference
primaryId = generateNewGUID()
conn = duckdbConnectInMemory()

// Create tables and insert initial record in primary table
duckdb(conn, "CREATE TABLE primary_table (id UUID PRIMARY KEY)")
duckdb(conn, "CREATE TABLE secondary_table (primary_id UUID, FOREIGN KEY (primary_id) REFERENCES primary_table(id))")
duckdbInsert(conn, "primary_table", {id: primaryId})

// Start transaction tx1 which will read from primary_table
tx1 = duckdbTxStart(conn)
readRecord = duckdb(tx1, "SELECT id FROM primary_table LIMIT 1")
// Note: tx1 remains open, holding locks/resources

// Outside of tx1, delete the record from primary_table
duckdbDelete(conn, "primary_table", {id: primaryId})

// Try to insert into secondary_table with foreign key reference to the now-deleted primary record
// This succeeds because tx1 is still open and the constraint isn't fully enforced yet
duckdbInsert(conn, "secondary_table", {primary_id: primaryId})

// Commit tx1, releasing any locks/resources
duckdbTxCommit(tx1)

// Verify the primary record is indeed deleted
count = duckdb(conn, "SELECT COUNT(*) FROM primary_table WHERE id = $primaryId",{primaryId: primaryId})
assert(count == 0, "Record should be deleted")

// Verify the secondary record with the foreign key reference exists
// This demonstrates the inconsistent state
count = duckdb(conn, "SELECT COUNT(*) FROM secondary_table WHERE primary_id = $primaryId",{primaryId: primaryId})
assert(count == 1, "Foreign key reference should exist")
```

If tx1 is committed before the secondary_table insert, the insert fails as expected, but if not the assertions succeed

rynoV:
Thanks @taniabogatsch , I updated the section, let me know if that covers your suggestions

I also opened https://github.com/duckdb/duckdb/issues/19002 based on your comment

taniabogatsch:
Thanks! Looks good, I think covering that inconsistent FK state makes sense, even though it is a bug (as far as I can tell) and has to be fixed - at which point we can update the documentation again. And thanks for opening that issue 👍

szarnyasg:
Thanks!

PR #5738, opened 2025-09-02
science page and small fixes and updates

Franz-Kafka:
- added science page
- added data-banner-version to banner element
- banner color can be changed with css classes
- fixed wrong page icon in docs on mobile pages in darkmode


szarnyasg:
Thanks!

PR #5743, opened 2025-09-03
Add BIRNE paper

timfi:
This PR adds BIRNE paper from DBPL 2025 (co-located with SIGMOD 2025) which builds and greatly extends upon the [Flummi paper](https://duckdb.org/science/sql-engines-excel-at-execution-of-imperative-programs/) already on the list.

Note that there wasn't a dedicated image/thumbnail for DBPL, so for now I co-opted the image for the co-located SIGMOD. I'm happy to amend this PR appropriately when the proper image becomes available.


szarnyasg:
Thanks!

PR #5748, opened 2025-09-04
Add CIDR 2024 paper on UDF inlining/decorrelation

Teggy:
Studies UDF-to-SQL transpilation and inlining and the associated role of decorrelation in DuckDB.  Led to Sam Arch's (CMU) implementation of decorrelation for nested LATERAL joins in DuckDB.


szarnyasg:
Thanks!

PR #5756, opened 2025-09-09
small improvements

Franz-Kafka:
- hyphens for science blog posts h1
- moved banner over header
- changed banner logic: banner can be turned off and on via front matter. banner content will be edited in include landingpage_banner.html
- added mini footer to docs pages


szarnyasg:
Thanks!

PR #5757, opened 2025-09-09
Fix link to Peloton DB

ajor:
The old domain appears to have been squatted and now redirects to some kind of Indonesian gambling site: armstrongforoakland[.]com (itself another squatted domain).


szarnyasg:
Thanks!

PR #5759, opened 2025-09-10
Update with.md

jdoig:
Improve readability of "connected components" example:
* lengthen CTE name to avoid confusion with 1-2 character aliases in body of query (`e.`, `n.`).
* give lateral joins aliases that make it clear which part of the iteration they refer to.


szarnyasg:
Thanks!

PR #5761, opened 2025-09-11
Use text block on Java example on landing page

ggam:
Text blocks are available from Java 15 (September 2020). This makes the example code simpler.


szarnyasg:
This is great, thanks!

PR #5764, opened 2025-09-12
updated installation page

Franz-Kafka:
@szarnyasg please have a look at the updated installation page. 
The current installation page is still live as before. The new version is available at /install.
- The code snippets are placed directly in _includes/installation.html. This gives us flexibility for adding additional information (like notes), and the code blocks are already pre-rendered correctly.
- There’s a new file _data/checksums.yml for the SHA hashes and filenames.
- The download links (with version numbers) are generated automatically using the current version from the config. 
- I tried to identify and include all installation methods, but please double-check if everything looks right.

### DIFF
--- a/_includes/installation.html
+++ b/_includes/installation.html
@@ -225,7 +225,7 @@ install.packages("duckdb")
     data-environment="java"
     data-docs="/docs/stable/clients/java">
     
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks!

PR #5766, opened 2025-09-12
Highlight storage difference between INT8 and UINT8 column type aliases

jgrg:
Closes https://github.com/duckdb/duckdb/discussions/18973


szarnyasg:
Thanks!

PR #5777, opened 2025-09-15
DuckDB Everywhere

Franz-Kafka:
- prepared for DuckDB Everywhere (/everywhere). 
- fixed issue with codebox in list item


szarnyasg:
excellent, thank you!

PR #5794, opened 2025-09-17
Expanded docs on using the ADBC driver

evertlammerts:
Ref: https://github.com/duckdblabs/voltron-data/issues/60


mlafeldt:
Just saw this - great update. 👍🏻 

One note on macOS: I usually use `DYLD_FALLBACK_LIBRARY_PATH` as recommnded at https://stackoverflow.com/a/3172515

PR #5845, opened 2025-09-26
Docker image

szarnyasg:
Add instructions for using DuckDB Docker, and for using the DuckDB UI.


jraymakers:
I tested the instructions for running the DuckDB UI from Docker (from a Apple Silicon Mac) and they worked!

szarnyasg:
Thanks!

PR #5851, opened 2025-09-29
Install page: Add arm64/ODBC windows nightly

szarnyasg:


szarnyasg:
meh, this file is not used anymore. merging it still

PR #5864, opened 2025-10-01
add merge by source/target conditions

guillesd:


guillesd:
Gonna merge this! Honestly it is not worth it to add to the rrdiagram. `BY SOURCE` only fits in the `NOT MATCHED` condition but can however `DELETE` and `UPDATE`, which will make the diagram explode.

PR #5869, opened 2025-10-01
remove duplicate nanoarrow community extension entry

jpdenford:


szarnyasg:
Thanks! This is auto-generated so we'll need to take a look why this happens

carlopi:
There are 2 community extensions, `arrow` and `nanoarrow`, and while running them executes the same code there are indeed two.
So current status looks about right.

PR #5878, opened 2025-10-03
describe aws validation usage

benfleis:
also tweak incorrectly tagged literals in examples

### DIFF
--- a/docs/preview/core_extensions/aws.md
+++ b/docs/preview/core_extensions/aws.md
@@ -30,9 +30,9 @@ The default provider, `config` (i.e., user-configured), allows access to the S3 
 CREATE OR REPLACE SECRET secret (
     TYPE s3,
     PROVIDER config,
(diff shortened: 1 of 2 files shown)


szarnyasg:
@benfleis, thanks! Two things:

1.

> also tweak incorrectly tagged literals in examples

the literals are supposed to mimic the AWS documentation, which also uses red+italic for highlighting the fields that you need to change in a script: https://docs.aws.amazon.com/glue/latest/dg/start-usage-profiles-managing.html#w9aac15c15c19b5

They are rendered as follows: https://duckdb.org/docs/stable/core_extensions/aws.html#config-provider

Screenshot 2025-10-04 at 09 36 33

2. There's no need to target both the `docs/stable/` and the `docs/preview/` directories: we periodically copy stuff from `stable` to `preview`.

szarnyasg:
Thanks!

PR #5886, opened 2025-10-05
Fix docker documentation links in FAQ

eitsupi:


szarnyasg:
Hi, thanks for the PR! This was indeed off but this PR fixed the wrong link! :) Will fix it via a separate commit.

PR #5916, opened 2025-10-06
Update interval.md

soerenwolfers:
more precise description of parts extraction from intervals.

note, I cargo culted `<!-- markdownlint-enable MD056 -->`, might need removal


szarnyasg:
Thanks! I dropped the MD056 linter instructions - we have to disable them in cases when there are `|` characters in the table, which are confused with column separators.

PR #5919, opened 2025-10-07
Adding JSON examples, BLOB options,examples

Ian-Fogelman:
Updating the COPY page per https://github.com/duckdb/duckdb-web/issues/5912

- I noticed that the `JSON` copy commands were lacking examples, so I added them as part of this PR. This is an extremely powerful feature that I feel needs examples for discoverability.
- Adding `BLOB` format content, as far as I could tell `COMPRESSION` is the only option currently supported for this format option. 
- I added comments for the returned values of the JSON examples, I am open to feedback here, I was just trying to keep the content minimal while still displaying nuances of the various JSON format options.

### Links

Stable: 
http://localhost:4000/docs/stable/sql/statements/copy.html#json-options
http://localhost:4000/docs/stable/sql/statements/copy.html#blob-options

### DIFF
--- a/docs/stable/sql/statements/copy.md
+++ b/docs/stable/sql/statements/copy.md
@@ -442,7 +442,7 @@ The below options are applicable when writing `BLOB` files.
 
 | Name | Description | Type | Default |
 |:--|:-----|:-|:-|
-| `COMPRESSION` | The compression type for the file. By default this will be detected automatically from the file extension (e.g., `file.json.gz` will use `gzip`, `file.json.zst` will use `zstd`, and `file.json` will use `none`). Options are `none`, `gzip`, `zstd`. | `VARCHAR` | `auto` |
+| `COMPRESSION` | The compression type for the file. By default this will be detected automatically from the file extension (e.g., `file.blob.gz` will use `gzip`, `file.blob.zst` will use `zstd`, and `file.blob` will use `none`). Options are `none`, `gzip`, `zstd`. | `VARCHAR` | `auto` |
 
 Type casts the string value `foo` to the `BLOB` data type and outputs the results to `blob_output.blob`:
 


szarnyasg:
Hi @Ian-Fogelman, thanks for the PR!

Three things:

1. I pushed some small edits to the PR, please sync your branch

2. There's no need to target docs/preview for now - we are going to target v1.4 / stable (LTS) for the next few months

3. I don't fully follow this example:

Sets the value of column `compression_type` to `gzip` and outputs the results to `compression.json.gz`:

```sql
COPY (SELECT 'gzip' AS compression_type) TO 'compression.json.gz';
-- RETURNS: {"compression_type":"gzip"}
```

It the `compression_type` field doesn't have an on the actual compression used to serialize the file, right? So I believe this example can be confusing for newcomers.

Ian-Fogelman:
Hi @szarnyasg , 



  1. I pushed some small edits to the PR, please sync your branch
  - I have synced your changes

  2. There's no need to target docs/preview for now - we are going to target v1.4 / stable (LTS) for the next few months
   - I have removed any edits from `/preview` folder, thanks for confirming this will keep in mind for upcoming PR's.

  3. I don't fully follow this example: Sets the value of column compression_type to gzip and outputs the results to compression.json.gz
    
  - The use case I am trying to highlight here is that despite not providing a explicit parameter such as `COMPRESSION 'GZIP'  i.e: 
  
  ```COPY (select 'foo'::BLOB) TO 'blob_output_gzip.blob' (FORMAT BLOB, COMPRESSION 'GZIP');```
    
  The results of the query will still be outputted in a gzip compressed format, this is highlighting the text in the table above the format options: 
    
    
image


I did check and it looks like query results are indeed compressed (`Size on disk` is smaller than `Size`):

image


Happy to remove the example if you think its confusing, as the previous code example does the same thing. Please LMK.

szarnyasg:
> - The use case I am trying to highlight here is that despite not providing a explicit parameter such as `COMPRESSION 'GZIP'  i.e: 

This happens because the extension is auto-detected by DuckDB. If you drop the extension, you get a plain text file despite what the subquery defines as the `compression_type` column.

```
D COPY (SELECT 'gzip' AS compression_type) TO 'compression.json';
D .sh cat compression.json
{"compression_type":"gzip"}
```

Ian-Fogelman:
Thanks @szarnyasg , [I have removed the two examples](https://github.com/duckdb/duckdb-web/pull/5919/commits/86ec22d1e017d1ae7e4ab8b328ea35f2c621b5a6) that did not specify `COMPRESSION` explicitly in the syntax. Please let me know if any other changes are needed, also a few nit cleanups.

szarnyasg:
This looks good, thanks!

PR #5921, opened 2025-10-07
Search on blog page with Ctrl+K

Franz-Kafka:
- Fixed JS for search on blog page with ctrl+k
- SCSS updates / sync


szarnyasg:
Thanks!

PR #5924, opened 2025-10-07
Update python dev docs

evertlammerts:
The previous docs were still assuming the Python extension is in-tree. This update might be a bit on the light side / not greatly structured, but at least the docs aren't incorrect anymore. Most people should be able to get going from here on out, I think.

### DIFF
--- a/docs/stable/dev/building/python.md
+++ b/docs/stable/dev/building/python.md
@@ -5,324 +5,228 @@ redirect_from:
 title: Python
 ---
 
-The DuckDB Python package lives in the main [DuckDB source on GitHub](https://github.com/duckdb/duckdb/) under the `/tools/pythonpkg/` folder. It uses [pybind11](https://pybind11.readthedocs.io/en/stable/) to create Python bindings with DuckDB.
+The DuckDB Python package has its own repository at [duckdb/duckdb-python](https://github.com/duckdb/duckdb-python) and uses [pybind11](https://pybind11.readthedocs.io/en/stable/) to create Python bindings with DuckDB.
 
 ## Prerequisites
 
-For everything described on this page we make the following assumptions:
+This guide assumes:
 
-1. You have a working copy of the DuckDB source (including the git tags) and you run commands from the root of the source.
-2. You have a suitable Python installation available in a dedicated virtual environment.
+1. You have a working copy of the DuckDB Python package source (including git submodules and tags)
+2. You have [Astral UV](https://docs.astral.sh/uv/) version >= 0.8.0 installed
+3. You run commands from the root of the duckdb-python source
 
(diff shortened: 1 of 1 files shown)


szarnyasg:
Great! Let's get this in because people are looking for this quite often.

docs/stable/dev/building/python.md
```diff
@@ -5,324 +5,228 @@ redirect_from:
...
-
-If you seen an error that looks like this:
-
-```console
-ImportError: dlopen(/usr/bin/python3/site-packages/duckdb/duckdb.cpython-311-darwin.so, 0x0002): symbol not found in flat namespace '_MD5_Final'
-```
-
-... then you've probably tried to link in a problematic extension. As mentioned above: `tools/pythonpkg/duckdb_extension_config.cmake` contains the default list of extensions that are built with the Python package. Any other extension might cause problems.
-
-### Python Fails with `No module named 'duckdb.duckdb'`
```
Tishj, line 314: Not sure why we're removing this, I think this is still a very common pitfall

docs/stable/dev/building/python.md
```diff
@@ -5,324 +5,228 @@ redirect_from:
...
 
-> For more details on building DuckDB extensions look at the [documentation]({% link docs/stable/dev/building/building_extensions.md %}).
-
-The DuckDB build process follows the following logic for building extensions:
-
-1. First compose the complete set of extensions that might be included in the build.
-1. Then compose the complete set of extensions that should be excluded from the build.
-1. Assemble the final set of extensions to be compiled by subtracting the set of excluded extensions from the set of included extensions.
-
-The following mechanisms add to the set of **_included_ extensions**:
```
Tishj, line 111: Why are we removing this section entirely?
I think it still holds true for the most part, only it doesn't need to be repeated so verbosely as these are duckdb core build flags.

The important thing in this section is to describe how CMake build flags can be provided when building duckdb-python

docs/stable/dev/building/python.md
evertlammerts: Well, if the extension is installed as described in this guide then the import clash does no longer happen. It's installed in editable mode, so importing `duckdb` from the repo root is fine.

docs/stable/dev/building/python.md
evertlammerts: It's definitely informative, but I don't think it's very relevant for the Python build anymore since the vendored duckdb is built the same way as core is. [Building Extensions](https://duckdb.org/docs/stable/dev/building/building_extensions) tells users how they can try to add extensions to the build. That's also a bit more straightforward, the section as it was here was correct and complete, but also provided too many knobs imo.

PR #5928, opened 2025-10-07
Fix download metrics example query

zfarrell:
There are two issues with the current "download metrics" [example query](https://duckdb.org/community_extensions/download_metrics):

Latest version of duckdb requires `_last_update` to be explicitly casted. Copying and pasting the query as-is, results in:
```
Binder Error:
No function matches the given name and argument types 'date_trunc(STRING_LITERAL, VARCHAR)'. You might need to add explicit type casts.
	Candidate functions:
	date_trunc(VARCHAR, DATE) -> TIMESTAMP
	date_trunc(VARCHAR, INTERVAL) -> INTERVAL
	date_trunc(VARCHAR, TIMESTAMP) -> TIMESTAMP
	date_trunc(VARCHAR, TIMESTAMP WITH TIME ZONE) -> TIMESTAMP WITH TIME ZONE
```

Additionally, historical data is missing and causes example to fail. Updating to when the data was fixed. If you don't update the date, you'll see errors like:
```
HTTP Error:
Unable to connect to URL "https://community-extensions.duckdb.org/download-stats-weekly/2025/37.json": 404 (Not Found).
```

### DIFF
--- a/community_extensions/download_metrics.md
+++ b/community_extensions/download_metrics.md
@@ -23,7 +23,7 @@ INTO NAME extension VALUE downloads_last_week
 ORDER BY downloads_last_week DESC;
 ```
 
-For example, to return the download counts for the weeks since October 1, 2024:
+For example, to return the download counts for the weeks since September 22, 2025:
 
 ```sql
 PIVOT (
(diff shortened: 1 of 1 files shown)


szarnyasg:
@zfarrell thanks for the PR. Interestingly, this query crashes on my macOS with 1.4.0/1.4.1/nightly with an assertion error. Will file an issue.

szarnyasg:
Okay, it turns out the crash occurs locally due to the presence of a(n unused) secret. This is good to be merged!

PR #5933, opened 2025-10-08
Update create_table.md

soerenwolfers:
avoid confusion like in https://github.com/duckdb/duckdb/issues/19292


szarnyasg:
Thanks!

PR #5942, opened 2025-10-10
clarify azure MI notes

benfleis:
udpate language on Managed Identity

clarify/emphasize that credential_chain "just works" for single ID
environments, and that MI provider needed in situation where multiple
IDs are available.


szarnyasg:
Thanks!

PR #5957, opened 2025-10-13
Add or fix aggregate/window function aliases in table

szarnyasg:
- The other aliases are there in the table so I feel `arbitrary` should be there too
- Fixed the anchor link for `rank_dense`


soerenwolfers:
my own plan was to remove any duplication. If there are other aliases in the short form table that are not in the long form table that'd be an oversight of mine.

soerenwolfers:
but if course I'm happy to follow whatever structure you prefer. I'm just feeling there'd be a lot more corrective work to do then.

szarnyasg:
I'm for removing duplication!

PR #5980, opened 2025-10-15
Add Haskell tertiary client

Tritlo:
This adds the `duckdb-haskell` repo, containing `duckdb-ffi` and `duckdb-simple` libraries for using `duckdb` with Haskell.


szarnyasg:
Great, thanks a lot for rolling out this new client!

PR #6003, opened 2025-10-18
Overhaul enum docs

NickCrews:
This was inspired by https://github.com/duckdb/duckdb/pull/19345

This was the motivation for these changes:
- Remove the very misleading line "The string `sad` is cast to the type `mood`". This is trying to get at the physical optimization that is happening, but logically that is NOT what is happening.
- Adds explicit warnings using `> Warning` of how comparing to raw strings can lead to footguns

While I was at it, I also did a bunch of other improvements:

- Simplifies the section on creating enums, and puts the most common case `CREATE TYPE mood AS ENUM ('sad', 'ok', 'happy');` front and center.
- merges the two sections on creating enums from a SELECT
- renames section on "comparing" to "ordering"
- replaces ordering examples using  'mood' to more clearly ordered 'priority'
- adds examples on case-sensitivity
- uses a more concrete example (sales data and regions) for creating enum from a SELECT
- move the section on functions before the section on removal because I thought that was more important
- merge the example of enum_range() into the functions section
- removes the person_2 table from the examples. I don't think it was needed, and just provided more mental overhead
- add example of creating an enum in a schema
- rename the 'funky' mood to 'bogus' to make it even more clear that this is supposed to be a bogus value.

### DIFF
--- a/docs/preview/sql/data_types/enum.md
+++ b/docs/preview/sql/data_types/enum.md
@@ -142,7 +142,7 @@ WHERE current_mood = 'sad';
 | Pagliacci | sad          |
 
 
-> Warning This means that comparing against a bogus string always results in `false`:
+> Warning This means that comparing against a random (non-equivalent) string always results in `false` (and does not error):
 
 ```sql
 SELECT * FROM person


docs/preview/sql/data_types/enum.md
```diff
@@ -11,218 +11,233 @@ title: Enum Data Type
...
+|:----------|:-------------|
+| Pedro     | happy        |
 
-```sql
-INSERT INTO person
-VALUES ('Pedro', 'happy'), ('Mark', NULL), ('Pagliacci', 'sad'), ('Mr. Mackey', 'ok');
-```
 
-The following query will fail since the mood type does not have `quackity-quack` value.
+When comparing an enum to a `VARCHAR`, duckdb will cast the enum to `VARCHAR` and perform a string comparison:
```
szarnyasg:
```suggestion
When comparing an enum to a `VARCHAR`, DuckDB will cast the enum to `VARCHAR` and perform a string comparison:
```

docs/preview/sql/data_types/enum.md
```diff
@@ -11,218 +11,233 @@ title: Enum Data Type
...
-VALUES ('Hannes', 'quackity-quack');
+SELECT * FROM person
+WHERE current_mood = name;
+-- Equivalent to `WHERE current_mood::VARCHAR = name`
+-- No rows returned
 ```
 
-The string `sad` is cast to the type `mood`, returning a numerical reference value.
-This makes the comparison a numerical comparison instead of a string comparison.
+When comparing against a constant string, duckdb will perform an optimization
```
szarnyasg:
```suggestion
When comparing against a constant string, DuckDB will perform an optimization
```

docs/preview/sql/data_types/enum.md
```diff
@@ -11,218 +11,233 @@ title: Enum Data Type
...
+and `try_cast(<constant string>, enum_type)` so that physically
```
szarnyasg:
```suggestion
and `try_cast(⟨constant string⟩, enum_type)`{:.language-sql .highlight} so that physically
```

docs/preview/sql/data_types/enum.md
```diff
@@ -11,218 +11,233 @@ title: Enum Data Type
...
 
-If you are importing data from a file, you can create an enum for a `VARCHAR` column before importing.
-Given this, the following subquery selects automatically selects only distinct values:
 
-```sql
-CREATE TYPE mood AS ENUM (SELECT mood FROM 'path/to/file.csv');
-```
-
-Then you can create a table with the enum type and import using any data import statement:
+> Warning This means that comparing against a bogus string always results in False:
```
szarnyasg:
```suggestion
> Warning This means that comparing against a bogus string always results in `false`:
```

szarnyasg: This is confusing due to two reasons:
1) we have a string `bogus` above
2) why _wouldn't_ comparing against a bogus string always return in false?

docs/preview/sql/data_types/enum.md
```diff
@@ -11,218 +11,233 @@ title: Enum Data Type
...
 ```
 
-When comparing the `past_mood` column (string), DuckDB will cast the `current_mood` enum to `VARCHAR` and perform a string comparison:
+|   m    |
+|:-------|
+| low    |
+| medium |
+| high   |
+
+> Warning If you compare an enum to a non-enum (eg a varchar or a different enum type),
```
szarnyasg:
```suggestion
> Warning If you compare an enum to a non-enum (e.g., a `VARCHAR` or a different enum type),
```

docs/preview/sql/data_types/enum.md
```diff
@@ -11,218 +11,233 @@ title: Enum Data Type
...
 ```
 
-## Enum Removal
+| name | priority_level  |
+|:-----|:----------------|
+| b    | medium          |
 
-Enum types are stored in the catalog, and a catalog dependency is added to each table that uses them. It is possible to drop an enum from the catalog using the following command:
+
+So, if you want to eg "get all priorities at or above 'medium'" then explicitly cast to the enum type:
```
szarnyasg:
```suggestion
So, if you want to e.g. "get all priorities at or above `medium`" then explicitly cast to the enum type:
```

docs/preview/sql/data_types/enum.md
NickCrews: I would expect it to error, not return false. Probably should be more explicit here about that or rephrase.

NickCrews:
Thanks for the review! Those suggestions all look good. Hopefully my answer made sense. Feel free to edit based on that, or I can add a fixup commit when I get to a computer

docs/preview/sql/data_types/enum.md
```diff
@@ -11,218 +11,233 @@ title: Enum Data Type
...
 
-If you are importing data from a file, you can create an enum for a `VARCHAR` column before importing.
-Given this, the following subquery selects automatically selects only distinct values:
 
-```sql
-CREATE TYPE mood AS ENUM (SELECT mood FROM 'path/to/file.csv');
-```
-
-Then you can create a table with the enum type and import using any data import statement:
+> Warning This means that comparing against a bogus string always results in `false`:
```
szarnyasg:
```suggestion
> Warning This means that comparing against a random (non-equivalent) string always results in `false` (and does not error):
```

szarnyasg:
It's good, I made a small adjustment. Thanks!

NickCrews:
Thank you, looks great!

PR #6027, opened 2025-10-22
Updated landing page with metrics bar

Franz-Kafka:
- Slightly changed typewriter animation and style
- Added detailed metrics bar to landing page
- Removed small metrics from landing page
- Added blog posts to landing page (@szarnyasg we would suggest adding the section for the moment)


szarnyasg:
Great, thanks!

PR #6040, opened 2025-10-24
Adding Additional Logging Refs

Ian-Fogelman:
### Description

Adding `set` commands to the `docs/stable/configuration/overview.md` page. These set commands seem to do the same work as the `CALL enable_logging()` and similar commands on https://duckdb.org/docs/stable/operations_manual/logging/overview. Adding them on this page makes sense IMO because the various logging options are in the [Global Configuration Options](https://duckdb.org/docs/stable/configuration/overview#global-configuration-options) table.

### Links

http://localhost:4000/docs/stable/configuration/overview#examples

### DIFF
--- a/docs/stable/configuration/overview.md
+++ b/docs/stable/configuration/overview.md
@@ -43,69 +43,71 @@ SET threads TO 1;
 ```
 
 Turn logging on and set the logging level to `debug`.
-For additional details on logging levels, see [Log Level]({% link docs/stable/operations_manual/logging/overview.md%}#log-level).
+For additional details on logging levels, see [Log Level]({% link docs/stable/operations_manual/logging/overview.md %}#log-level).
 
 ```sql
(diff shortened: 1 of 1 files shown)


docs/stable/operations_manual/logging/overview.md
```diff
@@ -68,7 +68,7 @@ only log messages of the 'HTTP' type will be written to the log.
 To enable multiple log types, simply pass:
 
 ```sql
-CALL enable_logging(['HTTP', 'QueryLog']);~~~~
+CALL enable_logging(['HTTP', 'QueryLog']);
```
Ian-Fogelman, line 71: I believe this is a typo: 

image

szarnyasg: indeed, thanks for spotting it!

PR #6042, opened 2025-10-24
Add benchmark example.

hawkfish:
I couldn't remember the `argument` syntax, and the documentation didn't explain it!


szarnyasg:
This is excellent and much needed. Thank you!

PR #6048, opened 2025-10-27
Update attach.md

yoonghm:
ROW_GROUP_SIZE must be a multiple of 2048.


szarnyasg:
Well-spotted, thanks!

Issue #6094, opened 2025-11-03
Incorrect SHA256 on Linux download page


ngdangtu-vn:
This is from [the website:](https://duckdb.org/install/?platform=linux&environment=cli)
[<img width="1151" height="485" alt="Image" src="https://github.com/user-attachments/assets/e7eb8616-3af0-467b-97c0-c9254e32e216" />](https://duckdb.org/install/?platform=linux&environment=cli)

This is my third time redownload:
<img width="1244" height="826" alt="Image" src="https://github.com/user-attachments/assets/b28f042f-550f-4ac6-a095-6f0579ad357e" />

What happened then? First time I downloaded the bin file from browser I couldn't do it properly (click and browser save it to locale). It looked like this:

<img width="2031" height="851" alt="Image" src="https://github.com/user-attachments/assets/6739ff8f-d57d-4660-8390-269fe7eccac3" />

Still I could try to download it with Ctrl + S. However, due to worry about file quality by downloading this way I double checked with sha but result is like the image above. It doesn't match. So I attempted to do that again with `curl`:

```sh
curl -o duckdb.zip https://install.duckdb.org/v1.4.1/duckdb_cli-linux-amd64.zip
```
 
The sha of downloaded file is still very consistent to the first time download. So I went stupid and tried again with... curl... The result is as expected. SHA didn't match. Could you guy verify if I was assaulted (🤪) by the infamous middle man attack or the site has some problem? Thanks

Additional info about broken download link:

```header
HTTP/2 200 
date: Mon, 03 Nov 2025 09:21:43 GMT
content-type: text/plain
report-to: {"group":"cf-nel","max_age":604800,"endpoints":[{"url":"..."}]}
nel: {"report_to":"...","success_fraction":0.0,"max_age":...}
server: cloudflare
cf-cache-status: HIT
last-modified: Fri, 17 Oct 2025 11:17:50 GMT
vary: Accept-Encoding
age: ...
cache-control: max-age=...
etag: ...
content-encoding: br
cf-ray: ...
```

I can tell that setting `content-type` to `text/plain` will make browser display the file at all cost (which harm user pc as well since it cost 18.2MB ram at once).

ngdangtu-vn:
About the header of download file, I can't not do anything. But I can help to fix the not download problem from frontend with HTML code. A new hotfix PR coming if you guys don't mind...

szarnyasg:
Thanks for the report and the PR! We'll investigate this.

PR #6095, opened 2025-11-03
Hotfix to force download binary files

ngdangtu-vn:
> [!note]
> This PR attempts to fix a small aspect in #6094, the whole issue is still need to be review and resolve.

This fix basically done in the frontend with HTML code. So instead of  demand browser to open the link in new tab, we make it download file  immediately. This will ensure the visitors doesn't get redirect from  download file and, most importantly, the file can be downloaded.

However, be aware that if visitors access the link directly this is not the  right solution. To actually fix the root of this problem, we have to update  config in Cloudflare R2 to ensure they setup proper HTTP header  `content-type`.

Reference:
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/download

### DIFF
--- a/_includes/installation.html
+++ b/_includes/installation.html
@@ -202,7 +202,7 @@ install.packages("duckdb")
     data-docs="{% link docs/stable/clients/java.md %}">
 
 <h4>Direct Download (JAR)</h4>
-<a href="https://repo1.maven.org/maven2/org/duckdb/duckdb_jdbc/{{ site.current_duckdb_java_version }}/duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar" class="download-btn"="rent_duckdb_java_version }}.jar" download rel="noopener noreferrer">duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar</a>
+<a href="https://repo1.maven.org/maven2/org/duckdb/duckdb_jdbc/{{ site.current_duckdb_java_version }}/duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar" class="download-btn" download="duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar" download rel="noopener noreferrer">duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar</a>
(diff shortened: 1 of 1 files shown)


szarnyasg:
Hi, thanks for the PR. There seems to be some regex search and replace error with some of the links at the bottom. Can you please double-check the links?

ngdangtu-vn:
@szarnyasg Thanks for the review, I've fixed the last 4 lines. Could you review it one more time please?

_includes/installation.html
```diff
@@ -202,7 +202,7 @@ <h4>Docker Image</h4>
     data-docs="{% link docs/stable/clients/java.md %}">
 
 <h4>Direct Download (JAR)</h4>
-<a href="https://repo1.maven.org/maven2/org/duckdb/duckdb_jdbc/{{ site.current_duckdb_java_version }}/duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar" class="download-btn" target="_blank" rel="noopener noreferrer">duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar</a>
+<a href="https://repo1.maven.org/maven2/org/duckdb/duckdb_jdbc/{{ site.current_duckdb_java_version }}/duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar" class="download-btn"="rent_duckdb_java_version }}.jar" download rel="noopener noreferrer">duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar</a>
```
szarnyasg:
```suggestion
<a href="https://repo1.maven.org/maven2/org/duckdb/duckdb_jdbc/{{ site.current_duckdb_java_version }}/duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar" class="download-btn" download="duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar" download rel="noopener noreferrer">duckdb_jdbc-{{ site.current_duckdb_java_version }}.jar</a>
```

szarnyasg:
Thanks, patched a small other thing and it's good to go now!

PR #6103, opened 2025-11-03
Update text.md

### LINKED ISSUE #6099: [duckdb/#19298] - pg_size_pretty return different in the 'mB' vs 'Mib' needs documentation
See https://github.com/duckdb/duckdb/issues/19298

soerenwolfers:
fix #6099


szarnyasg:
That's a swift resolution of the problem, thanks!

PR #6114, opened 2025-11-05
Add missing comma in `sagemaker` docs

sethmccombs:
Nothing huge, just a missing `,` 

I looked at the scripts for generating docs but I don't believe these pages are generated


szarnyasg:
Thanks!

PR #6124, opened 2025-11-06
Fix incorrect .arrow() method references in Python SQL on Arrow guide

andrewshandy:
# Description

This PR fixes an inconsistency in the Python guide “SQL on Arrow” (located at `guides/python/sql_on_arrow.md`) where example code uses the .arrow() method to return a pyarrow.Table. In reality:

The `.arrow()` method returns a `pyarrow.lib.RecordBatchReader`, not a `pyarrow.Table`.
The correct method to obtain a full Arrow Table from the query result is `.fetch_arrow_table()`. As specified in [Export To Arrow Guide](https://duckdb.org/docs/stable/guides/python/export_arrow)

Changes
Updated all examples in the guide from:
`results = con.execute("...").arrow()`

to:

`results = con.execute("...").fetch_arrow_table()`


szarnyasg:
Thanks for the PR! I ported the change to stable (v1.4) from preview (which will only be the default starting from v1.5)

PR #6135, opened 2025-11-09
Add QueryResult error handling example in C++ api docs

### LINKED ISSUE #6134: Mention GetErrorType() and GetErrorObject() in C++ API
Currently on the C++ API page under the Query section
https://duckdb.org/docs/stable/clients/cpp#:~:text=The%20MaterializedQueryResult%20instance,as%20a%20string
it does not mention the `GetErrorType()` and `GetErrorObject()` methods for `QueryResult` which is useful for error handling.

I'd like to add the mention of these two functions under that section.

monthdev:
Reference to issue I raised before, fixes #6134


szarnyasg:
Thanks!

PR #6136, opened 2025-11-09
Elaborate further on index limitation documentation

taniabogatsch:
We're still seeing issues where the index limitations cause confusion, so maybe more context can help here.

### DIFF
--- a/docs/stable/sql/indexes.md
+++ b/docs/stable/sql/indexes.md
@@ -114,7 +114,7 @@ Violates foreign key constraint because key "id: 1" is still referenced by a for
 The reason for this is that DuckDB does not yet support “looking ahead”.
 During the `INSERT`, it is unaware it will reinsert the foreign key value as part of the `UPDATE` rewrite.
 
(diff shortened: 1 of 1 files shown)


docs/stable/sql/indexes.md
```diff
@@ -110,63 +111,101 @@ Constraint Error:
 Violates foreign key constraint because key "id: 1" is still referenced by a foreign key in a different table. If this is an unexpected constraint violation, please refer to our foreign key limitations in the documentation
 ```
 
-The reason for this is because DuckDB does not yet support “looking ahead”.
+The reason for this is that DuckDB does not yet support “looking ahead”.
 During the `INSERT`, it is unaware it will reinsert the foreign key value as part of the `UPDATE` rewrite.
 
 ### Constraint Checking After Delete With Concurrent Transactions
```
szarnyasg:
```suggestion
### Constraint Checking after Delete with Concurrent Transactions
```

szarnyasg:
Thanks!

PR #6159, opened 2025-11-13
Add DBD::DuckDB in tertiary client list

giterlizzi:
Add DuckDB database driver for the DBI module.

The link contains the CPAN guide for the DuckDB DBD driver.

If we need a quick guide to the main features of the DBD driver, I can produce a quick-guide like PHP (https://duckdb.org/docs/stable/clients/php).


szarnyasg:
Thanks, adding it to the list is great. If you'd like to submit a guide that's fine but not required.

giterlizzi:
Thanks!

PR #6198, opened 2025-11-23
Update iceberg extension dependency information

djouallah:
small typo, iceberg has no dependencies on third party libraries


szarnyasg:
Thanks!

PR #6211, opened 2025-11-26
fix typo: prefix -> suffix

niykko:
I came across this small mistake, I don't think it's the end of the world or anything but I'd like to fix it =)


szarnyasg:
Thanks!

szarnyasg:
Thanks!

PR #6212, opened 2025-11-26
Update BUILDING.md

Ian-Fogelman:
Adding `-f` to git clean command, today when I attempted to use the command `git clean` I got the following error, the solution was to use  `git clean -f`: 

```bash
$ git clean
fatal: clean.requireForce defaults to true and neither -i, -n, nor -f given; refusing to clean
(base)
Ian@DESKTOP-6L1B155 MINGW64 /c/repos/duckdb-web (blog-post-merge-statement)
$ git clean -f
```


szarnyasg:
Hi, the incomplete command is there on purpose: I don't want people to wipe their local Git repository without thinking about using `-f` first.

PR #6217, opened 2025-11-28
Fix `date_diff function` parameters in documentation

balintdecsi:
`startdate` appears twice. For clearance, I fixed the typo to `startdate` and `enddate`.


szarnyasg:
Köszi!

PR #6184, opened 2025-11-19
table visualizer

Franz-Kafka:
- Added demo for table visualizer (published: false)

PR #6227, opened 2025-12-01
Table visualizer (ex https://github.com/duckdb/duckdb-web/pull/6184)

carlopi:
Added an extra cleanup on top (removing some fields not strictly needed)

Thanks @Franz-Kafka and @jonathanauch!


szarnyasg:
Thanks!

PR #6230, opened 2025-12-01
Correct pos range for quantile_cont and documented negative pos behav…

SoyIsrael:
This PR updates documentation for quantile_cont(x, pos):
- Changes `pos` takes values from `0` to `1` to `-1` to `1`.
- Adds explanation of negative `pos` behavior and mentions `quantile_cont(x, -y) = quantile_cont(x, 1 - y)`.


szarnyasg:
Thanks!

PR #6232, opened 2025-12-02
fix(docs): modify example 1 and the result for `rtrim()`, `ltrim()`, and `trim()`.

norangmangto:
### Changed files
* `docs/preview/sql/funcitons/text.md`
* `docs/stable/sql/functions/text.md`.

### Changes
* modify the example 1 to use the same string for those 3 functions
* fixed an issue that multiple spaces are not shown correctly. (it's because,  multiple spaces are trimmed in an inline code block)


szarnyasg:
Thanks!

PR #6242, opened 2025-12-03
Update installation commands for iceberg extension

dsevilla:
It incorrectly listed the json extension instead of the iceberg extension.


szarnyasg:
Oops, that was a copy-paste error from me. Thanks for submitting a fix!

PR #6256, opened 2025-12-09
More index limitation docs

taniabogatsch:
See https://github.com/duckdb/duckdb/issues/20072#issuecomment-3632770947

### DIFF
--- a/docs/stable/sql/indexes.md
+++ b/docs/stable/sql/indexes.md
@@ -133,6 +133,9 @@ Let's say that a `value-row_id` pair exists in the global index and there is a `
 In that case, it **also** stays visible to newer transactions until all **older**, dependent transactions have finished.
 That behavior causes two main limitations, which are listed in more detail below.
 
+Please also note that the limitation extends to concurrent changes to multiple tables.
+Older read transactions on table X can cause write-write conflicts on table Y, if they are in the same schema.
+
 The long-term solution to these limitations is to enable transaction-timestamp tracking in indexes.
 However, as-of now, DuckDB does not fully support MVCC for its indexes.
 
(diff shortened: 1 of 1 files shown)


szarnyasg:
Great, thank you!

docs/stable/sql/indexes.md
```diff
@@ -133,6 +133,9 @@ Let's say that a `value-row_id` pair exists in the global index and there is a `
 In that case, it **also** stays visible to newer transactions until all **older**, dependent transactions have finished.
 That behavior causes two main limitations, which are listed in more detail below.
 
+Please also note that the limitation extends to concurrent changes to multiple tables.
+Older read transactions on table X can cause write-write conflicts on table Y, if they are in the same schema.
```
Mytherin, line 137: This is not the case. Changes made to a table are completely separate from changes made to another table.

What is actually happening in the below example is that a thread is conflicting with earlier changes made by itself, e.g. consider the changes made by `thread 1`:

```sql
# i = 1
INSERT OR REPLACE INTO schema__test.test_table1 ...

# i = 2
INSERT OR REPLACE INTO schema__test.test_table1 ...

# etc...
```

Now the transaction at `i = 2` can conflict with the transaction made at `i = 1` if that transaction is not cleaned up.

In essence, the changes made to the other tables have nothing to do with this, other than the fact that they prevent the previous transaction from the same thread from being cleaned up.

Mytherin: I would also say this example might be a bit unnecessarily complicated - the same behavior can be shown just with `BEGIN`, e.g.:

```sql
statement ok
SET immediate_transaction_mode=true;

statement ok
CREATE TABLE tbl(i INT PRIMARY KEY);

statement ok
INSERT INTO tbl VALUES (1), (2), (3);

statement ok con1
BEGIN

statement ok con2
DELETE FROM tbl WHERE i=2;

statement ok con2
INSERT INTO tbl VALUES (2);
--- fails with: Duplicate key "i: 2" violates primary key constraint
```

taniabogatsch: We already have the second example in the documentation (IIRC)

> In essence, the changes made to the other tables have nothing to do with this, other than the fact that they prevent the previous transaction from the same thread from being cleaned up.

That's what I was trying to say - I'll rephrase it. I wanted to specifically cover the case where a change **to a different table** can prevent the transaction from being cleaned up, causing the subsequent conflict on the table.

taniabogatsch: > The issue is that the old index data is not being cleaned up because we have other transactions still active in that schema. Our transaction-support does not have table-granularity here. As a result we get an index conflict, because there is a read transaction active on the schema making us persist the old index version beyond it having been updated.

This is what I wrote in the linked comment - I can try to update the documentation accordingly?

taniabogatsch: > concurrent changes to multiple tables. [...]
> Older read transactions

ahhh I said `concurrent changes` and then `read transactions` 🤦

taniabogatsch: https://github.com/duckdb/duckdb-web/pull/6266

m-dr25: Does this happen only when there is an INDEX (due to Primary key)? If we remove INDEX, and use MERGE INTO or use Appender - there reads/writes/writes won't conflict as there is no 'schema' to being with for conflict, right?

Mytherin: This is all purely related to indexes yes, so no index or primary key/unique constraints and these issues will not happen

PR #6264, opened 2025-12-09
Update aggregates.md

soerenwolfers:
I find discrete elements (like strings) less confusing than using the same `[0, 1, 10]` that's used for the example of the regular `histogram(...)` function, to emphasize the distinction between the `histogram(..., upper_boundaries_of_interval)` and `histogram_exact(..., finitely_many_elements_i_care_for)`.


szarnyasg:
Good point, thanks!

PR #6278, opened 2025-12-11
Rework syntax highlighter to use fewer colors

szarnyasg:


szarnyasg:
See e.g., <https://syntax-highlight-rework.duckdb-web.pages.dev/docs/stable/sql/statements/pivot>

PR #6300, opened 2025-12-15
More changes to blogpost

carlopi:
Main change is bringing s3table sections together, @Tmonster if you want yet another go


szarnyasg:
Thanks!

PR #6307, opened 2025-12-16
docs: fix invalid SQL syntax in CTE example

vinayaksodar:
The example demonstrating column aliasing for CTEs was missing a  SELECT statement (showing only `FROM cte`). 

This commit updates the snippet to include `SELECT *` to ensure  the example contains valid, executable SQL.

### DIFF
--- a/docs/stable/sql/query_syntax/with.md
+++ b/docs/stable/sql/query_syntax/with.md
@@ -40,7 +40,7 @@ You can specify column names for CTEs:
 
 ```sql
 WITH cte(j) AS (SELECT 42 AS i)
-Select * FROM cte;
+SELECT * FROM cte;
 ```
 
 | j  |


vinayaksodar:
Please squash or only merge the latest commit the second commit just renamed Select to SELECT as that is the conventions used but I gave the commit the same name.

szarnyasg:
Hi, `SELECT`s are optional in DuckDB: https://duckdb.org/docs/stable/sql/query_syntax/from#from-first-syntax

So this is valid SQL:

```sql
WITH cte(j) AS (SELECT 42 AS i)
FROM cte;
```

```text
┌───────┐
│   j   │
│ int32 │
├───────┤
│    42 │
└───────┘
```

vinayaksodar:
Thanks for the clarification much appreciated.

PR #6310, opened 2025-12-16
add video

Tmonster:


szarnyasg:
Let's use a higher res version through blob storage.

PR #6326, opened 2025-12-19
Improve library placeholder images

Franz-Kafka:
- Use correct color scheme for library placeholder images
- Add missing placeholder SVGs


szarnyasg:
Thanks!

PR #6350, opened 2025-12-29
Fix order of primary key experiments in schema.md

nezda:
The prose order of the descriptions in Microbenchmark: The Effect of Primary Keys doesn't match the table of results.


szarnyasg:
Thanks!

PR #6370, opened 2026-01-06
Fix library thumbnails and placeholder color

Franz-Kafka:
- Updated several library thumbnails
- Fixed an issue in "Other Library Resources" using an incorrect placeholder color


szarnyasg:
Thanks!

PR #6392, opened 2026-01-12
docs: fix incorrect SQLNULL value in CSV docs

### LINKED ISSUE #6377: Incorrect value "CSV Import / `auto_type_candidates` Details" section
I think the document is a bit incorrect regard to type detection in `read_csv` function

### Problem

Page with incorrect `SQLNULL` value: https://duckdb.org/docs/stable/data/csv/overview#auto_type_candidates-details

Image

If we try to use this value in a query, it will break the query.

### Expected

The correct value is `NULL`: https://duckdb.org/docs/stable/data/csv/auto_detection#type-detection

Image

\* This problem exists since v1.0 doc (I think before 0.10.2, it is a valid keyword? : https://r.duckdb.org/news/index.html#features-0-10-2)

=> Should I correct all the versions except **0.10**?

harpreetseehra:
Fixes #6377

Change `SQLNULL` to `NULL` in the `auto_type_candidates` default value documentation.

The correct value is `NULL` as shown in the [auto_detection page](https://duckdb.org/docs/stable/data/csv/auto_detection#type-detection).

### DIFF
--- a/docs/stable/data/csv/overview.md
+++ b/docs/stable/data/csv/overview.md
@@ -135,7 +135,7 @@ Usage example:
 SELECT * FROM read_csv('csv_file.csv', auto_type_candidates = ['BIGINT', 'DATE']);
 ```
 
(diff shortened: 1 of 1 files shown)


harpreetseehra:
@szarnyasg

harpreetseehra:
This is one line change and shouldn't take much time to review. Thanks!

PR #6411, opened 2026-01-14
Remove redundant example

nateabele:
Remove redundant usage of 'omit read_parquet function' example. The example right below it restates the same thing but adds more detail. (Also the code doesn't match the description).


szarnyasg:
Thanks!

PR #6414, opened 2026-01-15
Improve contact form styling and cleanup SCSS

Franz-Kafka:
- Updated contact form styles on /foundation/support/
- Add "Organization" field
- Extract form styles into separate SCSS file
- Remove outdated form styles from foundation SCSS
- Normalize textbox padding to match code blocks


szarnyasg:
Thanks!

PR #6418, opened 2026-01-15
add first docs for unity catalog extension

benfleis:


szarnyasg:
Thanks, look great! I added it to the menu too.

PR #6430, opened 2026-01-18
Library submission: spanishoddata paper

e-kotov:
Paper about one of the first DuckDB-based R packages for data access.


szarnyasg:
Thanks!

Issue #6453, opened 2026-01-26
Azure write support for ADLS works, but documentation says otherwise


frantakalab:
Version: 1.4.4

[Here](https://github.com/duckdb/duckdb-web/blob/main/docs/stable/core_extensions/azure.md#writing-to-azure-blob-storage) it says:
> Writing to Azure Data Lake Storage (ADLS) via abfss:// paths is not yet supported. Only Azure Blob Storage (az:// or azure://) paths support writes.

But  I have just ran:
```
COPY 
	(select *
	from  table)
TO 'abfss://xxxx/table.parquet';
```

And it worked like a charm. 

Is this a documentation bug? I am happy to PR the change, but want to check first.

szarnyasg:
Hi, thanks for bringing this to our attention! I'm cc-ing @benfleis – Ben, can you please take a look and inform us about the state of support for `abfss://` writes? I found a closed PR here - https://github.com/duckdb/duckdb-azure/issues?q=abfss - but not sure about the support status.

benfleis:
Oops, got the code but forgot to update the docs to match. (The actual landed PR was https://github.com/duckdb/duckdb-azure/pull/140 btw.)

I'll put up a quick update right away!

benfleis:
Landed update, closing. Thanks @frantakalab !

frantakalab:
Perfect, thanks!

PR #6454, opened 2026-01-26
update azure write support note to include abfss

benfleis:
addresses #6453


docs/stable/core_extensions/azure.md
szarnyasg: @benfleis this should go too, right?

benfleis: yep, _too_ quick! I'll get it. :)

szarnyasg:
Thanks!

PR #6458, opened 2026-01-28
Document deprecation of recursive `UNION`s in `USING KEY` CTEs

timfi:
These are the relevant additions to document the deprecation of recursive `UNION`s for `USING KEY` CTEs in favor of recursive `UNION ALL`s—as discussed and implemented in https://github.com/duckdb/duckdb/pull/20689.


szarnyasg:
Thanks!

PR #6479, opened 2026-02-05
Update installation instructions for Arch Linux (AUR -> Extra)

the-solipsist:
Change installation instructions for DuckDB on Arch Linux to bring it in line with shift from AUR to Extra.


szarnyasg:
Thanks!

PR #6481, opened 2026-02-05
Improve storage version and compression docs

matthijskooijman:
Some small doc improvements, resulting from my quest to get compression working (see https://github.com/duckdb/duckdb/issues/20664).


szarnyasg:
Thanks!

PR #6498, opened 2026-02-10
Arrow api docs update

evertlammerts:
Related to https://github.com/duckdb/duckdb-python/pull/287

This PR:
* updates the stable docs to reflect what's currently in the API
* updates the preview docs to reflect what will be in there
* fixes scripts

Sorry if it's a bit much, I could also split it up if that works better.


evertlammerts:
Ah, this is an accumulated mess. I'll do this differently.

PR #6508, opened 2026-02-11
Add syntax highlighting colors to Design Manual

Franz-Kafka:
- Add table with syntax highlighting colors to design manual page


szarnyasg:
Thanks!

PR #6509, opened 2026-02-11
Warn users against the use of parametrized queries in .sql

evertlammerts:
Related to https://github.com/duckdb/duckdb-python/issues/238#issuecomment-3884098644

The performance of .sql with parameterized queries is so poor that we should discourage users from using it, and point them towards `.execute()` + replacement scan instead.


szarnyasg:
Thanks!

PR #6518, opened 2026-02-13
Apply the deprecation of `UNION` in `USING KEY` CTEs to Docs

timfi:
In #6458, I forgot to also switch the `USING KEY` examples in docs to using recursive `UNION ALL`s. Showing the deprecated recursive `UNION`s immediately after the big bold deprecation warning feels out of place to me.


szarnyasg:
Thanks!

PR #6538, opened 2026-02-23
Update .prompt command to show current database

Dtenwolde:
Noticed that with the `.prompt` command you now lose the useful information showing the current active database. Thought it would nice to add that in. Also played around with the colors and I think `yellow1` is closest to DuckDB's color. 
```
.prompt "{color:yellow1}{sql:select current_database()} ⚫◗ "
```

afbeelding


szarnyasg:
Good point, thank you!

PR #6550, opened 2026-03-02
document user-agent settings

benfleis:
Provide a quick summary of current HTTP user-agent settings coming from DuckDB and a couple extensions (azure, delta, httpfs and unity).


szarnyasg:
thanks!

PR #6562, opened 2026-03-06
Fix struct_update code example to use the struct_update function

stevenlw-porpoise:
Updating the code example.

Previously this accidentally invoked struct_insert in the example.

I checked the guidelines but did not see this documentation being auto generated. <https://github.com/duckdb/duckdb-web/blob/a787d0830d5d74221b8ad68cbfdddbefb860d8cc/CONTRIBUTING.md#generated-sql-function-lists>
Sorry if I missed anything.


szarnyasg:
Thanks!

PR #6575, opened 2026-03-10
Update search.js

Franz-Kafka:
- If the user typed a query before `search_data.json` finished loading, the search is now triggered automatically once the search data becomes available.


szarnyasg:
Thanks!

PR #6586, opened 2026-03-12
ODBC extension: add a warning about lazy eval

staticlibs:
This PR adds a warning to `odbc_copy` about its usage with Python Relational API.


szarnyasg:
Thanks!

Issue #6590, opened 2026-03-12
Python replacement scans with plain objects


rustyconover:
In 1.5.0 it seems we lost the ability to do replacement scans on plain python objects (if we ever could).

On this page it says:

https://duckdb.org/docs/stable/clients/python/data_ingestion

```
[Registering Objects](https://duckdb.org/docs/stable/clients/python/data_ingestion#registering-objects)
You can register Python objects as DuckDB tables using the [DuckDBPyConnection.register() function](https://duckdb.org/docs/stable/clients/python/reference/#duckdb.DuckDBPyConnection.register).
```

But when I do that with:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = ["duckdb==1.5.0"]
# ///
"""Demonstrate that DuckDB 1.5.0 cannot scan plain Python lists."""

import duckdb

con = duckdb.connect()

data = [
    {"name": "Alice", "age": 30},
    {"name": "Bob", "age": 25},
    {"name": "Charlie", "age": 35},
]

# Method 1: reference variable directly in SQL
print("--- SELECT * FROM data ---")
try:
    print(con.sql("SELECT * FROM data").fetchall())
except duckdb.InvalidInputException as e:
    print(f"FAILED: {e}")

# Method 2: register() then query
print("\n--- con.register('people', data) ---")
try:
    con.register("people", data)
    print(con.sql("SELECT * FROM people").fetchall())
except duckdb.InvalidInputException as e:
    print(f"FAILED: {e}")
```

I get

```
--- SELECT * FROM data ---
FAILED: Invalid Input Error: Python Object "data" of type "list" found on line "/Users/rusty/f.py:20" not suitable for replacement scans.
Make sure that "data" is either a pandas.DataFrame, duckdb.DuckDBPyRelation, pyarrow Table, Dataset, RecordBatchReader, Scanner, or NumPy ndarrays with supported format

--- con.register('people', data) ---
FAILED: Invalid Input Error: Python Object "people" of type "list" not suitable for replacement scans.
Make sure that "people" is either a pandas.DataFrame, duckdb.DuckDBPyRelation, pyarrow Table, Dataset, RecordBatchReader, Scanner, or NumPy ndarrays with supported format
```

evertlammerts:
[...]

PR #6599, opened 2026-03-13
Remove the register part of the python data ingestion docs

evertlammerts:
Ref: #6590

The register function is already discussed in the section above, the text that is removed in this pr was confusing.


szarnyasg:
Thanks!

PR #6606, opened 2026-03-15
Add DST transition documentation for TIMESTAMPTZ

rohitmannur007:
Fixes #20845

This PR adds documentation describing how DuckDB handles daylight saving
time (DST) transitions when adding calendar intervals to `TIMESTAMPTZ`
values.

During DST transitions, certain local timestamps may not exist. DuckDB
follows PostgreSQL behavior and adjusts the resulting timestamp forward
to the next valid time.

An example demonstrating this behavior has been added to the
`timestamptz` documentation.


szarnyasg:
@rohitmannur007 thanks for your contribution! I made a small fix and will merge it now.

PR #6607, opened 2026-03-16
Fix dark mode flash on page load

Franz-Kafka:
- Inline the color scheme detection instead of loading it as an external file
- Update _blog.scss


szarnyasg:
Thanks!

PR #6643, opened 2026-03-23
Add Lance to the list of supported Lakehouse formats

prrao87:
Adds Lance to the existing lakehouse format table now that it's been ported over to a core extension. cc @szarnyasg


szarnyasg:
Thanks!

PR #6670, opened 2026-03-31
meta data table in Library posts

Franz-Kafka:
- Add meta data table to Library posts
- Fix video width on iframes


szarnyasg:
Thanks!

PR #6701, opened 2026-04-07
Add section for viewing macros and table macros

yoonghm:
Added a section on viewing the list of macros and table macros with an example query.


szarnyasg:
Thanks!

PR #6704, opened 2026-04-07
Generate Metric Documentation

maiadegraaf:
This PR adds generated documentation based on the [metric_type.json](https://github.com/duckdb/duckdb/blob/main/src/common/enums/metric_type.json) file in duckdb core. 

The generated file: [metrics.html](https://github.com/user-attachments/files/26540180/metrics.html)

### DIFF
--- a/_data/menu_docs_current.json
+++ b/_data/menu_docs_current.json
@@ -1731,6 +1731,10 @@
               "url": "release_cycle"
             },
             {
+              "page": "Metrics",
+              "url": "metrics"
+            },
+            {
               "page": "Profiling",
               "url": "profiling"
             },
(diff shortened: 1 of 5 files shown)


guillesd:
Thanks @maiadegraaf ! This should run in CI so that we can redeploy new metrics without running manually. You can add it to a plugin like this one https://github.com/duckdb/duckdb-web/blob/061c365bb2836370a8824a8e116d204d455d581a/_plugins/search_index.rb#L11 and then it will run before building the content. Can you also check the linting errors? should be pretty easy to deal with looking at this config https://github.com/duckdb/duckdb-web/blob/061c365bb2836370a8824a8e116d204d455d581a/.github/workflows/lint.yml (maybe we should make this better with some pre-commit hooks)

PR #6736, opened 2026-04-15
docs: add Tigris to S3-compatible platforms and add import guide

davidmyriel:
Hi! We recently [wrote a blog](https://www.tigrisdata.com/blog/fifty-agents-one-bucket/) about DuckDB and I thought I'd shoot over some more info.

Small docs addition for Tigris, an S3-compatible object storage service.

Two changes, mirroring how Cloudflare R2 and Fastly Object Storage are already documented:

1. **`core_extensions/httpfs/s3api.md`** — adds Tigris next to the existing Cloudflare R2 reference in the "other services that implement the S3 API should also work" sentence. One word in a parenthetical, applied to both `current` and `lts`.

2. **`guides/network_cloud_storage/tigris_import.md`** — new page following the structure of `fastly_object_storage_import.md` and `cloudflare_r2_import.md`. Includes the `CREATE SECRET` snippet plus three notes I thought were worth calling out:
   - Why `REGION 'auto'` (single global endpoint, edge-routed — the value is required for SigV4 signing but not used for routing)
   - `URL_STYLE` does not need to be set (Tigris uses virtual-hosted-style, unlike Fastly which needs `path`)
   - A tip about Fly.io Machine colocation, since Tigris runs on Fly infrastructure

Menu entries added to `menu_docs_current.json` and `menu_docs_lts.json` alongside the existing Fastly entry.

Happy to fold the content into the existing `s3_import.md` under the "GCS and Cloudflare R2" section instead if you'd rather not add another per-vendor page — let me know what you prefer.

Thanks for taking a look!

### DIFF
--- a/_data/menu_docs_current.json
+++ b/_data/menu_docs_current.json
@@ -1371,6 +1371,10 @@
                 {
                   "page": "Fastly Object Storage Import",
                   "url": "fastly_object_storage_import"
+                },
+                {
+                  "page": "Tigris Import",
+                  "url": "tigris_import"
                 }
               ]
             },
--- a/_data/menu_docs_lts.json
+++ b/_data/menu_docs_lts.json
@@ -1349,6 +1349,10 @@
                 {
                   "page": "Fastly Object Storage Import",
                   "url": "fastly_object_storage_import"
+                },
+                {
+                  "page": "Tigris Import",
+                  "url": "tigris_import"
                 }
               ]
             },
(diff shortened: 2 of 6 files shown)


szarnyasg:
This looks great, thank you! There's no need to fold this into an existing page, I think the way forward for our docs it to 'unfold' the generic "S3-compatible storage" page into separate subpages.

PR #6741, opened 2026-04-16
Modify SQL query in explain.md

YanKleiner:
The example on the EXPLAIN page uses EXPLAIN ANALYZE but should be EXPLAIN, since the output shows estimated cardinalities (EC) and the accompanying text states the query is not actually executed.


szarnyasg:
Indeed. Thanks for the fix!

PR #6748, opened 2026-04-19
docs: fix key-value metadata link to include correct anchor

### LINKED ISSUE #6723: Issue found on page 'Reading and Writing Parquet Files': link for "key-value metadata" does not point to right anchor.
It reads:

"Write to Parquet file with key-value metadata" where "key-value metadata is a link. 
Currently that links to https://duckdb.org/docs/lts/data/parquet/metadata.html, which is the right page. 
But that page has an actual anchor for key value metadata:

https://duckdb.org/docs/lts/data/parquet/metadata#parquet-key-value-metadata

This destination would be more appropriate.

Page URL: <https://duckdb.org/docs/lts/data/parquet/overview>

avasis-ai:
## Summary

The "key-value metadata" link in the Parquet overview page currently points to `metadata.html` but should include the specific anchor `#parquet-key-value-metadata` for direct navigation. This matches the existing link on line 89 of the same file which already uses the correct anchor.

## Changes

Updated the link in `docs/current/data/parquet/overview.md` (write parquet section) to include the `#parquet-key-value-metadata` anchor.

Closes #6723

### DIFF
--- a/docs/current/data/parquet/overview.md
+++ b/docs/current/data/parquet/overview.md
@@ -260,7 +260,7 @@ COPY tbl
     (FORMAT parquet, COMPRESSION zstd, COMPRESSION_LEVEL 1);
 ```
 
-Write to Parquet file with [key-value metadata]({% link docs/current/data/parquet/metadata.md %}):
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks!

PR #6769, opened 2026-04-23
Add tooltips

Franz-Kafka:
tooltips via jekyll include:
`{% include tooltip.html label="Primary" text="Primary clients are the first to receive new features and are covered by community support." %}`

or from `_data/tooltips.yml`:

`{% include tooltip.html label="Primary" id="support_tier_primary" %}`

@szarnyasg please add support_tier_secondary to tooltips.yml


szarnyasg:
Thanks, I added the extra info

PR #6780, opened 2026-04-26
Update documentation for R-tree index example

ATL2001:
Corrected the comment to reflect the number of random points that are created.


szarnyasg:
Thanks

PR #6824, opened 2026-05-07
Add missing profiling mode `all`

taniabogatsch:
@szarnyasg - I hope I added it to the correct docs pages (`v1.5` and above line).


szarnyasg:
Great, thanks!

Issue #6833, opened 2026-05-10
Issue found on page 'Logging' 'Buffer size' section


joostas:
Incorrect example given with disable bufered logging:
>Or imagine you are debugging a crash in DuckDB and you want to use the file logger to understand what's going on: Simply disable the buffering using:
```
CALL enable_logging(storage_path = '/tmp/mylogs', storage_buffer_size = 2048);
```

If I understand correctly, the given example should be that:
```
CALL enable_logging(storage_path = '/tmp/mylogs', storage_buffer_size = 0);
```

szarnyasg:
Resolved via #6834

PR #6837, opened 2026-05-11
Chore logging docs

joostas:
Follow up on #6833 for LTS docs version
Additional fix to solve example formatting problem:
image


szarnyasg:
Thank you!

PR #6851, opened 2026-05-15
Fix Quack Stateless Query Example

ethagnawl:
- Add missing comma to stateless query example

Screenshot from 2026-05-15 12-13-53


szarnyasg:
Thanks!

PR #6862, opened 2026-05-19
duckdb sso auth implementation

codingeologist:
need to use SESSION_TOKEN instead of TOKEN when authenticating DuckDB to access AWS S3.

Provided a simple implementation of this in Python at the bottom of the documentation.


szarnyasg:
This is great, thanks a lot for the contribution!

PR #6887, opened 2026-05-30
Fix version number for ODBC in core_nightly

staticlibs:
Because ODBC scanner extension is portable between versions, we only build it with `v1.2.0` and publish to `core_nightly` under `v1.2.0` version. The same binary also can be published for other versions too, but currently there is no automation for that.

So this PR is changing `current_duckdb_version` placeholder to harcoded `v1.2.0`.


szarnyasg:
Thanks for the fix and the explanation!

PR #6902, opened 2026-06-07
Add QuackDB tertiary client

dannote:
Adds QuackDB to the tertiary clients list and renames the existing Elixir row to “Elixir (Duckdbex)” so the two Elixir clients are distinguishable in the table.

The two projects cover different integration architectures:

- Duckdbex embeds DuckDB in the BEAM through native C++/NIF bindings.
- QuackDB is an Elixir client for DuckDB's experimental Quack protocol. It is built around DBConnection/Ecto, supervised DuckDB processes, connection pooling, native append APIs, Explorer/Table.Reader integration, telemetry, and managed DuckDB binaries.

So this gives Elixir users both current options: embedded native bindings and the Quack protocol/client-driver approach.


szarnyasg:
Thanks!

PR #6928, opened 2026-06-16
DPhyp not DPccp algorithm

jgrg:
The linked paper is about `DPhyp` which improves on the previous `DPccp` algorithm. There are two mentions of `DPhyp` in comments in the DuckDB source, so I'm fairly certain it's the one which is used.


szarnyasg:
Thanks for the fix!

PR #6930, opened 2026-06-16
docs(quack): document client-side HTTP connection caching

npoggi:
Adds an HTTP Connection Caching sub-section to Client on the Quack main page. 

The `httpfs_connection_caching setting` is off by default, so users need to enable it manually to avoid avoiding a fresh TCP/TLS handshake per call and substantially lowering per-query latency.


szarnyasg:
Thanks!

Issue #6935, opened 2026-06-17
Issue found on page 'Numeric Types', section: Fixed-Point Decimals; document exact syntax


rpbouman:
The section "Fixed-Point Decimals" (https://duckdb.org/docs/lts/sql/data_types/numeric#fixed-point-decimals) on the numeric types page describes the syntax of a decimal type as DECIMAL(WIDTH, SCALE), where the keyword DECIMAL may also be NUMERIC. It also reads 

> The default WIDTH and SCALE is DECIMAL(18, 3), if none are specified.

This raises the question what syntax is acceptable if width and scale are unspecified. I found the following variants are accepted:

```
  select 123.45678::DECIMAL(18,3) as canonical
  , 123.45678::DECIMAL as minimal
  , 123.45678::DECIMAL() as missing_params
  , 123.45678::DECIMAL(18) as only_width
  , 123.45678::DECIMAL(18,) as missing_scale
  ;
```
with this result:
```
┌───────────────┬───────────────┬────────────────┬───────────────┬───────────────┐
│   canonical   │    minimal    │ missing_params │  only_width   │ missing_scale │
│ decimal(18,3) │ decimal(18,3) │ decimal(18,3)  │ decimal(18,0) │ decimal(18,0) │
├───────────────┼───────────────┼────────────────┼───────────────┼───────────────┤
│       123.457 │       123.457 │        123.457 │           123 │           123 │
└───────────────┴───────────────┴────────────────┴───────────────┴───────────────┘
```
I think this result justifies extra documentation.

- the ```minimal``` example, which omits both scale and with results in ```DECIMAL(18,3)```. This is already documented but it might be good to also document that this behavior is different from products like Postgres and Oracle, where such a decimal would be "unconstrained", that is, accept numeric values of any length, up to the implementation limits.
- when only specifying the width and omitting the scale (as in the ```only_width``` and ```missing_scale``` ), the resulting scale is ```0```. This requires documentation because the current documentation appears to imply that leaving out the scale result in the default of ```3```.
- The ```missing_params``` (empty parenthesis) and ```missing_scale``` (dangling comma) example are interesting in that they are rejected by other SQL dialects (tested MySQL, Oracle, Postgres).
[...]

PR #6963, opened 2026-06-22
docs: clarify DECIMAL default scale when only WIDTH is given

Michael-WhiteCapData:
Towards #6935. The docs state the default is `DECIMAL(18, 3)`, but that only applies when neither `WIDTH` nor `SCALE` is specified. If only `WIDTH` is given (e.g. `DECIMAL(18)`), `SCALE` defaults to `0` and fractional digits are truncated, which surprises users. Adds one clarifying sentence with an example.


szarnyasg:
Thanks! This is quite valuable, and I'm happy to merge it but I think fully resolving #6935 will require some further clarifications.

PR #6983, opened 2026-06-30
Add Parquet tip: sort rows before writing to improve row group pruning

CedricConday:
The "Tips for Writing Parquet Files" page documents `ROW_GROUP_SIZE` and notes that the reader uses row-group min/max metadata for predicate pushdown, but it does not mention the most effective way to take advantage of that pruning: **sorting the data on commonly-filtered columns before writing**.

Sorting makes each row group's min/max ranges tight and non-overlapping, so highly selective queries on those columns scan far fewer row groups. This comes up regularly (e.g. when compacting Parquet files by a key), so I added a short subsection with an example next to the existing row-group tips.

Targeted `docs/current`. Happy to also apply it to `docs/lts` if youd like.

---
*Disclosure: drafted with AI assistance (Claude Code); reviewed by me.*


szarnyasg:
This is great, thank you, and extra thanks for the AI disclosure.

PR #7021, opened 2026-07-15
Added PHP native PDO client (pdo_duckdb)

thomas-0816:
I created a native PHP extension in C to support the PHP PDO interface: https://github.com/thomas-0816/pdo-duckdb-php

It would be nice if it could be listed on the website under "Client APIs".

PDO is used by all big frameworks (Symfony/Doctrine, Laravel/Eloquent, etc.)

License: MIT

Supported DuckDB versions: 1.5.4+

Supported PHP versions: 8.2+

Supported operating systems: Linux, macOS, Windows
(Ubuntu 24+, Debian 12+, Fedora 42+, openSUSE 16+, AmazonLinux 2023+, Wolfi OS, Windows Server 2022+ (x64), macOS 14+ (arm64))


szarnyasg:
Hi @thomas-0816, thanks for the PR! A few things:

1) We only include clients in the [primary/secondary tier](https://duckdb.org/docs/current/clients/overview#support-tiers) if the core DuckDB team or close long-time contributors are maintaining them. This way, we can guarantee some degree of quality of services. Other clients are under the tertiary category.

2) In the tertiary category, we already have a [PHP client](https://duckdb.org/docs/current/clients/tertiary_clients/php). This is not a problem per se: there are also two Elixir clients. However, please make make a separate page for your clients and explain the differences from the existing PHP client.

Could you please implement these changes?

thomas-0816:
Thanks for the feedback, I moved the page to tertiary clients.

The differences between satur.io/duckdb and pdo_duckdb in short:

- native extension in C/C++, thread-safe, no FFI
- uses PDO interface (every application using PDO can directly use it next to mysql, postgresql, sqlite, etc.)
- support for 1.5.4+

szarnyasg:
Thanks for the changes. Looks great now!

thomas-0816:
Thank you!

PR #7027, opened 2026-07-18
fix: correct preserve_insertion_order instruction in llms.txt

arijitroy003:
The llms.txt file currently instructs to re-enable insertion order preservation with `SET preserve_insertion_order = false`, but this actually disables it. The correct statement to re-enable is `SET preserve_insertion_order = true`.


szarnyasg:
Indeed, thanks!

PR #7028, opened 2026-07-18
fix: use .length for jQuery element existence check

arijitroy003:
In js/script.js, the condition `$('.wrap.documentation') != 0` always evaluates to true because a jQuery object is never equal to the integer 0. This causes the content list cloning code to run on every page instead of only documentation pages. Changed to use `.length` which correctly checks whether matching elements exist.


szarnyasg:
Thanks!

PR #7029, opened 2026-07-18
fix: correct string slice and comparison in C API doc generator

arijitroy003:
Two bugs in `scripts/generate_c_api_docs.py`:

1. After the regex `finditer` loop, `function_prototype[last_pos:start]` uses the stale `start` value from the last match iteration, truncating the remaining prototype text. Changed to `function_prototype[last_pos:]`.

2. String group name comparison uses `is not` (identity check) instead of `!=` (value comparison). While CPython interns short strings making this work in practice, `!=` is the correct operator for value comparison.


szarnyasg:
Thanks!

PR #7030, opened 2026-07-18
Replace C language icon with Kernighan and Ritchie book cover C

jgrg:
Drawn in Inkscape from photo of my copy of K&R second edition, replacing the current C++ without the "++" icon.


szarnyasg:
Thanks, that's indeed the correct official logo!

jgrg:
I see the new icon is in use on the documentation page: https://duckdb.org/docs/current/
but not yet in the ecosystem diagram on the home page: https://duckdb.org/
I guess the ecosystem diagram would need to be rebuilt somehow?

szarnyasg:
Indeed that's a separate image. Thanks for reporting this. We'll address it.

PR #7037, opened 2026-07-20
Fix Link Checker workflow failing on transient external link errors

arijitroy003:
## Summary

The Link Checker workflow has been failing for 50+ consecutive daily runs. Every run since at least early June 2026 shows `failure` because `linkchecker` exits non-zero when *any* external link is broken — which is inevitable for a site linking to hundreds of third-party domains.

**Changes:**

- Add `continue-on-error: true` to the linkchecker step so broken external links are reported in the GitHub Actions step summary without failing the workflow
- Upgrade Python from 3.9 (EOL since October 2025) to 3.12
- Use `if: always()` for the summary step (idiomatic equivalent of the previous `success() || failure()`)
- Remove redundant `echo $?` after the linkchecker command

The HTML report is still generated and appended to the step summary on every run, so broken links remain visible and actionable — they just no longer block the workflow.

### DIFF
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -15,17 +15,17 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up Python 3.9
+      - name: Set up Python 3.12
         uses: actions/setup-python@v6
         with:
-          python-version: '3.9'
+          python-version: '3.12'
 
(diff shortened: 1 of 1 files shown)


szarnyasg:
Haha, fair enough. It's indeed been broken for a long time due to the sheer amount of links. Thanks for the PR.

PR #7038, opened 2026-07-20
Remove unused imports, fix no-op copies, and fix typo in scripts

arijitroy003:
## Summary

- Remove unused imports across several scripts: `csv` and `io` from `generate_config_docs.py`, `sys` from `copy_stable_to_preview.py`, `os` and `jinja2.optimizer.Optimizer` from `generate_profiling_docs.py`, `glob` from `redirect.py`
- Remove no-op `shutil.copy` calls where source and destination are identical (`_data/menu_docs_current.json` to itself) in `copy_stable_to_preview.py` and `archive_docs.py`
- Fix `CUMALATIVE` typo to `CUMULATIVE` in variable name in `generate_profiling_docs.py`
- Remove duplicate `python-frontmatter` entry from `requirements.txt`

## Test plan

- [x] Verified each issue exists in the current codebase before fixing
- [ ] Confirm no behavioral changes: unused imports are never referenced, no-op copies had identical src/dst, typo fix is internal variable naming only

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,5 +12,3 @@ icalendar
 
 ipython
 black
-
-python-frontmatter
--- a/scripts/archive_docs.py
+++ b/scripts/archive_docs.py
@@ -110,7 +110,6 @@ shutil.copy(
 )
 archive_stable(old_stable_version)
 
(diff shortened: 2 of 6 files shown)


szarnyasg:
Thanks!

PR #7041, opened 2026-07-20
Add SRI hashes to external CDN script and link tags

arijitroy003:
## Summary

Adds [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) hashes to all external CDN resources loaded in `_layouts/default.html`:

- `mapbox-gl.css` (v3.17.0-beta.1) from api.mapbox.com
- `mapbox-gl.js` (v3.17.0-beta.1) from api.mapbox.com
- `swiper-bundle.min.css` (@12) from cdn.jsdelivr.net
- `swiper-bundle.min.js` (@12) from cdn.jsdelivr.net

Each tag now includes `integrity="sha384-..."` and `crossorigin="anonymous"` attributes. This ensures the browser verifies the fetched resource against the expected hash before executing it, mitigating supply-chain attacks where a CDN serves tampered content.

`_layouts/docu.html` was also reviewed and contains no external CDN resources.

## Test plan

- [ ] Verify DuckCon pages load correctly (these are the only pages that use the affected resources, gated behind `page.body_class contains "duckcon"`)
- [ ] Confirm Mapbox GL map renders on DuckCon pages
- [ ] Confirm Swiper carousel functions on DuckCon pages
- [ ] Check browser console for SRI-related errors on non-DuckCon pages (there should be none, since these tags are conditionally included)

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -28,8 +28,8 @@
 	{% if page.body_class contains "duckcon" %}
 	<link rel="preconnect" href="https://api.mapbox.com" crossorigin>
 	<link rel="preconnect" href="https://cdn.jsdelivr.net" crossorigin>
-	<link href='https://api.mapbox.com/mapbox-gl-js/v3.17.0-beta.1/mapbox-gl.css' rel='stylesheet' />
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks!

PR #7043, opened 2026-07-21
Revises description column of the extensions summary table

pchemguy:
All extensions provide some functionality, so "Add support" only generates noise without providing any value.


szarnyasg:
That is actually very true. Thanks for the PR!

PR #7051, opened 2026-07-23
Add Julia repo URL

evertlammerts:
Ref: https://github.com/duckdblabs/duckdb-internal/issues/1148#issuecomment-5055568657

We took Julia out of tree, this adds the repo URL to the list in the docs. I don't see any other references to the repo that need updating.


szarnyasg:
Thanks!

PR #7063, opened 2026-07-28
Update Lakehouse Documentation

Tmonster:
Remove specific pages for S3Tables and Glue, instead have one general `Iceberg Rest Catalogs` page.


szarnyasg:
Lovely, thank you!

Issue #7064, opened 2026-07-29
Issue found on page 'List Functions': top section (toc) does not list `generate_subscripts()`


rpbouman:
`generate_subscripts()` is a list function, but it is not fully described on the list functions page: 

1) the table of function signatures at the top does not list it. (it was expected right after `generate_series(start[, stop][, step])`)
2) it does not have its own entry below the toc 
3) the functions page does describe it with an example in the section [Range Functions](https://duckdb.org/docs/lts/sql/functions/list#range-functions): https://duckdb.org/docs/lts/sql/functions/list#generate_subscriptsarr-dim

re 3: I think the placement as range function is not correct. I think the functions page should get a new mini-section "unnesting functions" that names both `unnest()` and `generate_subscripts()` and links to the page about the unnesting concept (https://duckdb.org/docs/lts/sql/query_syntax/unnest#keeping-track-of-list-entry-positions)

Page URL: <https://duckdb.org/docs/current/sql/functions/list.html>

PR #7074, opened 2026-07-30
change '/dev/null' to os.devnull to allow the script to run on windows.

rpbouman:
I never ran the generate_all_docs.sh script before but had to now for https://github.com/duckdb/duckdb-web/issues/7064.
Initially it failed running scripts/generate_config_docs.py. but by changing the hardcoded '/dev/null' to os.devnull I could get it to work. I figured I might as well share that.


szarnyasg:
Thanks!

PR #7078, opened 2026-07-30
doc: additional options to COPY

romulusFR:
Documentation of options LAYER_NAME (resp. -nln in ogr2ogr ), GEOMETRY_TYPE (-nlt), CRS and DATASET_CREATION_OPTIONS (-dsco) added. Explicit list support for -lco et -dsco. Analyzed from https://github.com/duckdb/duckdb-spatial/blob/v1.5-variegata/src/spatial/modules/gdal/gdal_module.cpp#L1441


szarnyasg:
Thanks!

PR #7081, opened 2026-07-30
Fix installation and shutdown docs for R

krlmlr:
The full set of alternative flavors is in https://duckdb.r-universe.dev/builds, though the "dev" flavors are not really for general use. Should we link that too?

@szarnyasg: Please take a look.


szarnyasg:
Yes, that seems a better alternative. Thanks a lot!

PR #7098, opened 2026-08-04
Use S3 URL in "Attach over S3" example

Flogex:
In https://github.com/duckdb/duckdb-web/commit/bda4e90895e655e6c9fddc92500d64575f29662d, the `s3://duckdb-blobs/...` URLs got replaced with `http://` URLs because the S3 bucket is not accessible.
However, I think that in the "Attaching to a Database over the S3 API" section, a `s3://` URL should be used in the example regardless.


szarnyasg:
That's fair enough, thank you!

PR #7100, opened 2026-08-05
Extend documentation for Iceberg ATTACH options

Tishj:
Fixes typo `Rest` -> `REST`

Prioritizes the `ATTACH` options over the Scan Options (rationale: most common use case will be attaching to catalog, not scanning a separate table)

Separate the giant list of `ATTACH` options into the categories they belong to, documenting each one individually.


szarnyasg:
Thanks for the PR! I did a manual and an AI-assisted pass on the formatting/grammar. I also restored `Gemfile.lock`, which should be committed into the repo.

Tishj:
Ah, alright
I figured Gemfile.lock was just a local file-level lock protecting the tracker of locally installed packages against concurrent access, weird name

PR #7108, opened 2026-08-06
Fix doubled words in tpcds and geometry docs

latent-9:
Two small doubled-word typos in the docs prose:

- `docs/current/core_extensions/tpcds.md`: "compatible with with TPC-DS version 4" -> "compatible with TPC-DS version 4".
- `docs/current/sql/functions/geometry.md`: "the functions for for examining and manipulating" -> "the functions for examining and manipulating".

Documentation only.


szarnyasg:
Thanks!

Issue #7109, opened 2026-08-06
Issue found on page 'Enum Functions'


rpbouman:
1) it reads:

> With the exception of enum_range_boundary, the result depends only on the type of the argument and not on its value.

but that's not entirely true: `enum_code( enum_value )` (https://duckdb.org/docs/current/sql/functions/enum#enum_codeenum_value) also returns a value that depends on the actual value, not its type.

2) for `enum_code()`: for completeness it would be good to mention that the codes are zero-based.

3) argument names: `enum_code()` is listed with an argument labeled `enum_value`. This is good! But the `enum_range_boundary()`, which also operates on the the argument values, labels its arguments as `enum`. It would make sense to align the labels. Since `eum_value` is the more descriptive name, that would IMO be better.

4) In extension to 3), it would reinforce the remark 1) about most enum functions operating on the type of the argument rather than its actual value if the arguments for those functions would be labeled `enum_value` instead.

Page URL: <https://duckdb.org/docs/current/sql/functions/enum.html>

PR #7110, opened 2026-08-06
Fixes some items mentioned in issue 7109 - (enum functions)

rpbouman:
Status of items mentioned in https://github.com/duckdb/duckdb-web/issues/7109

1) is addressed. "Most enum functions return a value that depends only on the type of the argument, and not on its value. The functions enum_code() and enum_range_boundary() are exceptions; they return a value that depends on the argument value, as well as its type." 

2) is addressed: "Returns the (0-based) numeric value backing the given enum value." 

3) and 4) remain unaddressed. Changing the argument labels would change the anchors and potentially links elsewhere in the documentation. Right now I'm not sure that is desired/worth it. Happy to pick it up if that is approved/desired.


szarnyasg:
Thanks!

PR #7117, opened 2026-08-10
docs: fix best-time window example (ASC + missing comma)

STATUS-Z:
## Summary
Data-analytics docs fix: \"best time to date\" should use `ORDER BY time ASC`, and the SELECT list was missing a comma after `time`.

Re-opened after an accidental fork-visibility change closed the prior PR.

## Test plan
- [ ] Example matches intended semantics (fastest time to date)
- [ ] SQL snippet is syntactically valid


szarnyasg:
HI @STATUS-Z, indeed this was returning an incorrect result. Thanks for the PR!

PR #7119, opened 2026-08-10
Document BETWEEN collation behavior

### LINKED ISSUE #6718: [duckdb/#18989] - BETWEEN and equivalent comparison operators produce inconsistent results with COLLATE NOCASE needs documentation
See https://github.com/duckdb/duckdb/issues/18989

Himanshuagrawal4:
## Summary

Document that an explicit collation on any input to `BETWEEN` applies to the entire expression, while separately written comparisons apply the collation only to the comparison containing it. Include a compact query demonstrating the different results.

Closes #6718.
Addresses duckdb/duckdb#18989.

## Validation

- verified the example result with DuckDB 1.5.5
- markdownlint
- `git diff --check`

## AI assistance

Codex (GPT-5) assisted with investigating the behavior and drafting the documentation. I manually reviewed the final diff and independently verified the SQL result.


szarnyasg:
Thanks!

PR #7126, opened 2026-08-12
Correct COPY multi-file option descriptions

matthewbayer:
## Summary

- correct the `COPY` multi-file option descriptions that currently claim they only apply with `PARTITION_BY`
- list `PER_THREAD_OUTPUT` and `FILE_SIZE_BYTES` as additional multi-file modes

These options are handled for directory-producing writes, including partitioned output, per-thread output and file rotation.

## Testing

- `npx markdownlint-cli docs/current/sql/statements/copy.md --config .markdownlint.jsonc`
- `vale docs/current/sql/statements/copy.md`
- `git diff --check`


szarnyasg:
Great, thanks for the clarification!

PR #7129, opened 2026-08-14
Mention that fmod and fdiv implement the floored definion

rsund:
Mention that fmod and fdiv implement the floored definion of modulo and integer devision (i.e. may give different results for negative arguments than the DuckDB default % and //). See duckdb/duckdb#2981, duckdb/duckdb#3030 and duckdb/duckdb#24761.


szarnyasg:
Thanks!

PR #7131, opened 2026-08-14
Fix nested comparison semantics description.

hawkfish:
I can't get local serving to work, so hope this is good!


szarnyasg:
Thanks!

PR #7132, opened 2026-08-15
Fix duckdb_functions() docs to match returned columns

daviddallakyan2005:
The `duckdb_functions()` docs were out of date against core `src/function/table/system/duckdb_functions.cpp`.

The functions overview said descriptions and parameter names are not available, but the sample result table on that same page already shows both, and the table function binds and emits `description` and `parameters`.

The metadata page said `description` is always `NULL` and omitted `alias_of`, `extension_name`, and `categories`. Those columns are bound and populated; empty description, alias, or extension name is returned as `NULL`.

## How to test

```bash
npx markdownlint-cli docs/current/sql/functions/overview.md docs/current/sql/meta/duckdb_table_functions.md --config .markdownlint.jsonc
```

Both pages are hand-written `docs/current/` (no generator markers).


szarnyasg:
@daviddallakyan2005 thanks for the PR. Let's split this into two: these features will be shipped in v2.0, while the rest of this PR is already true for v1.5.5:

```markdown
| `extension_name` | The name of the extension that provides this function, or `NULL` if none. | `VARCHAR` |
| `categories` | Categories of the function. Used to generate the documentation. | `VARCHAR[]` |
```

PR #7138, opened 2026-08-17
Update post card

Franz-Kafka:
- Post card update: reading time instead of tags
- Small CSS adjustments
- SCSS sync


szarnyasg:
Thanks!

PR #7144, opened 2026-08-18
New Blog Post: Chunked Query Results in DuckDB Java Driver

geertjanw:
This pull request adds a new blog post that introduces and explains the new chunked query results API in the DuckDB Java driver. The post details how the new API allows Java applications to efficiently fetch query results as columnar data chunks, bypassing the traditional JDBC row-at-a-time interface and its associated overhead. It also provides example code, highlights the benefits and current limitations, and points readers to further resources.

New blog post: Chunked Query Results in DuckDB Java Driver

**Overview and Feature Introduction**
* Introduces `DuckDBChunkedResult`, a new API in the DuckDB Java driver that allows fetching query results as a sequence of lazily loaded columnar data chunks, aligning with DuckDB's internal vectorized execution model.
* Explains the inefficiencies of the traditional JDBC `ResultSet` row-at-a-time interface and how the new API sidesteps these issues for large result sets and columnar destinations.

**Usage and Practical Guidance**
* Provides a code example demonstrating how to use the new API, including iterating over chunks and vectors, and clarifies index conventions (1-based for parameters, 0-based for columns/rows).
* Outlines current limitations: support for only basic scalar types and requirement to use prepared statements (no direct query string overload yet).


_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
...
+
+JDBC, on the other hand, was designed in 1997 around a very different idea. The `ResultSet` API hands you one row at a time (`next()`) and one value at a time (`getInt(1)`, `getString(2)`, and so on). It is a solid API, familiar to every Java developer and supported by every tool in the ecosystem, but it forces a shape onto the data that DuckDB never had on the inside.
+
+For the DuckDB Java driver, that mismatch has a concrete cost. The driver embeds the native DuckDB library and talks to it over JNI. When the engine has already produced a columnar chunk of 2,048 rows, the JDBC specification requires the driver to slice it back into rows and cells before you can read it. If your application is going to put those values right back into columnar form, such as arrays, Arrow buffers, or a machine learning feature matrix, everyone involved is doing work that nobody asked for.
+
+Starting with recent releases of the [Java driver](https://github.com/duckdb/duckdb-java/pull/682), there is a way out.
```
szarnyasg: Which is the first version where this is available?

szarnyasg:
Good stuff! Let's schedule this for Friday (2026-08-21)

_posts/2026-08-19-chunked-query-results-java-driver.md
staticlibs: First version that included that was `1.5.3.0`, release in May.

_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
+---
+layout: post
+title: "Chunked Query Results in the DuckDB Java Driver"
+author: "Alex Kasko, Geertjan Wielenga"
```
staticlibs: I suggest to swap the order of the authors.

_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
...
+excerpt: "The DuckDB Java driver can now return query results as a lazily fetched sequence of columnar data chunks, sidestepping JDBC's row-at-a-time ResultSet and its per-value overhead."
+tags: ["deep dive"]
+thumb: "/images/blog/thumbs/java.svg"  # TODO: thumbnail
+image: "/images/blog/thumbs/java.png"  # TODO: social image
+---
+
+DuckDB is a columnar, vectorized database. Every operator inside the engine works on *data chunks*: batches of column vectors, up to 2,048 rows at a time. This is a big part of why DuckDB is fast, because the engine amortizes interpretation overhead over thousands of values instead of paying it once per value.
```
staticlibs: Suggest to add a link to [Data Chunks C API page](https://duckdb.org/docs/current/clients/c/data_chunk), JDBC uses this C API to work with data chunks.

_posts/2026-08-19-chunked-query-results-java-driver.md
szarnyasg:
```suggestion
Starting with version 1.5.3.0 of the [Java driver](https://github.com/duckdb/duckdb-java/pull/682), there is a way out.
```

_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
...
+
+JDBC, on the other hand, was designed in 1997 around a very different idea. The `ResultSet` API hands you one row at a time (`next()`) and one value at a time (`getInt(1)`, `getString(2)`, and so on). It is a solid API, familiar to every Java developer and supported by every tool in the ecosystem, but it forces a shape onto the data that DuckDB never had on the inside.
+
+For the DuckDB Java driver, that mismatch has a concrete cost. The driver embeds the native DuckDB library and talks to it over JNI. When the engine has already produced a columnar chunk of 2,048 rows, the JDBC specification requires the driver to slice it back into rows and cells before you can read it. If your application is going to put those values right back into columnar form, such as arrays, Arrow buffers, or a machine learning feature matrix, everyone involved is doing work that nobody asked for.
```
staticlibs: The description is correct. Just if we want to make it slightly more deep - standard JDBC ResultSet is prepared for consumption per-chunk (so no JNI calls per row), but to conform with JDBC requirements some of data types require untrivial work, for example when user-specified `Calendar` [is used to fetch timestamps](https://github.com/duckdb/duckdb-java/pull/166).

Besides that, to support multiple result sets on the same connection, the [streaming](https://duckdb.org/docs/current/clients/java#streaming-results) is disabled by default, so the whole result is read into the memory by default (most users not aware about `jdbc_stream_results` option).

Additionally, extensive result set metadata is prepared, it is done per result (not per row), but still adds some overhead that is not needed in most cases.

_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
...
+try (ResultSet rs = stmt.executeQuery("SELECT a, b FROM measurements")) {
+    while (rs.next()) {
+        long a = rs.getLong(1);
+        double b = rs.getDouble(2);
+        // ...
+    }
+}
+```
+
+For every row, the driver needs to track a cursor position. For every cell, it needs to locate the value inside the current native chunk, convert it, and hand it across the JNI boundary in a form the JDBC interface allows. None of these steps is expensive on its own. All of them together, multiplied by millions of rows times a handful of columns, add up to real time, and the profile of that time is exactly the per-value interpretation overhead that vectorized execution was invented to eliminate.
```
staticlibs: Perhaps some details from the previous comment can be moved into this paragraph. Also this bit `locate the value inside the current native chunk, convert it, and hand it across the JNI boundary` (only applies to non-primitive fields) may be better to rephrase it to something like: `locate the value inside the current native chunk and convert it to a Java representation - for some of non-primitive types such conversion requires to cross the JNI boundary for every value`.

_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
...
+
+The Java driver now exposes the engine's native chunk stream directly, via the same mechanism the [C API](https://duckdb.org/docs/current/clients/c/overview) offers as `duckdb_fetch_chunk`. A query result becomes a lazily fetched sequence of data chunks, and you read column vectors out of each chunk in batches, the way the engine produced them, avoiding the per-row overhead mandated by the JDBC specification.
+
+Here is the whole flow:
+
+```java
+try (DuckDBConnection conn = DriverManager
+        .getConnection("jdbc:duckdb:")
+        .unwrap(DuckDBConnection.class);
+     DuckDBPreparedStatement ps = conn.prepare("SELECT ? AS col1")) {
```
staticlibs: I suggest to use a Parquet file query here. For example, some TPC-H data. Not sure, whether the numbers comparison may be useful here, or just  to make it clear that the query is to the big Parquet.

Also it may (or may not) be useful to add a reference to the [async blog post](https://duckdb.org/2026/07/31/asynchronous-io) here - by using an example of reading Parquet from S3, first with a stable 1.5.5.1 JAR version, and then to repeat it with 2.0-alpha snaphot version and mention that now async read is performed automatically.

_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
...
+                }
+            }
+        }
+    }
+}
+```
+
+A few things are to point out:
+
+- **It is lazy.** `nextChunk()` pulls one chunk at a time from the engine. The full result is never materialized on the Java side, so you can stream through results far larger than your heap, the same property that makes the engine itself happy to produce them.
```
staticlibs: > never materialized on the Java side

With standard JDBC by default it is materialized fully only on native side. So here perhaps rephrase as `never materialized neither on native nor on Java side`.

_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
...
+
+- **It is columnar.** Inside a chunk you work vector by vector. If your destination is columnar too, such as a `long[]`, an Arrow `VectorSchemaRoot`, or a Parquet writer, you copy values in tight, monomorphic loops instead of bouncing between columns on every row.
+
+- **It is familiar if you have written a UDF.** Chunk contents are accessed through the same `DuckDBDataChunkReader` API that the driver's [user-defined functions](https://github.com/duckdb/duckdb-java/blob/main/UDF.MD) use to read their input vectors. Learn the API once and use it on both sides: reading function arguments inside a UDF, and reading query results outside of one.
+
+- **Watch your indices.** In keeping with JDBC tradition, statement *parameters* remain 1-based. Chunk *columns and rows* are 0-based, matching the C API and the UDF interfaces. This is a difference to keep in mind in both directions.
```
staticlibs: Suggest also to mention (in this section, not in limitations) that a minimal metadata - column types and column count is available.

_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
...
+            // get the current chunk from the result
+            DuckDBDataChunkReader chunk = res.chunk();
+
+            // iterate over the chunk columns, all indices are 0-based
+            for (long col = 0; col < chunk.columnCount(); col++) {
+
+                // get a vector for the specified column
+                DuckDBReadableVector vector = chunk.vector(col);
+
+                // iterate over vector rows
```
staticlibs: Perhaps a Java Stream example of iterating over the `Vector` can be used here, and mentioning a fork-join pool and that parallel stream on 2048 rows may be not useful. Some info about Streams was available in [UDFs description](https://github.com/duckdb/duckdb-java/blob/main/UDF.MD#advanced-api-duckdbscalarfunction).

_posts/2026-08-19-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,99 @@
...
+
+- **It is familiar if you have written a UDF.** Chunk contents are accessed through the same `DuckDBDataChunkReader` API that the driver's [user-defined functions](https://github.com/duckdb/duckdb-java/blob/main/UDF.MD) use to read their input vectors. Learn the API once and use it on both sides: reading function arguments inside a UDF, and reading query results outside of one.
+
+- **Watch your indices.** In keeping with JDBC tradition, statement *parameters* remain 1-based. Chunk *columns and rows* are 0-based, matching the C API and the UDF interfaces. This is a difference to keep in mind in both directions.
+
+## Current Limitations
+
+This is the first iteration of the API, and there are two limitations you should know about before adopting it:
+
+- **Basic data types only, for now.** The scalar types are supported. Composite types (`LIST`, `STRUCT`) are not yet readable through the chunked interface. Support for them is planned for a future release.
```
staticlibs: Looking at the reader API now, it looks too limited for no particular reason, we really should allow to read `VARCHAR`s as UTF-8 `byte[]`s without converting them to `String`s. Perhaps add a hint here, that improvements to the reader are going to be added.

staticlibs (approved):
Looks good to me! Added a few minor comments.

_posts/2026-08-19-chunked-query-results-java-driver.md
staticlibs: > Not sure, whether the numbers comparison may be useful here

Looking more on the current state, I think it may be better to skip the numbers, as they may look disappointing. The implementation is doing the right things (and is done in the right direction), just with current state the performance may be pretty close to the standard JDBC case. For example, we can read `STRUCT`s or `ARRAY`s very quickly in this case (without `Object[]` representation used in standard JDBC), but currently `STRUCT`s/`ARRAY`s are not supported. We can easily batch-read the `VARCHAR`s (in a single huge UTF-8 buffer - so minimal number of JNI-calls/objects/allocations), but we currently have such kind of batching only in the writer (in UDFs - duckdb/duckdb-java#813). It is still great to have some users to try this API, just we may expect some complains in the feedback.

_posts/2026-08-21-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,108 @@
...
+layout: post
+title: "Chunked Query Results in the DuckDB Java Driver"
+author: "Geertjan Wielenga, Alex Kasko"
+excerpt: "The DuckDB Java driver can now return query results as a lazily fetched sequence of columnar data chunks, avoiding JDBC's row-at-a-time ResultSet and its per-value overhead."
+tags: ["deep dive"]
+thumb: "/images/blog/thumbs/java.svg"
+image: "/images/blog/thumbs/java.png"
+---
+
+DuckDB is a columnar, vectorized database. Every operator inside the engine works on [*data chunks*]({% link docs/current/clients/c/data_chunk.md %}): batches of column vectors, up to 2,048 rows at a time. That is a big part of why DuckDB is fast: the engine amortizes interpretation overhead over thousands of values instead of paying it once per value.
```
staticlibs: > up to 2,048 rows at a time

Lets change this to "2,048 rows at a time"

_posts/2026-08-21-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,108 @@
...
+
+JDBC, on the other hand, was designed in 1997 around a very different idea. The `ResultSet` API hands you one row at a time (`next()`) and one value at a time (`getInt(1)`, `getString(2)`, and so on). It is a solid API, familiar to every Java developer and supported by every tool in the ecosystem, but it forces the data into a form that DuckDB never used internally.
+
+The DuckDB Java driver has to bridge the two. It embeds the native DuckDB library and talks to it over JNI. When the engine has already produced a columnar chunk of 2,048 rows, the JDBC specification requires the driver to slice it back into rows and cells before you can read it. If your application is going to put those values right back into columnar form, such as arrays, Arrow buffers, or a machine learning feature matrix, the data is split into rows only to be reassembled into columns, wasting work on both sides.
+
+Starting with version 1.5.3.0 of the [Java driver](https://github.com/duckdb/duckdb-java/pull/682), there is an alternative.
```
staticlibs: I suggest to remote the PR link (leaving only the version number), it does not have any additional information to the one in the article.

_posts/2026-08-21-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,108 @@
...
+try (ResultSet rs = stmt.executeQuery("SELECT a, b FROM measurements")) {
+    while (rs.next()) {
+        long a = rs.getLong(1);
+        double b = rs.getDouble(2);
+        // ...
+    }
+}
+```
+
+For every row, the driver tracks a cursor position. For every cell, it locates the value inside the current native chunk and converts it to a Java representation. For some non-primitive types that conversion has to cross the JNI boundary for every value, and a few conversions take extra work to get right, such as fetching a timestamp against a caller-supplied `Calendar`. None of these steps is expensive on its own. Multiplied by millions of rows times a handful of columns, they add up, and this is exactly the per-value interpretation overhead that vectorized execution is designed to avoid.
```
staticlibs: > such as fetching a timestamp against a caller-supplied `Calendar`

Suggest to add here that this is mandated by the JDBC spec. Maybe add a note (as this targeted to Java users), that we are talking about ancient `java.util.Calendar` that predates `java.time`.

_posts/2026-08-21-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,108 @@
...
+
+The Java driver now exposes the engine's native chunk stream directly, via the same mechanism the [C API]({% link docs/current/clients/c/overview.md %}) offers as `duckdb_fetch_chunk`. A query result becomes a lazily fetched sequence of data chunks, and you read column vectors out of each chunk in batches, the way the engine produced them, avoiding the per-row overhead required by the JDBC specification.
+
+Here is the whole flow:
+
+```java
+try (DuckDBConnection conn = DriverManager
+        .getConnection("jdbc:duckdb:")
+        .unwrap(DuckDBConnection.class);
+     DuckDBPreparedStatement ps = conn.prepare("SELECT ? AS col1")) {
```
staticlibs: > SELECT ? AS col1

I think it may be better to change the query to something like:

`FROM s3://my-bucket-name/lineitems.parquet`

_posts/2026-08-21-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,108 @@
...
+
+    ps.setInt(1, 42); // statement parameters are still 1-based
```
staticlibs: Suggest to drop the parameter from the example, as it is a part of standard JDBC - not a new API.

_posts/2026-08-21-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,108 @@
...
+
+A few properties to note:
+
+* **Laziness.** `nextChunk()` pulls one chunk at a time from the engine. The full result is never materialized, neither on the native side nor on the Java side, so you can stream through results far larger than your heap, the same way the engine itself produces them.
+
+* **Columnar access.** Inside a chunk you work vector by vector. If your destination is columnar too, such as a `long[]`, an Arrow `VectorSchemaRoot`, or a Parquet writer, you copy values in tight, monomorphic loops instead of bouncing between columns on every row.
+
+* **Enough metadata to dispatch.** Each result still reports its column count and column types, so you can pick the right `get*` call per vector without carrying a separate schema description alongside the data.
+
+* **A familiar API if you have written a UDF.** Chunk contents are accessed through the same `DuckDBDataChunkReader` API that the driver's [user-defined functions](https://github.com/duckdb/duckdb-java/blob/main/UDF.MD) use to read their input vectors. Learn the API once and use it on both sides: reading function arguments inside a UDF, and reading query results outside of one.
```
staticlibs: Lets change the `[user-defined functions]` link to the Java page section on duckdb.org

_posts/2026-08-21-chunked-query-results-java-driver.md
```diff
@@ -0,0 +1,108 @@
...
+
+* **Index bases differ.** In keeping with JDBC tradition, statement *parameters* remain 1-based. Chunk *columns and rows* are 0-based, matching the C API and the UDF interfaces, so keep the difference in mind in both directions.
```
staticlibs: Lets remove parameters mentions from here too.

staticlibs (approved):
Looks good to me! Added a few nits.

PR #7157, opened 2026-08-19
New & updated Java JDBC content

### LINKED ISSUE #5301: Improve JDBC documentation
See https://github.com/duckdb/duckdb-java/issues/203

geertjanw:
Fixes #5301 # Improve JDBC documentation
Fixes #7165 # A few notes to new Java page

Documents duckdb/duckdb-java#113   # Parquet binary column read as string (already closed upstream)
Documents duckdb/duckdb-java#180   # GraalVM Native Image limitation (documented, not fixed)
Documents duckdb/duckdb-java#203   # clean shutdown / WAL checkpoint behavior (already closed upstream)
Documents duckdb/duckdb-java#222   # schemas not visible (already closed upstream)
Documents duckdb/duckdb-java#239   # DuckLake instance-caching nuance (already closed upstream)
Documents duckdb/duckdb-java#252   # session_init_sql_file option (already closed upstream)
Documents duckdb/duckdb-java#682   # chunked result API (DuckDBChunkedResult) (already closed upstream)

This pull request significantly expands and improves the DuckDB Java (JDBC) client documentation. It adds detailed explanations of advanced features, connection properties, type mappings, and new APIs, aiming to make the documentation much more comprehensive and user-friendly for Java developers. The most important changes are grouped below:

**Connection Management and Properties**

- Added instructions for using preview/nightly JDBC builds and how to add the staging Maven repository.
- Documented new and existing JDBC connection properties in a table, including `duckdb.read_only`, `access_mode`, `custom_user_agent`, `jdbc_stream_results`, and others, with explanations on how to use them via the `Properties` object or JDBC URL.
- Clarified the shutdown process for DuckDB databases, including checkpointing behavior, the role of the write-ahead log, and best practices for closing connections.
- Explained the interaction between `duckdb.read_only` and `access_mode`, including error behavior if they disagree.

**Advanced Query and Data Handling**

- Added a section on reading DuckDB's nested and composite types via JDBC, mapping each DuckDB type to its Java representation and providing code examples.
- Introduced the chunked results API (`DuckDBChunkedResult`), which allows efficient, columnar, lazily-fetched query results, including usage details and code samples.
- Improved documentation on the `Appender` API, including how to append nested/collection types and use new methods like `beginStruct`, `endStruct`, and handling of `LIST`/`ARRAY` columns.

**Monitoring, Profiling, and Query Control**

- Added sections on query progress reporting and cancellation (using `Statement.cancel()` and `DuckDBPreparedStatement.getQueryProgress()`), and on profiling queries with `getProfilingInformation` and the `ProfilerPrintFormat` enum.
- Documented memory monitoring using Java Flight Recorder (JFR), explaining how to enable it via JDBC properties and interpret the resulting events.

**User-Defined Functions (UDFs) in Java**

- Added comprehensive documentation for registering Java scalar and table UDFs, including code examples for both row-wise and vectorized scalar functions, table function lifecycle (bind/init/apply), resource cleanup, parallel execution, and type mapping between Java and DuckDB types.

These improvements make the documentation a much more complete reference for developers using DuckDB from Java, covering both basic usage and advanced features.


docs/current/clients/java.md
```diff
@@ -18,6 +18,8 @@ title: Java (JDBC) Client
 
 The DuckDB Java JDBC API can be installed from [Maven Central](https://search.maven.org/artifact/org.duckdb/duckdb_jdbc). Please see the [installation page]({% link install/index.html %}?environment=java) for details.
 
+To try features before they reach a stable release, preview (nightly) builds of the JDBC driver are published as `SNAPSHOT` versions to the Sonatype Central snapshots repository at `https://central.sonatype.com/repository/maven-snapshots/`. Add this repository to your build and depend on a `duckdb_jdbc` `SNAPSHOT` version. See the [preview builds page]({% link install/preview.md %}) for a full Maven example.
```
staticlibs: This is no longer the case, we publish shapshots on duckdb-staging.duckdb.org now, just we don't have a browseable page of published versions, some versions are listed here - https://github.com/duckdblabs/duckdb-internal/issues/9742#issuecomment-5309838057

docs/current/clients/java.md
```diff
@@ -80,6 +96,24 @@ connectionProperties.setProperty("temp_directory", "/path/to/temp/dir/");
...
+<div class="monospace_table"></div>
+
+| Property | Description |
+|--|--|
+| `duckdb.read_only` | Open the database in read-only mode. |
+| `access_mode` | Standard DuckDB access mode: `READ_ONLY`, `READ_WRITE`, or `AUTOMATIC`. |
+| `custom_user_agent` | Append a custom string to the user agent reported to DuckDB. |
+| `jdbc_stream_results` | Stream result sets instead of materializing them. See [Streaming Results](#streaming-results). |
+| `jdbc_auto_commit` | Set the default auto-commit mode for new connections. |
+| `jdbc_pin_db` | Keep the database instance alive after its last connection closes. Defaults to `true` for in-memory databases. |
```
staticlibs: > Defaults to `true` for in-memory databases.

This bit is incorrect, it is `false` by default for in-memory databases. It defaults to `true` for DuckLake URLs  - duckdb/duckdb-java#239

docs/current/clients/java.md
```diff
@@ -310,3 +666,15 @@ Exception in thread "main" java.lang.ClassNotFoundException: org.duckdb.DuckDBDr
 ```
 
 These errors stem from the DuckDB Maven/Gradle dependency not being detected. To ensure that it is detected, force refresh the Maven configuration in your IDE.
+
+### Parquet String Column Returns a Blob
```
staticlibs: Perhaps it may be better to move this section to some other page, as does not seem to be related to Java.

docs/current/clients/java.md
```diff
@@ -284,6 +414,232 @@ stmt.executeBatch();
...
+DuckDBConnection conn = (DuckDBConnection) DriverManager.getConnection("jdbc:duckdb:");
+try (Statement stmt = conn.createStatement()) {
+    stmt.execute("PRAGMA enable_profiling = 'json'");
+    stmt.executeQuery("SELECT count(*) FROM range(1000)").close();
+    String profile = conn.getProfilingInformation(ProfilerPrintFormat.JSON);
+    System.out.println(profile);
+}
+```
+
+### Memory Monitoring with JFR
```
staticlibs: We did not touch JFR in the discussion yesterday, I think it can be a good topic for more detailed description (like separate page or a separate blog post). As it is both Java-specific, obscure and can be useful in Java apps. Just a point for future.

docs/current/clients/java.md
```diff
@@ -284,6 +438,232 @@ stmt.executeBatch();
...
```
staticlibs: We did not touch JFR in the discussion yesterday, I think it can be a good topic for more detailed description (like separate page or a separate blog post). As it is both Java-specific, obscure and can be useful in Java apps. Just a point for future.

staticlibs (approved):
Looks good to me! Left a few comments

docs/current/clients/java.md
geertjanw: I originally had that URL, @staticlibs: duckdb-staging.duckdb.org but it doesn't resolve.

geertjanw: I.e., when I go to https://duckdb-staging.duckdb.org/duckdb/duckdb-java/maven in my browser, there's a 404.

staticlibs: This is our S3 (R2) object storage, there is no web page over it (perhaps should be added), but Maven can consume it.

docs/current/clients/java.md
```diff
@@ -284,6 +414,232 @@ stmt.executeBatch();
...
+
+For long-running queries, `DuckDBPreparedStatement.getQueryProgress()` returns a `QueryProgress` snapshot while the query executes. It is intended to be called from a separate thread and exposes `getPercentage()`, `getRowsProcessed()`, and `getTotalRowsToProcess()`.
+
+```java
+DuckDBPreparedStatement ps =
+    (DuckDBPreparedStatement) conn.prepareStatement("SELECT count(*) FROM range(1_000_000_000)");
+
+// from a monitoring thread, while execute() runs on another thread
+QueryProgress progress = ps.getQueryProgress();
+System.out.println(progress.getPercentage());
```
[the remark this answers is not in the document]
geertjanw: Fixed. The example now runs the query on a background CompletableFuture and polls getQueryProgress() in a loop while it executes, and it includes the imports. It lives on the new profiling.md page after the page split. (Kept the (DuckDBPreparedStatement) cast for consistency with the rest of the page rather than unwrap.)

docs/current/clients/java.md
```diff
@@ -284,6 +414,232 @@ stmt.executeBatch();
...
+Cast a connection to `DuckDBConnection` and call `getProfilingInformation(ProfilerPrintFormat)` to retrieve profiling output for the most recent query on that connection, after enabling profiling with `PRAGMA enable_profiling`. The `ProfilerPrintFormat` enum selects the output format: `DEFAULT`, `TEXT`, `QUERY_TREE`, `QUERY_TREE_OPTIMIZER`, `NO_OUTPUT`, `JSON`, `HTML`, `GRAPHVIZ`, `YAML`, and `MERMAID`.
+
+```java
+DuckDBConnection conn = (DuckDBConnection) DriverManager.getConnection("jdbc:duckdb:");
+try (Statement stmt = conn.createStatement()) {
+    stmt.execute("PRAGMA enable_profiling = 'json'");
+    stmt.executeQuery("SELECT count(*) FROM range(1000)").close();
+    String profile = conn.getProfilingInformation(ProfilerPrintFormat.JSON);
+    System.out.println(profile);
+}
```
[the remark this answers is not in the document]
geertjanw: Fixed. Both the profiling and the query-progress examples now use try-with-resources for the connection and statement, on the new profiling.md page.

docs/current/clients/java.md
geertjanw: Updated the Java overview page to point at https://duckdb-staging.duckdb.org/duckdb/duckdb-java/maven and noted it's an object-storage repo with no browseable web page, consumed directly by Maven. Thanks for the pointer.

docs/current/clients/java.md
geertjanw: Corrected. The table now reads: "Disabled by default. Enabled by default, together with jdbc_stream_results, for ducklake: URLs …" — matching duckdb/duckdb-java#239.

docs/current/clients/java/overview.md
```diff
@@ -0,0 +1,56 @@
...
+In JDBC, database connections are created through the standard `java.sql.DriverManager` class, using the `jdbc:duckdb:` JDBC URL prefix:
+
+```java
+import java.sql.Connection;
+import java.sql.DriverManager;
+
+Connection conn = DriverManager.getConnection("jdbc:duckdb:");
+```
+
+Used on its own, `jdbc:duckdb:` opens an in-memory database. Appending a file name, for example `jdbc:duckdb:/tmp/my_database`, opens a persistent database instead.
```
nantiamak, line 48: After opening the connection, would it be useful here to run a very simple query and mention what is the expected result as a way to verify that installation completed successfully?

docs/current/clients/java/connecting.md
```diff
@@ -0,0 +1,294 @@
...
+---
+
+## Overview
+
+Every DuckDB feature reachable from Java starts with a `Connection`. This page covers the JDBC URL forms the driver accepts, how DuckDB and driver options are passed and which of them win, how connections come to share (or not share) an underlying database instance, how DuckDB's threads relate to that instance, and how to reach a remote DuckDB server over [Quack]({% link docs/current/quack/overview.md %}).
+
+## Opening a Connection
+
+In JDBC, database connections are created through the standard `java.sql.DriverManager` class.
+The driver should auto-register in the `DriverManager`. If that does not work for some reason, you can enforce registration using the following statement:
```
nantiamak: This is probably Java specific and Java users would understand it. Myself I don't know what auto-register means and what `Class.forName("org.duckdb.DuckDBDriver");` does. I'll leave up to you if you think we should say something more here or that's enough for Java users.

docs/current/clients/java/connecting.md
```diff
@@ -0,0 +1,294 @@
...
+import java.sql.DriverManager;
+import org.duckdb.DuckDBConnection;
+
+DuckDBConnection conn = (DuckDBConnection) DriverManager.getConnection("jdbc:duckdb:");
+```
+
+Additional connections can be created using the `DriverManager`. The `DuckDBConnection#duplicate()` method opens another connection to the same database instance without re-reading the configuration:
+
+```java
+Connection conn2 = ((DuckDBConnection) conn).duplicate();
```
nantiamak: Should `conn2` be a `DuckDBConnection` instead of `Connection`?

docs/current/clients/java/connecting.md
```diff
@@ -0,0 +1,294 @@
...
+```java
+Properties connectionProperties = new Properties();
+connectionProperties.setProperty("temp_directory", "/path/to/temp/dir/");
+connectionProperties.setProperty("memory_limit", "4GB");
+Connection conn = DriverManager.getConnection("jdbc:duckdb:/tmp/my_database", connectionProperties);
+```
+
+The available options are listed on the [Configuration page]({% link docs/current/configuration/overview.md %}) and can also be queried at runtime with `duckdb_settings()`. The driver exposes the same list through the standard `Driver.getPropertyInfo()` call, which returns one `DriverPropertyInfo` per DuckDB setting alongside the driver's own options.
+
+> Note Options passed at connection time are applied as **global** settings on the database instance. A setting that only has session scope has to be applied with a `SET` statement after connecting; setting it at connection time fails with `Could not set option "⟨name⟩" as a global option`.
```
nantiamak: Should we mention somewhere which settings have session scope only?

docs/current/clients/java/querying.md
```diff
@@ -0,0 +1,98 @@
...
+redirect_from:
+- /docs/clients/java/querying
+- /docs/preview/clients/java/querying
+- /docs/stable/clients/java/querying
+title: Run Queries
+---
+
+## Overview
+
+The DuckDB JDBC driver implements the standard JDBC API, so Java applications send queries and read result sets through the familiar `Statement`, `PreparedStatement`, and `ResultSet` interfaces. This page covers sending queries and reading DuckDB's nested and composite types. For opening the `Connection` these run on, see [Define Connections]({% link docs/current/clients/java/connecting.md %}).
```
nantiamak: I think "The DuckDB JDBC driver implements the standard JDBC API" is redundant as this was already said in the Overview page.

docs/current/clients/java/querying.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Sending Queries
+
+DuckDB supports the standard JDBC methods to send queries and retrieve result sets. First a `Statement` object has to be created from the `Connection`, this object can then be used to send queries using `execute` and `executeQuery`. `execute()` is meant for queries where no results are expected like `CREATE TABLE` or `UPDATE` etc. and `executeQuery()` is meant to be used for queries that produce results (e.g., `SELECT`). Below are two examples. See also the JDBC [`Statement`](https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html) and [`ResultSet`](https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html) documentations.
```
nantiamak: nit: should we add parentheses to `execute` and `executeQuery` as we in the next sentence?

docs/current/clients/java/data_import.md
```diff
@@ -0,0 +1,128 @@
...
+    appender.append(20);
+    appender.append(-8.1);
+    appender.append("world");
+    appender.endRow();
+}
+```
+
+An appender can also be created with `createAppender(tableName)` for the default schema, or with `createAppender(catalogName, schemaName, tableName)` to target a specific catalog. Buffered rows are written to the table when the appender is closed, or earlier by calling `flush()`. Within a row, `appendNull()` writes a `NULL` and `appendDefault()` writes the column's `DEFAULT` value.
+
+Nested and collection types are appended within a row as well. Use `beginStruct()` and `endStruct()` for a `STRUCT` column, and `beginUnion(tag)` and `endUnion()` for a `UNION` column. For a `LIST` or `ARRAY` column, pass a Java array as a single value, optionally with a boolean null mask for its elements:
```
nantiamak: What is `tag` in `beginUnion(tag)`?

nantiamak (commented):
@geertjanw Thanks for all the effort you put into this PR! I still need to review `result_handling.md`, `profiling.md`, `functions.md` and `known_issues.md`. I'll get to them after lunch.
Here are some comments on the pages I already went over. 
Just a note that I didn't run the code examples.
Links so far are correct, I click all of them.

docs/current/clients/java/data_import.md
staticlibs: Hi Nantia, thanks for the review! The `tag` here is the name of the `UNION` member, for example:

```sql
CREATE TABLE tab1 (col1 UNION(u1 INTEGER, u2 VARCHAR))
```

Then the insert can be one of:

```java
.beginUnion("u1")
.append(42)
.endUnion()
```

or 

```java
.beginUnion("u2")
.append("foo")
.endUnion()
```

nantiamak: @staticlibs Thanks for clarifying! I guess it's fine not to mention what it is in this page and let the user figure it out via the API reference?

staticlibs: Yes, this is supposed to be covered by the API reference (Javadoc, that should cover our extensions to standard JDBC API), just we don't have that reference yet.

docs/current/clients/java/result_handling.md
```diff
@@ -0,0 +1,141 @@
...
+}
+```
+
+### Arrow Import
+
+The following example consumes an Arrow stream produced by the Java Arrow bindings and registers it on a DuckDB connection with `registerArrowStream()`. Once registered, the stream is addressed by the name passed to that method and can be queried like any other table.
+
+```java
+import org.apache.arrow.memory.RootAllocator;
+import org.apache.arrow.vector.ipc.ArrowReader;
```
nantiamak: I didn't run the code, but seems to me that here we need to import `ArrowStreamReader` instead of `ArrowReader`, because we're using `ArrowStreamReader` below and we don't use `ArrowReader` at all in this example.

docs/current/clients/java/overview.md
geertjanw: Good idea, added a SELECT 42 sanity-check snippet right after opening the connection, showing the expected output (42) so readers can confirm the driver is on the classpath and connecting.

docs/current/clients/java/connecting.md
geertjanw: Expanded the sentence to explain that on modern JVMs the driver registers itself automatically via Java's service-provider mechanism (with a link), and that Class.forName(...) just force-loads the driver class to trigger registration if that doesn't happen.

docs/current/clients/java/connecting.md
geertjanw: Fixed, duplicate() returns a DuckDBConnection, so conn2 is now declared as one. While checking for the same pattern I also found and fixed a broken Appender snippet in data_import.md (a missing closing brace on the outer try-with-resources block).

docs/current/clients/java/connecting.md
geertjanw: Added a pointer to the note: query duckdb_settings() and filter on scope = 'LOCAL' to find them, with search_path, schema, and enable_profiling / profiling_mode called out as common examples.

docs/current/clients/java/querying.md
geertjanw: OK, dropped it. The sentence now starts "Java applications send queries and read result sets through the familiar Statement, PreparedStatement, and ResultSet interfaces."

docs/current/clients/java/querying.md
geertjanw: Done, now is execute() and executeQuery(), consistent with the following sentence.

nantiamak (approved):
Just one more comment. Otherwise it looks good to me. Thanks again!

docs/current/clients/java/result_handling.md
geertjanw: Fixed.

PR #7172, opened 2026-08-22
Add blog post: Querying MongoDB from DuckDB with Java Table Functions

geertjanw:
Adds a deep-dive post describing an example project that implements a mongo_query() DuckDB table function in Java, reading a MongoDB collection directly into SQL. Reuses the Java thumbnail images and follows the existing post metadata conventions. Dated 2026-08-28.


_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
+---
+layout: post
+title: "Heterogeneous Joins in DuckDB with Java Table Functions"
```
staticlibs: Suggested title: "DuckDB Table Functions in Java" or "Writing DuckDB Table Functions in Java"

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+author: "Geertjan Wielenga, Alex Kasko"
+excerpt: "The DuckDB Java client can register table functions written in pure Java, exposing any Java-accessible data source as a SQL table. That turns DuckDB into a single-node query engine for heterogeneous joins across remote systems and local files, with no native extension and no export step."
+tags: ["deep dive"]
+thumb: "/images/blog/thumbs/java.svg"
+image: "/images/blog/thumbs/java.png"
+---
+
+In a large organization, data is spread across many systems: relational databases, document stores such as MongoDB, message queues, and older services reached only through a vendor SDK or a SOAP endpoint. The layer that ties these together is often a JVM-based distributed query engine, such as Trino, that can run a single query joining across all of them. The Java clients for those systems are mature and fast, with streaming and asynchronous APIs. They are also already in production and cleared by corporate security.
```
staticlibs: Lets remove MongoDB from here. Suggest to mention cloud systems (Snowflake, without mentioning it) and data lakes instead.

staticlibs: > and older services reached only through a vendor SDK or a SOAP endpoint

Suggested wording: moving this to a separate sentence like: Some of these systems can be reached only by a vendor SDK (usually provided in Java) or a SOAP endpoint [dont' need to mention that, but reader should realize that no sane code except Java will talk to these services]. Also suggest to mention custom authentication/SSO, again implying that it is difficult to talk to them without Java. All these is not true for MSFT-centric environments, but for this blog post the target is Java (application servers, Spark) environment.

staticlibs: > The Java clients for those systems are mature and fast, with streaming and asynchronous APIs

Suggest to extend this mentioning both streaming and virtual threads (for async)

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+Teams increasingly add DuckDB to these environments for fast single-node analytics. Because the surrounding infrastructure is Java, they use it through its [JDBC driver]({% link docs/current/clients/java/overview.md %}). But a query engine is only as useful as the data it can reach, and in these environments much of that data is accessible only through Java client libraries.
```
staticlibs: > they use it

Suggestion: "they use DuckDB"

staticlibs: Suggest to add another 1-2 sentence paragraph below about: now Java developers require to process some of the data in SQL in DuckDB (in a performant and low-overhead way) and some processing, that requires external Java-sourced data to be processed in Java application code on top of DuckDB SQL result set. May be more convenient (and, in many cases - performant) to process this external data inside DuckDB's SQL. Table functions are intended for that.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+DuckDB already reads almost everything. It connects to PostgreSQL, MySQL, and SQLite directly. Community extensions cover many more systems. JSON functions can query REST APIs from SQL. And the ODBC extension reaches Oracle, SQL Server, DB2, and other proprietary databases. For anything not covered, DuckDB has always supported *table functions* as a way to plug in a custom source, and those functions could be written in C++ or Rust.
```
staticlibs: > reads almost everything

Suggested order of examples:

 - parquet from data lakes
 - tightly integrated relational DBs - Postgres, MySQL, SQlite
 - proprietary DBs - Oracle, MSSQL, DB2 with ODBC
 - REST-like JSON services by quering them as remote JSON files
 - community extensions to many other DBs

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+The DuckDB Java client can now register table functions written in [pure Java]({% link docs/current/clients/java/functions.md %}#table-functions). Any data source you can reach from Java can be exposed as a SQL table function, reusing the same streaming, asynchronous client libraries your infrastructure already runs and trusts. You can then query it alongside local Parquet and CSV files, wildcard globs included, joining and filtering across all of them in a single SQL statement.
+
+Until now, adding a custom source in native code meant building and shipping a DuckDB extension: a C++ or Rust toolchain, a build system that Java developers find complex and brittle (CMake plus vcpkg), and the operational risk that a crash in native code is a segfault that takes down the whole JVM, and with it the node. A pure-Java table function is an ordinary Maven dependency that runs on the JVM alongside the rest of the application, with far less risk.
```
staticlibs: > a C++ or Rust toolchain

Lets mention only C++ here, almost all community extensions are C++

staticlibs: > a build system that Java developers find complex and brittle 

Suggested rephrase:  "with notoriously complex build systems" [without mentioning Java developers]

staticlibs: > , and with it the node

Suggest to remove.

staticlibs: Perhaps add a sentence, that any JVM-based language can be used, perhaps with Kotlin as example

_posts/2026-08-25-table-functions-in-java.md
```diff
@@ -0,0 +1,197 @@
...
+
+The DuckDB Java client can now register table functions written in [pure Java]({% link docs/current/clients/java/functions.md %}#table-functions). Any data source you can reach from Java can be exposed as a SQL table function, reusing the same streaming, asynchronous client libraries your infrastructure already runs and trusts. You can then query it alongside local Parquet and CSV files, wildcard globs included, joining and filtering across all of them in a single SQL statement.
+
+Until now, adding a custom source in native code meant building and shipping a DuckDB extension: a C++ or Rust toolchain, a build system that Java developers find complex and brittle (CMake plus vcpkg), and the operational risk that a crash in native code is a segfault that takes down the whole JVM, and with it the node. A pure-Java table function is an ordinary Maven dependency that runs on the JVM alongside the rest of the application, with far less risk.
+
+This post walks through an [example project](https://github.com/staticlibs/duckdb_mongo_example/blob/master/README.md) that adds a `mongo_query()` function to DuckDB, and uses it to explain the mechanism.
```
staticlibs, line 27: Suggested a sentence introduction of Mongo (in neutral tone), that it is a popular document storage with a robust Java API. [it also restricts JDBC and ODBC access in various way, no need to mention this to not bring negativity].

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+This post walks through an [example project](https://github.com/staticlibs/duckdb_mongo_example/blob/master/README.md) that adds a `mongo_query()` function to DuckDB, and uses it to explain the mechanism.
+
+> The `mongo_query()` function below is a deliberately minimal, illustrative example, not a production connector. There are better ways to reach MongoDB in particular, including the [MongoDB community extension]({% link community_extensions/extensions/mongo.md %}) and the [ODBC extension]({% link docs/current/core_extensions/odbc/overview.md %}). MongoDB is used here only because it is a familiar system with a good Java client. The same pattern applies to any source that has one, such as a SOAP service or a proprietary vendor SDK.
```
staticlibs: Lets swap the order of ODBC and mongo-community. ODBC is restricted/deprecated in Mongo, still in general we don't promote community extensions over in-house ones, not because they are bad, just because we don't provide paid support for them.

staticlibs: > it is a familiar system

Suggestion: popular system

staticlibs: > that has one

Lets refrase this mentioning JDBC [as one of the most common clients]

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+    'tab1',
+    '{ "col1": "foo", "col2": { "$lt": 42 } }',
+    columns = '["col1", "col2", "col3"]',
+    hostname = 'localhost',
+    port = 27017,
+    database = 'db1'
+);
+```
+
+The first argument is a MongoDB collection name. The second is a MongoDB filter document, in native MongoDB query syntax, passed through unchanged. The named `columns` parameter specifies which fields to project and the shape of the result. Run against a seeded collection, the function returns the documents that match the filter:
```
staticlibs: Suggest an explanatory note to the args: because we are using Mongo's BSON filters instead of full SQL queries, we need to pass the table name and the desired resulting column names in separate parameters.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+┌─────────┬───────┬─────────┐
+│  col1   │ col2  │  col3   │
+│ varchar │ int32 │ varchar │
+├─────────┼───────┼─────────┤
+│ foo     │    10 │ match-a │
+│ foo     │    41 │ match-b │
+└─────────┴───────┴─────────┘
+```
+
+There is no export step and no intermediate table. The rows are read from a MongoDB cursor during query execution. And because the result is an ordinary table function, it can be joined with anything else DuckDB can read. Here it joins a remote collection against a directory of local CSV files in one statement:
```
staticlibs: > no export step

"no export to Parquet step"

staticlibs: > result is an ordinary table function

"ordinary relation" may be better here.

_posts/2026-08-25-table-functions-in-java.md
```diff
@@ -0,0 +1,197 @@
...
+FROM mongo_query(
+         'orders',
+         '{ "status": "shipped" }',
+         columns = '["customer_id", "amount"]',
+         database = 'app'
+     ) AS o
+JOIN 'customers/*.csv' AS c
+    ON c.customer_id = o.customer_id
+GROUP BY c.region;
+```
```
staticlibs, line 74: No change suggested, just I think this query is really good illustration to the idea!

_posts/2026-08-25-table-functions-in-java.md
```diff
@@ -0,0 +1,197 @@
...
+        .withParameter(String.class)   // Mongo filter (JSON)
+        .withNamedParameter("columns", String.class)
+        .withNamedParameter("hostname", String.class)
+        .withNamedParameter("port", Integer.class)
+        .withNamedParameter("database", String.class)
+        .withNamedParameter("username", String.class)
+        .withNamedParameter("password", String.class)
+        .withFunction(new MongoQueryFunction())
+        .register(connection);
+```
```
staticlibs, line 95: Need to add a note box here explaining "user/password" parameters:

"In real non-toy implementation it may be desired to open a connection to a MongoDB server once and run a few queries on it, instead opening it for every query. This can be done by implementing scalar functions - `mongo_connect()` and `mong_query()` where the former will open a connection and return its handle as a DuckDB variable. This connection then may be used by `mongo_query` and can be closed by `mongo_close`. Such implementation lies outside of scope of this blog post [actually an idea for a follow up post], see a similar example in the ODBC extension".

_posts/2026-08-25-table-functions-in-java.md
```diff
@@ -0,0 +1,197 @@
...
+
+This completes registration. `mongo_query(...)` is then available as a table function on that DuckDB connection and can be used in `FROM` clauses, joins, CTEs, and subqueries like any built-in function.
+
+The implementation is a class that implements `DuckDBTableFunction`, which follows DuckDB's native table-function lifecycle with three callbacks:
+
+- **`bind`** runs at prepare time. It reads the parameters, determines the output schema, and declares each result column with `addResultColumn()`.
+- **`init`** runs once before execution and sets up shared state, such as opening a cursor or issuing the remote request.
+- **`apply`** is called repeatedly, each time filling a chunk of the output, until it returns `0`.
```
staticlibs, line 105: Need a note box mentioning `initLocal` that may be covered in future blog posts.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+- **`bind`** runs at prepare time. It reads the parameters, determines the output schema, and declares each result column with `addResultColumn()`.
+- **`init`** runs once before execution and sets up shared state, such as opening a cursor or issuing the remote request.
+- **`apply`** is called repeatedly, each time filling a chunk of the output, until it returns `0`.
+
+The example implements these in [`MongoQueryFunction.java`](https://github.com/staticlibs/duckdb_mongo_example/blob/master/src/main/java/org/duckdb/example/MongoQueryFunction.java), with named parameters defined in [`MongoQueryParameters.java`](https://github.com/staticlibs/duckdb_mongo_example/blob/master/src/main/java/org/duckdb/example/MongoQueryParameters.java).
+
+### Bind: Declare the Output Schema
+
+`bind` reads the call parameters and declares the output columns. Because MongoDB documents have no fixed schema, this example asks the caller to list the columns rather than inferring them, and declares each one to DuckDB:
```
staticlibs: > rather than inferring them

"rather than inferring them from a prepared statement"

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+        columns.add(name);
+        info.addResultColumn(name, String.class);   // declare it to DuckDB
+    }
+
+    Document query = Document.parse(queryJson);
+    return new MongoQueryBindData(client, collection, columns, query);
+}
+```
+
+Two aspects here generalize beyond MongoDB. The source's own query is parsed and passed to its driver unchanged, so DuckDB never interprets the source's query language and the full syntax stays available. The object returned from `bind`, here a [`MongoQueryBindData`](https://github.com/staticlibs/duckdb_mongo_example/blob/master/src/main/java/org/duckdb/example/MongoQueryBindData.java), then carries that state forward to the next callback.
```
staticlibs: > parsed and passed to its driver unchanged

This bit is confusing, while technically correct. Need to emphasize that it is NOT parsed (beyond JSON->BSON, which can be ignored).

staticlibs: Additionally need to mention in this section that `bind` runs on `EXPLAIN`

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Init: Open the Cursor
+
+`init` runs once and prepares the shared state that `apply` will consume. In the example it issues the query and stores the resulting cursor in a [`MongoQueryInitData`](https://github.com/staticlibs/duckdb_mongo_example/blob/master/src/main/java/org/duckdb/example/MongoQueryInitData.java). It also pins execution to a single thread with `setMaxThreads(1)`, because a single cursor is stateful and cannot be read by multiple DuckDB workers concurrently:
```
staticlibs: > because a single cursor is stateful and cannot be read by multiple DuckDB workers concurrently

I think this is the wrong reason (and likely to be technically wrong), actual reason - we are not covering multithreaded execution in this post.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+    info.setMaxThreads(1);
+    MongoQueryBindData bindData = info.getBindData();
+    FindIterable<Document> iter = bindData.collection.find(bindData.query);
+    return new MongoQueryInitData(iter.cursor());
+}
+```
+
+### Apply: Stream Rows into Vectors
+
+`apply` reflects DuckDB's vectorized execution model. Rather than returning rows, it writes them into the output data chunk column by column, up to the chunk's capacity, and returns the number of rows produced:
```
staticlibs: > up to the chunk's capacity

"up to the chunk's capacity of 2048 rows"

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+        Document doc = cursor.next();
+        for (long col = 0; col < output.columnCount(); col++) {
+            copyValueFromResultSetToVector(doc, output.vector(col), row, columns.get((int) col));
+        }
+    }
+    return row;   // 0 signals "no more data"
+}
+```
+
+Values are written into typed vectors through the vector API, using `setString`, `setInt`, `setDouble`, `setNull`, and related methods. DuckDB calls `apply` repeatedly, draining the source one vector-sized batch at a time, so the full result is never materialized in memory on either side. In the example, the BSON-to-vector conversion lives in [`MongoTypes.java`](https://github.com/staticlibs/duckdb_mongo_example/blob/master/src/main/java/org/duckdb/example/MongoTypes.java).
```
staticlibs: > never materialized in memory

A bit too strong, it may may not be materialized by SQL operators called on top of the `mongo_query`. I think correct formulation will be "doesn't require to be materialized".

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Implications
+
+This example is a MongoDB connector, but the broader point is that DuckDB can now be extended in pure Java, putting the JVM ecosystem within reach as a data source:
```
staticlibs: > the JVM ecosystem

Suggestion: "rich ecosystem of Java libraries".

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+- **No native build.** The project contains no C++ and no platform-specific binary. It is a Maven project with two dependencies, the DuckDB JDBC driver and the MongoDB driver, built and run with `mvn test`. Any Java developer can add a new source.
```
staticlibs: Lets remove "and no platform-specific binary", it sounds ambiguous.

staticlibs: > Any Java developer can add a new source.

Suggest to extend and reformulate keeping the meaning.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+- **In-process access.** The data is not exported and reloaded. It is read from a cursor during query execution. Because the result is an ordinary table function, it can be joined with Parquet files, CSV globs, or attached databases in a single SQL statement, so DuckDB acts as the query layer over the source rather than a copy of it.
+- **Predicates run at the source.** The filter is written in the source's own query language and handed to its driver unchanged, so it executes at the source, against its indexes, and only matching rows cross the wire. This is not DuckDB filter pushdown: the C API that Java table functions build on supports projection pushdown but not filter pushdown. It does not need to be, because the caller expresses the filter directly in the function call, which in practice is often remote SQL. The selective work happens remotely, and DuckDB streams back only what it needs.
```
staticlibs: > This is not DuckDB filter pushdown: the C API that Java table functions build on supports projection pushdown but not filter pushdown. It does not need to be, because the caller expresses the filter directly in the function call, which in practice is often remote SQL.

This is confusing and required a few paragraphs to explain, lets remove this completely.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+- **A Java table function cannot yet be packaged as a DuckDB extension.** A DuckDB extension is a native shared library loaded into the DuckDB process, whereas a Java table function lives in the JVM, on the other side of the JDBC boundary, and is registered on a live connection through the driver. There is no way to package JVM code as a loadable native extension, so the function is available only to the JVM process that registered it, not to other clients or the CLI, and not to a database opened without the Java driver. <!-- TODO(Alex): confirm/expand the technical detail here. --> This may or may not change in a future release.
+- **Bind and init objects are not managed automatically.** In the current release, the caller is responsible for the lifecycle of the objects returned from `bind` and `init`, for example closing the MongoDB client and cursor after the query completes. An upcoming release manages that lifecycle for you: the returned objects implement `AutoCloseable`, and DuckDB calls `close()` when the function is done. <!-- TODO(Alex): replace with the real example from the upcoming release. --> The init object then becomes:
```
staticlibs: This is about [DuckDBTableFunctionState](https://duckdb.org/docs/current/clients/java/functions#cleaning-up-resources), lets link to https://github.com/duckdb/duckdb-java/pull/803 here to add attribution to a third-party contributor.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+          cursor.close();
+      }
+  }
+  ```
+
+- **Composite DuckDB types are not yet supported.** The vector API currently covers scalar types, but `STRUCT`, `LIST`, and other nested types are planned for the future. Until then, nested source data has to be flattened or serialized to a scalar column.
+
+## Conclusion
+
+For many analytical workloads, the data does not have to move to a cluster, and you do not need a distributed query engine to join across systems. With pure-Java table functions, the client libraries already running in your infrastructure become SQL tables, and DuckDB joins them against each other and against local files in a single query on one node, using the JDBC driver you already have.
```
staticlibs: > JDBC driver

Suggest to change to "Java client libraries"

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Current Limitations
+
+These are limitations of the current Java table-function API, not of the toy example:
+
+- **A Java table function cannot yet be packaged as a DuckDB extension.** A DuckDB extension is a native shared library loaded into the DuckDB process, whereas a Java table function lives in the JVM, on the other side of the JDBC boundary, and is registered on a live connection through the driver. There is no way to package JVM code as a loadable native extension, so the function is available only to the JVM process that registered it, not to other clients or the CLI, and not to a database opened without the Java driver. <!-- TODO(Alex): confirm/expand the technical detail here. --> This may or may not change in a future release.
```
staticlibs: There are two ways to package Java UDFs as extensions, so they can be loaded from any DuckDB client:

 - have a JNI shim (or use FFM without shim https://openjdk.org/jeps/454 ) that will load the JVM into the caller process, point to to the JAR file that implements the UDF and do the data marshaling
 - compile the above into a native image with Graal - basically doing the same without requiring JVM at runtime

Both are possible, neither is high priority. Not sure we need these details in this section, maybe shorter reasoning for the limitation will be better with the resume that limitation may or may not be lifter in future.

staticlibs (approved):
The content looks great! Added a few comments.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: - Done. Changed to just "a C++ toolchain" and also changed the earlier "written in C++ or Rust" on the paragraph above to "written in C++" for consistency.
-  Done: "…with notoriously complex build systems (CMake plus vcpkg)…"
-  Removed.
- Added: "And because it is just JVM code, any JVM language works: the same function could be written in Kotlin, Scala, or Clojure."

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: Reordered to: Parquet from data lakes → tightly integrated relational DBs (Postgres, MySQL, SQLite) → proprietary DBs via ODBC (Oracle, SQL Server, DB2) → REST-like JSON services as remote JSON files → community extensions.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: - "they use it" → "they use DuckDB"
- Added: "In these environments, some data is best processed in SQL, which DuckDB handles in a performant, low-overhead way, while other data arrives from external Java-sourced systems and is often processed in application code on top of a DuckDB result set. It is frequently more convenient, and in many cases more performant, to bring that external data into DuckDB's SQL directly. Table functions are intended for exactly that."

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: -  The list is now "relational databases, document stores, message queues, data lakes, and cloud data warehouses", no MongoDB, and cloud warehouses/data lakes are in.
- Done. Split into a separate sentence: "Some of these systems can be reached only through a vendor SDK, usually provided in Java, or a SOAP endpoint, and many sit behind custom authentication or single sign-on that is awkward to satisfy from anything but a JVM client."
- Done: "…with streaming APIs and, increasingly, virtual threads for asynchronous work."

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: Done. Retitled to "DuckDB Table Functions in Java" and renamed the file to match that title.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: Softened to "so the full result does not have to be materialized in memory on either side."

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: "up to the chunk's capacity" → "…of 2048 rows"

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: Fixed to "because this post does not cover multithreaded execution" instead of the cursor-concurrency explanation.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: - Reworded to emphasize DuckDB never parses/interprets it: "DuckDB never parses or interprets the source's query language, the filter is handed to the source's own driver essentially unchanged, apart from a JSON-to-BSON conversion that can be ignored…"
- Added: "Note that bind also runs when a query is EXPLAINed, before any rows are read, since DuckDB needs the output schema to plan the query."

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: "rather than inferring them" → "…from a prepared statement"

_posts/2026-08-25-table-functions-in-java.md
geertjanw: Added: "The API also exposes an initLocal callback for per-thread local state used in multithreaded execution. This post keeps to single-threaded execution and does not use it, it may be covered in a future post."

_posts/2026-08-25-table-functions-in-java.md
geertjanw: Added a note box that explains the username/password params open a fresh connection per call, and that a real implementation would keep a connection open via mongo_connect() / mongo_query() / mongo_close() scalar functions, which is out of scope here, possible follow-up, cf. the ODBC extension.

_posts/2026-08-25-table-functions-in-java.md
geertjanw: Thanks. :-)

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: - "no export step" → "no export to Parquet step"
- "ordinary table function" → "ordinary relation"

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: Added: "Because the filter is a MongoDB BSON document rather than a full SQL query, the collection name and the desired result columns cannot be read out of the query itself and are passed as separate parameters instead."

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: -  Done. ODBC extension is now listed first, community extension second.
- "familiar system" → "popular system"
- Rephrased: "…any source reachable from the JVM, whether through a JDBC driver, one of the most common Java clients, a SOAP service, or a proprietary vendor SDK."

_posts/2026-08-25-table-functions-in-java.md
geertjanw: Added at the start of the note box: "MongoDB is a popular document store with a robust Java API."

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: Done.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: Kept it short as suggested, the limitation now notes it's technically possible via a JNI or FFM shim that embeds the JVM, or a GraalVM native image, but that neither is supported today.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: Done, the lifecycle bullet now links to the DuckDBTableFunctionState docs section and credits the contributor via duckdb-java#803.

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: Removed the "This is not DuckDB filter pushdown…" sentences entirely. The bullet now stops at "…only matching rows cross the wire. The selective work happens remotely, and DuckDB streams back only what it needs."

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: -  Removed, the bullet now just reads "The project contains no C++."
-  Reworded to: "Adding a new source is ordinary Java work, done with the tools a team already uses and without touching a native toolchain."

_posts/2026-08-28-querying-mongodb-with-java-table-functions.md
geertjanw: "the JVM ecosystem" → "rich ecosystem of Java libraries"

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
+---
+layout: post
+title: "DuckDB Table Functions in Java"
+author: "Geertjan Wielenga, Alex Kasko"
+excerpt: "The DuckDB Java client can register table functions written in pure Java, exposing any Java-accessible data source as a SQL table. That turns DuckDB into a single-node query engine for heterogeneous joins across remote systems and local files, with no native extension and no export step."
```
staticlibs: > with no native extension

Lets remove this, it is not immediately clear of what native extentions is this about

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+tags: ["deep dive"]
+thumb: "/images/blog/thumbs/java.svg"
+image: "/images/blog/thumbs/java.png"
+---
+
+In a large organization, data is spread across many systems, including relational databases, document stores, message queues, data lakes, and Cloud data warehouses. Some of these systems can be reached only through a vendor SDK, usually provided in Java, or a SOAP endpoint, and many sit behind custom authentication or single sign-on that is awkward to satisfy from anything but a JVM client. The layer that ties these together is often a JVM-based distributed query engine, such as Trino, that can run a single query joining across all of them. The Java clients for those systems are mature and fast, with streaming APIs and, increasingly, virtual threads for asynchronous work. They are also already in production and cleared by corporate security.
```
staticlibs: > across all of them

Suggest to change to "across all data source", or "across multiple data sources"

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Extending DuckDB in Pure Java
+
+The DuckDB Java client can now register table functions written in [pure Java]({% link docs/current/clients/java/functions.md %}#table-functions). Any data source you can reach from Java can be exposed as a SQL table function, reusing the same streaming, asynchronous client libraries your infrastructure already runs and trusts. You can then query it alongside local Parquet and CSV files, wildcard globs included, joining and filtering across all of them in a single SQL statement.
+
+Until now, adding a custom source in native code meant building and shipping a DuckDB extension. That involved a C++ toolchain, notoriously complex build systems (CMake plus vcpkg), and the operational risk that a crash in native code is a segfault that takes down the whole JVM. A pure-Java table function is an ordinary Maven dependency that runs on the JVM alongside the rest of the application, with far less risk. And because it is just JVM code, any JVM language works, so the same function could be written in Kotlin, Scala, or Clojure.
```
staticlibs: > (CMake plus vcpkg)

Lets remove this, in theory one can build C++ extension without either of them.

staticlibs: > And because it is just JVM code

"JVM code" is a bit confusing, perhaps - "just a Java API"

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+        .withNamedParameter("database", String.class)
+        .withNamedParameter("username", String.class)
+        .withNamedParameter("password", String.class)
+        .withFunction(new MongoQueryFunction())
+        .register(connection);
+```
+
+This completes registration. `mongo_query(...)` is then available as a table function on that DuckDB connection and can be used in `FROM` clauses, joins, CTEs, and subqueries like any built-in function.
+
+> The `username` and `password` parameters open a fresh connection on every call. In a real, non-toy implementation it is often preferable to open a connection to the MongoDB server once and run several queries on it. That can be done with a set of scalar functions. A `mongo_connect()` opens a connection and returns its handle as a DuckDB value, a `mongo_query()` reads through that handle, and a `mongo_close()` closes it. That approach is outside the scope of this post and could be the subject of a follow-up. For a similar example, see the ODBC extension.
```
staticlibs: > The `username` and `password` parameters open a fresh connection on every call.

Confusing a bit, perhaps something like: "this example takes `username` and `password` parameters an opens a fresh connection on every call".

staticlibs: > to open a connection to the MongoDB server

Suggest to replace with: "to open a connection to the remote server"

staticlibs: > see the ODBC extension

suggest to add a link to `odbc_connect` - https://duckdb.org/docs/current/core_extensions/odbc/functions#odbc_connect

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+        columns.add(name);
+        info.addResultColumn(name, String.class);   // declare it to DuckDB
+    }
+
+    Document query = Document.parse(queryJson);
+    return new MongoQueryBindData(client, collection, columns, query);
+}
+```
+
+Two aspects here generalize beyond MongoDB. DuckDB never parses or interprets the source's query language. The filter is handed to the source's own driver essentially unchanged, apart from a JSON-to-BSON conversion that can be ignored, so the full source syntax stays available. The object returned from `bind`, here a [`MongoQueryBindData`](https://github.com/staticlibs/duckdb_mongo_example/blob/master/src/main/java/org/duckdb/example/MongoQueryBindData.java), then carries that state forward to the next callback. Note that `bind` also runs when a query is `EXPLAIN`ed, before any rows are read, since DuckDB needs the output schema to plan the query.
```
staticlibs: >  that can be ignored

Lets remove this bit

staticlibs: > when a query is `EXPLAIN`ed, before any rows are read

This is confusing a bit, no data is read on `EXPLAIN`, lets just remove `before any rows are read`

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Init: Open the Cursor
+
+`init` runs once and prepares the shared state that `apply` will consume. In the example it issues the query and stores the resulting cursor in a [`MongoQueryInitData`](https://github.com/staticlibs/duckdb_mongo_example/blob/master/src/main/java/org/duckdb/example/MongoQueryInitData.java). It also pins execution to a single thread with `setMaxThreads(1)`, because this post does not cover multithreaded execution:
```
staticlibs: > prepares the shared state

"prepares the execution state", or alternatively need to expand to explain that this state will be "shared" when the execution is multithreaded.

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+        Document doc = cursor.next();
+        for (long col = 0; col < output.columnCount(); col++) {
+            copyValueFromResultSetToVector(doc, output.vector(col), row, columns.get((int) col));
+        }
+    }
+    return row;   // 0 signals "no more data"
+}
+```
+
+Values are written into typed vectors through the vector API, using `setString`, `setInt`, `setDouble`, `setNull`, and related methods. DuckDB calls `apply` repeatedly, draining the source one vector-sized batch at a time, so the full result does not have to be materialized in memory on either side. In the example, the BSON-to-vector conversion lives in [`MongoTypes.java`](https://github.com/staticlibs/duckdb_mongo_example/blob/master/src/main/java/org/duckdb/example/MongoTypes.java).
```
staticlibs: > one vector-sized batch at a time

"one vector-sized batch of 2048 rows at a time"

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Implications
+
+This example is a MongoDB connector, but the broader point is that DuckDB can now be extended in pure Java, putting the rich ecosystem of Java libraries within reach as a data source:
+
+- **No native build.** The project contains no C++. It is a Maven project with two dependencies, the DuckDB JDBC driver and the MongoDB driver, built and run with `mvn test`. Adding a new source is ordinary Java work, done with the tools a team already uses and without touching a native toolchain.
```
staticlibs: > , built and run with `mvn test`

Lets remove this bit.

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+- **Predicates run at the source.** The filter is written in the source's own query language and handed to its driver unchanged, so it executes at the source, against its indexes, and only matching rows cross the wire. The selective work happens remotely, and DuckDB streams back only what it needs.
+
+## Current Limitations
+
+These are limitations of the current Java table-function API:
+
+- **A Java table function cannot yet be packaged as a DuckDB extension.** A DuckDB extension is a native shared library loaded into the DuckDB process, whereas a Java table function lives in the JVM and is registered on a live connection through the driver. As a result, the function is available only to the JVM process that registered it, not to other clients or the CLI, and not to a database opened without the Java driver. Packaging JVM code as a loadable extension is technically possible, for example through a JNI or [FFM](https://openjdk.org/jeps/454) shim that embeds the JVM, or by compiling to a GraalVM native image, but neither is supported today.
```
staticlibs: > lives in the JVM and is registered on a live connection through the driver

Lets replace with "requires a JVM to [be?] run"

staticlibs: > As a result, the function is available only to the JVM process that registered it, not to other clients or the CLI, and not to a database opened without the Java driver.

"As a result, the function can only be used from DuckDB Java client".

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+- **Bind and init objects are not managed automatically.** In the current release, the caller is responsible for the lifecycle of the objects returned from `bind` and `init`, for example closing the MongoDB client and cursor after the query completes. A [`DuckDBTableFunctionState`]({% link docs/current/clients/java/functions.md %}#cleaning-up-resources) mechanism ([contributed](https://github.com/duckdb/duckdb-java/pull/803) by a community member) manages that lifecycle for you. The returned objects implement `AutoCloseable`, and DuckDB calls `close()` when the function is done. The init object then becomes:
```
staticlibs: > The returned objects implement `AutoCloseable`, and DuckDB calls `close()` when the function is done.

Lets remove this bit - technically not true

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+
+  ```java
+  public class MongoQueryInitData implements AutoCloseable {
```
staticlibs: Lets write as `implements AutoCloseable /* DuckDBTableFunctionState */`

_posts/2026-08-24-table-functions-in-java.md
```diff
@@ -0,0 +1,207 @@
...
+          cursor.close();
+      }
+  }
+  ```
+
+- **Composite DuckDB types are not yet supported.** The vector API currently covers scalar types, but `STRUCT`, `LIST`, and other nested types are planned for the future. Until then, nested source data has to be flattened or serialized to a scalar column.
+
+## Conclusion
+
+For many analytical workloads, the data does not have to move to a cluster, and you do not need a distributed query engine to join across systems. With pure-Java table functions, the client libraries already running in your infrastructure become SQL tables, and DuckDB joins them against each other and against local files in a single query on one node, using the Java client libraries you already have.
```
staticlibs: > client libraries [...] become SQL tables

Perhaps rephrase a bit, "clients provide access to remove data sources as SQL tables". Or "expose remote data sources as SQL tables". Or something more streamlined, just libraries themselves are not tables.

geertjanw: Done, reworded to "the client libraries already running in your infrastructure expose remote data sources as SQL tables".

_posts/2026-08-24-table-functions-in-java.md
geertjanw: Done, now implements AutoCloseable /* DuckDBTableFunctionState */.

_posts/2026-08-24-table-functions-in-java.md
geertjanw: Done, removed.

_posts/2026-08-24-table-functions-in-java.md
geertjanw: - Done, changed to "requires a JVM to run".
- Done, changed to "As a result, the function can only be used from the DuckDB Java client."

_posts/2026-08-24-table-functions-in-java.md
geertjanw: Done, removed.

_posts/2026-08-24-table-functions-in-java.md
geertjanw: Done, changed to "one vector-sized batch of 2048 rows at a time".

_posts/2026-08-24-table-functions-in-java.md
geertjanw: Done, changed to "prepares the execution state".

_posts/2026-08-24-table-functions-in-java.md
geertjanw: - Done, removed.
- Done, removed. Now is "also runs when a query is EXPLAINed, since DuckDB needs the output schema to plan the query."

_posts/2026-08-24-table-functions-in-java.md
geertjanw: - Done, reworded to "This example takes username and password parameters and opens a fresh connection on every call."
- Done, changed to "the remote server".
-  Done, now links to odbc_connect.

_posts/2026-08-24-table-functions-in-java.md
geertjanw: - Done, removed. Left it as "notoriously complex build systems".
 - Done, changed to "just a Java API".

_posts/2026-08-24-table-functions-in-java.md
geertjanw: Done, changed to "joining across multiple data sources".

_posts/2026-08-24-table-functions-in-java.md
geertjanw: Done, removed "with no native extension".

_posts/2026-08-25-table-functions-in-java.md
```diff
@@ -0,0 +1,220 @@
...
+
+Teams increasingly add DuckDB to these environments for fast single-node analytics. Because the surrounding infrastructure is Java, they use the [DuckDB Java client]({% link docs/current/clients/java/overview.md %}). But a query engine is only as useful as the data it can reach, and in these environments much of that data is accessible only through Java client libraries.
```
staticlibs: > DuckDB Java client

I suggest to make another pass over the terminology used for "JVM client/Java driver to remote DB" and "DuckDB Java client/JDBC driver" throughout the whole article. Perhaps  "DuckDB Java client" is indeed better term stylistically, though readers (Java developers) maybe more familiar with "JDBC driver" term (as "client" meaning is a bit wider, and we are referring). The meaning is close enough for these to be fully interchangeable, so this comment is only about the consistent usage throughout the whole article.

_posts/2026-08-25-table-functions-in-java.md
```diff
@@ -0,0 +1,220 @@
...
+
+- **A Java table function cannot yet be packaged as a DuckDB extension.** A DuckDB extension is a native shared library loaded into the DuckDB process, whereas a Java table function requires a JVM to run. As a result, the function can only be used from the DuckDB Java client. Packaging JVM code as a loadable extension is technically possible, for example through a JNI or [FFM](https://openjdk.org/jeps/454) shim that embeds the JVM, or by compiling to a GraalVM native image, but neither is supported today.
+- **Bind and init objects are not managed automatically.** In the current release, the caller is responsible for the lifecycle of the objects returned from `bind` and `init`, for example closing the MongoDB client and cursor after the query completes. A [`DuckDBTableFunctionState`]({% link docs/current/clients/java/functions.md %}#cleaning-up-resources) mechanism ([contributed](https://github.com/duckdb/duckdb-java/pull/803) by a community member) manages that lifecycle for you. The init object then becomes:
+
+  ```java
+  public class MongoQueryInitData implements AutoCloseable
+      /* DuckDBTableFunctionState */ {
```
staticlibs: Lets move this comment back one line higher to make it:

> implements AutoCloseable /* DuckDBTableFunctionState */

The idea behind this is that when 

> implements DuckDBTableFunctionState

is used - only then it will be called from the native part. But we cannot use it in this form, because it is not included in the latest release (only coming in 1.5.6), so it makes the code not-compilable. Thus using `AutoCloseable` to keep it compilable and `DuckDBTableFunctionState` to be uncommented by users on newer version (and to highlight the lifecycle support changes from PR 803)

_posts/2026-08-25-table-functions-in-java.md
geertjanw: Maybe the first instance "DuckDB Java Client (the JDBC driver)" and all other instances only "DuckDB Java Client"?

_posts/2026-08-25-table-functions-in-java.md
```diff
@@ -0,0 +1,220 @@
...
+
+Teams increasingly add DuckDB to these environments for fast single-node analytics. Because the surrounding infrastructure is Java, they use the [DuckDB Java client]({% link docs/current/clients/java/overview.md %}). But a query engine is only as useful as the data it can reach, and in these environments much of that data is accessible only through Java client libraries.
+
+## Bringing External Data into DuckDB
+
+Before reaching for the DuckDB Java (JDBC) client, it is worth asking what DuckDB can already read on its own, and the answer is almost everything. It reads Parquet and other files directly from data lakes. It connects to relational databases such as PostgreSQL, MySQL, and SQLite directly. The ODBC extension reaches proprietary databases such as Oracle, SQL Server, and DB2. JSON functions can query REST-like services by treating them as remote JSON files, and community extensions cover many more systems.
```
staticlibs: Sorry, this paragraph needs to be reformulated another time, otherwise the logic is skewed:

> Before reaching for the DuckDB Java (JDBC) client, it is worth asking what DuckDB can already read on its own

DuckDB Java client is used to access DuckDB itself, so not directly comparable/related to other sources that DuckDB can read. This parts of the sentence should highlight only the UDF functionality of the driver (in conjunction with the JVM-based client to a remote data source) when comparing it with existing DuckDB capabilities.

_posts/2026-08-25-table-functions-in-java.md
```diff
@@ -0,0 +1,220 @@
...
+
+For a source that none of those cover, the data typically ends up in application code, processed by hand on top of a DuckDB result set. It is often more convenient, and frequently more performant, to bring that data into DuckDB's SQL directly, where it can be filtered and joined like any other table. [*Table functions*]({% link docs/current/clients/c/table_functions.md %}) are meant for exactly that. DuckDB has always let you plug in a custom source this way, but until now such a function had to be written in C++.
```
staticlibs: > but until now such a function had to be written in C++

Not sure if it is worth mentioning here "and some other clients". In theory can be also done in Rust and .NET. Though all (or almost all) community extensions are C++, so it may be not worthy mentioning others, perhaps to change the formulation somehow to highlight, that while C++ is not mandatory, it is was the only practical/popular choice until now to add a new UDF to be usable from DuckDB java client.

_posts/2026-08-25-table-functions-in-java.md
staticlibs: Absolutely, this was only about the consistency in naming DuckDB java client (which is JDBC + extension) vs Java clients to other systems (which are likely JDBC, but can well be non-JDBC as well)

_posts/2026-08-25-table-functions-in-java.md
geertjanw: Reworded to avoid implying C++ was mandatory: "DuckDB has always let you plug in a custom source this way, but until now that meant writing the function in C++, which in practice was the only realistic choice." Dropped any mention of other languages.

_posts/2026-08-25-table-functions-in-java.md
geertjanw: Reworked so the comparison is about the new capability rather than the client itself. It now reads: "Before writing a table function to reach a new source, it is worth asking what DuckDB can already read on its own, and the answer is almost everything." That sets the section against DuckDB's existing read capabilities without implying the DuckDB Java client is comparable to the remote sources.

_posts/2026-08-25-table-functions-in-java.md
geertjanw: Done a pass over the whole article. "DuckDB Java client" is used consistently for the DuckDB side (with "(the JDBC driver)" spelled out on first use), and "Java client / Java driver / JVM client" is reserved for the clients to remote systems, so the two are no longer conflated.

_posts/2026-08-25-table-functions-in-java.md
geertjanw: It now reads implements AutoCloseable /* DuckDBTableFunctionState */ on one line, so users on 1.5.6+ can swap in DuckDBTableFunctionState directly while the code stays compilable on the current release.

PR #7178, opened 2026-08-24
php_pdo: added links for integration into Laravel and Symfony

thomas-0816:
I've implemented integrations for PHP Laravel and Symfony frameworks:

https://github.com/thomas-0816/laravel-duckdb

https://github.com/thomas-0816/doctrine-dbal-duckdb

Also added support for https://github.com/swoole/swoole-src and https://github.com/true-async


szarnyasg:
Thanks!

PR #7197, opened 2026-08-28
Update quick install script from sh to bash

asarama:
Just a copy replace job...

The installer script is explicitly bash (#!/bin/bash -e on line 1), so piping to sh breaks on dash-based systems like Ubuntu. macOS is fine (sh → bash).

The script uses [[ at install.duckdb.org line 128, a bash-only construct.


szarnyasg:
@asarama, thanks! This is truly in the "how did this ever work" territory but I guess `sh` is lenient on most systems...

image

PR #7201, opened 2026-08-28
docs: new & updated DuckDB Go content

geertjanw:
This pull request restructures the Go and Rust client documentation to introduce subpages for each major topic, and updates the navigation and internal links to point to these new subpages. The main client pages for Go and Rust have been split into multiple topic-focused files (such as Overview, Connect, Import Data, etc.), improving discoverability and maintainability. Navigation menus and ecosystem diagrams now link directly to the new overview pages.

**Documentation Restructuring and Navigation Updates:**

* The Go and Rust client sections in `_data/menu_docs_current.json` now use a `slug` and include a `subsubfolderitems` array, listing new subpages for topics like Overview, Connect, Import Data, Run Queries, Handle Results, User Defined Functions, Profiling, and Troubleshooting. [[1]](diffhunk://#diff-f61b38e2f9a878690155277cc3e45bd42f789a3d76b0a1719fbaa9d86a397120L296-R330) [[2]](diffhunk://#diff-f61b38e2f9a878690155277cc3e45bd42f789a3d76b0a1719fbaa9d86a397120L428-R496)
* The Go and Rust client links in `_includes/ecosystem_diagram.html` and `_includes/installation.html` now point to their new `overview.md` pages instead of the old monolithic `go.md` and `rust.md` files. [[1]](diffhunk://#diff-bde39b8859d2232195c19051ff19cfa71b5794d13111b8852e8fffe79527c882L123-R123) [[2]](diffhunk://#diff-bde39b8859d2232195c19051ff19cfa71b5794d13111b8852e8fffe79527c882L169-R169) [[3]](diffhunk://#diff-dbe7ce2d3074282bf3d2530b8ae2efb149413106e83d09b38c238811c8189da2L447-R447) [[4]](diffhunk://#diff-dbe7ce2d3074282bf3d2530b8ae2efb149413106e83d09b38c238811c8189da2L469-R469)

**Go Client Documentation Refactoring:**

* The original content of `docs/current/clients/go.md` has been removed, and its topics have been split into new, focused subpages (e.g., `overview.md`, `connecting.md`, etc.).
* A new `docs/current/clients/go/connecting.md` page has been added, providing detailed instructions on how to connect to DuckDB from Go, including DSN options, connectors, thread safety, connection lifetime, and cleanup.


docs/current/clients/go/overview.md
```diff
@@ -0,0 +1,179 @@
...
+
+The import registers a driver named `duckdb` with `database/sql`. Code that calls the client's own types and functions directly, such as the Appender or user-defined functions, imports the package under its `duckdb` name instead of with the blank identifier:
+
+```go
+import "github.com/duckdb/duckdb-go/v2"
+```
+
+## Versioning
+
+Starting with DuckDB v1.5.0, the `duckdb-go` version encodes the DuckDB version it bundles in its second semantic-versioning component. The format is `v2.⟨major_minor_patch⟩.x`: for example, DuckDB v1.5.0 maps to `duckdb-go` v2.10500.x, and DuckDB {{ site.current_duckdb_go_version }} maps to v2.10505.0. The [README](https://github.com/duckdb/duckdb-go#readme) has the full mapping table for earlier releases.
```
taniabogatsch, line 54: > The format is `v2.⟨major_minor_patch⟩.x`:

not sure here - isn't the format `v2.⟨majorminorpatch⟩.x`? I guess the underscores are for readability.

docs/current/clients/go/overview.md
```diff
@@ -0,0 +1,179 @@
...
+
+For the LTS release line that stays on DuckDB 1.4 Andium, use the `v1.4-andium` branch.
```
taniabogatsch, line 56: > For the LTS release line that stays on DuckDB 1.4 Andium, use the v1.4-andium branch.

I don't think people need to use that branch except if they want to push a PR? The `v1.4` line should have its own tags, which people can just pull in their `go.mod` file by specifying the version. See for example here: https://github.com/duckdb/duckdb-go/releases/tag/v2.5.6. Note that the `v1.4` line is still on the old semantic versioning scheme.

docs/current/clients/go/overview.md
```diff
@@ -0,0 +1,179 @@
...
+
+This example follows the client's [`simple` example](https://github.com/duckdb/duckdb-go/tree/main/examples/simple). [Run Queries]({% link docs/current/clients/go/querying.md %}) covers sending queries, binding parameters, and reading results in full.
+
+### Data Source Names
+
+The DSN passed to `sql.Open()` is the database path followed by optional [DuckDB configuration options]({% link docs/current/configuration/overview.md %}) as URL-style query parameters:
+
+```go
+// In-memory database.
+db, err := sql.Open("duckdb", "")
```
taniabogatsch, line 116: I think (IIRC) that you can also pass `:memory:` here for that?

docs/current/clients/go/overview.md
```diff
@@ -0,0 +1,179 @@
...
+## Build Tags
+
+Some client features are gated behind Go [build tags](https://pkg.go.dev/go/build#hdr-Build_Constraints), passed to `go build` with the `-tags` flag, for example `go build -tags=duckdb_arrow`. The available tags:
+
+<div class="monospace_table"></div>
+
+| Build tag | Enables |
+|--|--|
+| `duckdb_arrow` | The [Apache Arrow interface]({% link docs/current/clients/go/result_handling.md %}). It is a heavy dependency, so it is opt-in. |
+| `duckdb_use_lib` | Dynamically link against a `libduckdb` library on the system instead of statically linking the bundled one. See [Troubleshoot]({% link docs/current/clients/go/troubleshoot.md %}#linking-a-dynamic-library). |
```
taniabogatsch, line 136: Doesn't have to be called `libduckdb`, its whatever people want to call it, `libmycoolduckdynamiclibrary` also works ;) So maybe we drop the `libduckdb` here and just say library?

docs/current/clients/go/overview.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Further Reading
+
+* [Connect]({% link docs/current/clients/go/connecting.md %}) — opening in-memory and file-backed databases, DSN configuration, connectors, and the connection lifetime.
+* [Run Queries]({% link docs/current/clients/go/querying.md %}) — `Exec`, `Query`, prepared statements, parameter binding, transactions, and scanning results into Go values.
+* [Import Data]({% link docs/current/clients/go/data_import.md %}) — bulk loading with the Appender and reading directly from Parquet, CSV, and JSON files.
+* [Handle Results]({% link docs/current/clients/go/result_handling.md %}) — the Apache Arrow interface for columnar result exchange.
+* [Write User Defined Functions]({% link docs/current/clients/go/functions.md %}) — scalar and table user-defined functions, and replacement scans.
+* [Profile and Monitor]({% link docs/current/clients/go/profiling.md %}) — query profiling and log storage.
```
taniabogatsch, line 171: nit: query profiling and logging

docs/current/clients/go/overview.md
```diff
@@ -0,0 +1,179 @@
...
+* [Troubleshoot]({% link docs/current/clients/go/troubleshoot.md %}) — linking, cgo, Windows setup, and other build and runtime issues.
+* [Clients Overview]({% link docs/current/clients/overview.md %}) — the other client APIs DuckDB provides alongside Go.
+
+## Acknowledgements
+
+We would like to thank [Marc Boeker](https://github.com/marcboeker) for the initial implementation of the DuckDB Go client and for his continued work on it as part of this joint effort with the DuckDB team.
+</content>
+</invoke>
```
taniabogatsch, line 179: Not sure if these are intended to be visible on the website?

taniabogatsch (commented):
Left some comments on the overview, will try to have a look at the rest next week.

mlafeldt:
@geertjanw Thanks. The structure looks correct, and the Go snippets compile and run. Three things before I do a full pass:

  1. Every page ends with a stray `</content>` - Tania caught it on the overview, but it's on all eight and renders as visible text in the preview. Looks like leftover AI markup that should be deleted.
  2. Could you rebase onto `main`? Removing all the Rust changes would make the review easier.
  3. The code holds up, but the prose doesn't always match the source. For example,`functions.md:12` says replacement scans register on a `db.Conn()` connection, but `RegisterReplacementScan` takes a `*Connector`. There are more like it. Since you're already running Claude here, a pass over the prose with the `duckdb-go` source in context should catch them cheaply.

Happy to review properly once those land.

docs/current/clients/go/overview.md
```diff
@@ -0,0 +1,177 @@
...
+- /docs/clients/go/overview
+- /docs/current/clients/go
+- /docs/preview/clients/go
+- /docs/preview/clients/go/overview
+- /docs/stable/clients/go
+- /docs/stable/clients/go/overview
+title: Go Client
+---
+
+> Installation To use the DuckDB Go client, visit the [Go installation page]({% link install/index.html %}?environment=go).
```
mlafeldt, line 16: Why should someone visit that site when the same installation method is documented in a better way below?

docs/current/clients/go/overview.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Installation
+
+The client is a Go module. Add it to a project with `go get`, using the `/v2` major-version suffix:
+
+```batch
+go get github.com/duckdb/duckdb-go/v2
+```
+
+DuckDB is written in C++, so the client uses [cgo](https://pkg.go.dev/cmd/cgo) and requires a C compiler to build. By default it statically links a pre-built DuckDB library into the binary, so no separate DuckDB installation is needed. Pre-built libraries ship for macOS (amd64, arm64), Linux (amd64, arm64), and Windows (amd64). Other platforms, custom builds, and dynamic linking are covered in [Troubleshoot]({% link docs/current/clients/go/troubleshoot.md %}#linking-duckdb).
```
mlafeldt, line 32: The README also documents `go mod vendor` for vendoring the pre-built DuckDB libraries along with the rest of the dependencies. Probably worth a sentence here.

docs/current/clients/go/overview.md
```diff
@@ -0,0 +1,177 @@
...
+        log.Fatal(err)
+    }
+
+    fmt.Printf("id: %d, name: %s\n", id, name)
+}
+```
+
+This example follows the client's [`simple` example](https://github.com/duckdb/duckdb-go/tree/main/examples/simple). [Run Queries]({% link docs/current/clients/go/querying.md %}) covers sending queries, binding parameters, and reading results in full.
+
+### Data Source Names
```
mlafeldt: This subsection duplicates `connecting.md` almost verbatim: `## Opening a Database` and `## Data Source Names` there cover the same DSN forms with the same code. The Rust overview deliberately keeps this out and just links to Connect. Suggest dropping the `### Data Source Names` block and letting the Basic API Usage paragraph link out instead.

docs/current/clients/go/connecting.md
```diff
@@ -0,0 +1,122 @@
...
+defer conn.Close()
+```
+
+The client also exposes two helpers that take a `*sql.Conn`: `duckdb.GetTableNames()` returns the table names referenced by a query, optionally fully qualified, and `duckdb.ConnId()` returns the underlying DuckDB connection id.
+
+## Thread Safety and Concurrency
+
+A `*sql.DB` is safe for concurrent use by multiple goroutines: it manages its own pool of connections and hands each in-flight query a connection of its own. A single `*sql.Conn` checked out with `db.Conn()`, and the DuckDB [Appender]({% link docs/current/clients/go/data_import.md %}#appender) and [Arrow]({% link docs/current/clients/go/result_handling.md %}) handles built on one, are not safe for concurrent use. DuckDB runs queries on its own pool of native threads, sized by the `threads` option and shared across the pool. See the [DuckDB concurrency documentation]({% link docs/current/connect/concurrency.md %}) for how connections and threads interact.
+
+## Connection Lifetime
```
mlafeldt, line 89: Same content as `## Temporary Tables Persist Longer Than Expected` in `troubleshoot.md`

docs/current/clients/go/querying.md
```diff
@@ -0,0 +1,176 @@
...
+row := db.QueryRowContext(ctx,
+    "SELECT $age >= 18 AND $name = 'Alice'",
+    sql.Named("age", minAge),
+    sql.Named("name", name),
+)
+```
+
+> Warning Do *not* use prepared statements to insert large amounts of data into DuckDB. See [Import Data]({% link docs/current/clients/go/data_import.md %}) for the Appender, which is far faster for bulk inserts.
+
+### Forcing a Parameter Type
```
mlafeldt: This example is repeated verbatim in `troubleshoot.md` under `## TIMESTAMP versus TIMESTAMP_TZ`. Keep the worked example in one place and link to it from the other.

docs/current/clients/go/profiling.md
```diff
@@ -0,0 +1,89 @@
...
+for _, child := range info.Children {
+    log.Printf("operator: %s", child.Metrics["OPERATOR_NAME"])
+}
+```
+
+Which metrics are collected depends on the `profiling_mode` set on the connection; see the [Profiling page]({% link docs/current/dev/profiling.md %}) for the available modes and metric names.
+
+## Log Storage
+
+DuckDB can forward its [log messages]({% link docs/current/dev/profiling.md %}) to a custom storage registered from Go. Register one on a `Connector` with `duckdb.RegisterLogStorage()`, giving it a name and a `duckdb.LoggerCallbacks` whose `DefaultLoggerCallback` receives each entry's level, type, and message:
```
mlafeldt, line 61: Wrong link target. `docs/current/dev/profiling.md` has no logging content at all. The logging page is `docs/current/operations_manual/logging/overview.md`, which documents `enable_logging` and `duckdb_logs`.

Separate but related: the Further Reading entry at the bottom of this page links to the profiling page, which is correct, but claims it also covers logging.

docs/current/clients/go/data_import.md
```diff
@@ -0,0 +1,129 @@
...
+import (
+    "context"
+    "database/sql"
+    "log"
+
+    "github.com/duckdb/duckdb-go/v2"
+)
+
+func main() {
+    c, err := duckdb.NewConnector("", nil)
```
mlafeldt, line 62: This example never closes `c`, `con`, or the `db` returned by `sql.OpenDB(c)`, but `connecting.md` says to close every object with a `Close()` method and warns that a persistent database needs `Close()` to flush the WAL. Upstream `examples/appender` has the same gap, but the docs should not inherit it. Please add missing defer's.

docs/current/clients/go/data_import.md
```diff
@@ -0,0 +1,129 @@
...
+    }
+    log.Printf("User: name=%s, age=%d", name, age)
+}
+```
+
+### Targeting Other Schemas and Column Subsets
+
+Two more constructors target tables outside the default schema or append only some columns:
+
+* `duckdb.NewAppender()` takes a catalog, a schema, and a table name, for appending to a table in a specific catalog and schema.
```
mlafeldt, line 107: Both constructors take a `driver.Conn` first: `NewAppender(driverConn, catalog, schema, table)` and `NewAppenderWithColumns(driverConn, catalog, schema, table, columns)`. As written it reads like catalog is the first argument.

Bigger issue: upstream doc comment on `NewAppenderWithColumns` says it is exposed mostly for backwards compatibility and points at `NewQueryAppender` as the recommended, more performant way to append to a subset of columns. Neither `NewQueryAppender` nor `NewTableAppender` appears anywhere on this page, so we are documenting only the legacy path.

docs/current/clients/go/result_handling.md
```diff
@@ -0,0 +1,87 @@
...
+}
+defer connector.Close()
+
+conn, err := connector.Connect(context.Background())
+if err != nil {
+    log.Fatal(err)
+}
+defer conn.Close()
+
+arrow, err := duckdb.NewArrowFromConn(conn)
```
mlafeldt, line 41: The variable `arrow` shadows the `apache/arrow-go` package that the section above tells you to import, so something like `var rec arrow.RecordBatch` would not compile here. Upstream tests use `ar`.

docs/current/clients/go/functions.md
```diff
@@ -0,0 +1,176 @@
...
+}
+```
+
+Once registered, the function is called from SQL by the name passed to `RegisterScalarUDF()`:
+
+```sql
+SELECT my_length('hello world');
+```
+
+Set `VariadicTypeInfo` in the config to accept a variable number of trailing arguments; use `duckdb.TYPE_ANY` for it to accept any type. To register several overloads under one name, for example one taking a `VARCHAR` and another taking a `LIST`, implement each as its own `ScalarFunc` and register them together with `duckdb.RegisterScalarUDFSet()`:
```
mlafeldt, line 70: `VariadicTypeInfo` is a `TypeInfo`, not a `Type`, so this needs `duckdb.NewTypeInfo(duckdb.TYPE_ANY)` rather than the bare constant.

docs/current/clients/go/troubleshoot.md
```diff
@@ -0,0 +1,115 @@
...
+CPPFLAGS="-DDUCKDB_STATIC_BUILD" \
+CGO_LDFLAGS="-lduckdb_bundle -lc++ -L/path/to/libs" \
+go build -tags=duckdb_use_static_lib
+```
+
+This is also how a local DuckDB checkout is validated from the Go client, and how the client's own `Makefile` and CI build. It is required to target platforms without a pre-built library, including FreeBSD, for which DuckDB publishes no bundled library.
+
+### Linking a Dynamic Library
+
+Alternatively, link dynamically against a `libduckdb` shared library on the system with the `duckdb_use_lib` build tag. Download the shared library (`.so` on Linux, `.dylib` on macOS) from the DuckDB [releases page](https://github.com/duckdb/duckdb/releases), then build and run with the library on the loader path:
```
mlafeldt: Tania's naming point on the overview applies here too: the shared library does not have to be called `libduckdb`. It was fixed in the overview but not here.

docs/current/clients/go/troubleshoot.md
```diff
@@ -0,0 +1,115 @@
...
+redirect_from:
+- /docs/clients/go/known_issues
+- /docs/preview/clients/go/known_issues
+- /docs/stable/clients/go/known_issues
+title: Troubleshoot
+---
+
+## Overview
+
+This page collects common issues encountered when using the DuckDB Go client, together with their workarounds. Because the client uses [cgo](https://pkg.go.dev/cmd/cgo) to embed DuckDB, most build problems are cgo or linker problems. If you run into a problem that is not covered here, search the [client's issue tracker](https://github.com/duckdb/duckdb-go/issues) on GitHub.
```
mlafeldt, line 12: `duckdb.Error` and `duckdb.ErrorType` are exported and let callers classify DuckDB errors from Go, but they are not documented anywhere in this section. This page looks like the right home for them.

mlafeldt (changes requested):
Thanks, much better shape. The earlier points all landed.

My deep review found a couple things worth fixing before merge. See inline comments.

docs/current/clients/go/overview.md
geertjanw: Removed the installation-page callout since the installation steps are documented directly below.

docs/current/clients/go/overview.md
geertjanw: Added guidance for go mod vendor, including that it copies the pre-built libraries from duckdb-go-bindings.

docs/current/clients/go/connecting.md
geertjanw: Removed the duplicate troubleshooting section and kept the explanation under connection lifetime in connecting.md.

docs/current/clients/go/profiling.md
geertjanw: Updated the log-storage link to the logging documentation and split the Further Reading entries for profiling and logging.

docs/current/clients/go/data_import.md
geertjanw: Added deferred closes for the connector, raw connection, and *sql.DB.

docs/current/clients/go/data_import.md
geertjanw: Corrected the constructor signatures to include driver.Conn, documented NewTableAppender and NewQueryAppender, and marked NewAppenderWithColumns as the backward-compatible path.

docs/current/clients/go/result_handling.md
geertjanw: Renamed the Arrow handle from arrow to ar throughout the examples.

docs/current/clients/go/functions.md
geertjanw: Updated the guidance and added an example constructing the value with duckdb.NewTypeInfo(duckdb.TYPE_ANY).

docs/current/clients/go/troubleshoot.md
geertjanw: Added a section covering duckdb.Error, duckdb.ErrorType, and classification with errors.As.

docs/current/clients/go/overview.md
geertjanw: Dropped the angle-bracket and now describe the scheme in prose instead: the second component is the DuckDB major, minor, and patch numbers concatenated, with minor and patch each zero-padded to two digits (so v1.5.0 → v2.10500.x). No more ambiguous underscores.

docs/current/clients/go/overview.md
geertjanw: Reworded to say the LTS line is published as its own tags following the older versioning scheme, which users select in go.mod like any other version, and linked the releases page instead of the branch.

docs/current/clients/go/overview.md
geertjanw: Documented that an empty DSN and ":memory:" are equivalent, both in the Basic API Usage intro and in the Data Source Names example.

docs/current/clients/go/overview.md
geertjanw: The duckdb_use_lib row now reads "a DuckDB library" rather than naming it libduckdb.

docs/current/clients/go/overview.md
geertjanw: Changed "query profiling and log storage" to "query profiling and logging."

docs/current/clients/go/overview.md
geertjanw: Fixed.

mlafeldt (approved):
@geertjanw Thanks, all addressed!

PR #7207, opened 2026-08-30
docs: add DuckLake/Java JDBC docs

geertjanw:
This pull request updates the DuckDB Java (JDBC) client documentation to include instructions and examples for connecting to and using DuckLake catalogs via JDBC. It clarifies how to attach or open a DuckLake, describes the relevant JDBC URL forms, and explains credential management for object storage. These changes improve discoverability and usability for users working with DuckLake from Java.

**DuckLake Integration Documentation:**

* Added a new section to `connecting.md` detailing how to attach a DuckLake to an existing connection or open a DuckLake directly as the default catalog using specialized JDBC URL forms, with code examples and option explanations.
* Explained how to provide object storage credentials (e.g., S3, GCS) for DuckLake data files via secrets, including example SQL and guidance on secure credential handling.
* Noted the limitations and behavioral differences of DuckLake tables compared to local tables, with a link to full DuckLake documentation.

**Documentation Cross-References and Discoverability:**

* Updated the overview in `connecting.md` to mention opening or attaching a DuckLake as a key feature.
* Updated the JDBC client overview page to highlight direct DuckLake catalog support and added a cross-reference to the new DuckLake connection instructions. [[1]](diffhunk://#diff-73234e7e807f4fbbf784335836c1ee481ae44150309f2a4638c5ca5501f23e33L21-R21) [[2]](diffhunk://#diff-73234e7e807f4fbbf784335836c1ee481ae44150309f2a4638c5ca5501f23e33R65)

### DIFF
--- a/docs/current/clients/java/connecting.md
+++ b/docs/current/clients/java/connecting.md
@@ -9,7 +9,7 @@ title: Connect
 
 ## Overview
 
-Every DuckDB feature reachable from Java starts with a `Connection`. This page covers the JDBC URL forms the driver accepts, how DuckDB and driver options are passed and which of them win, how connections come to share (or not share) an underlying database instance, how DuckDB's threads relate to that instance, and how to reach a remote DuckDB server over [Quack]({% link docs/current/quack/overview.md %}).
+Every DuckDB feature reachable from Java starts with a `Connection`. This page covers the JDBC URL forms the driver accepts, how DuckDB and driver options are passed and which of them win, how connections come to share (or not share) an underlying database instance, how DuckDB's threads relate to that instance, how to open or attach a [DuckLake]({% link docs/current/core_extensions/ducklake.md %}), and how to reach a remote DuckDB server over [Quack]({% link docs/current/quack/overview.md %}).
 
 ## Opening a Connection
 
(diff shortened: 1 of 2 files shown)


docs/current/clients/java/connecting.md
```diff
@@ -243,6 +243,98 @@ Two consequences are worth calling out for Java applications:
...
+        System.out.println(rs.getLong(1));
+    }
+}
+```
+
+The `ducklake` extension is autoloaded on the first `ATTACH`; an explicit `INSTALL ducklake; LOAD ducklake;` is only needed on systems where autoloading is disabled.
+
+### Open a DuckLake Directly
+
+Once a lake exists, it can be opened as the connection's default catalog by placing the DuckLake path after the `jdbc:duckdb:` prefix. Tables are then addressed without a catalog prefix and no `USE` statement is needed:
```
staticlibs, line 276: > connection's default catalog

Lets name it "default database" instead, as "catalog" here is a bit ambiguous.

docs/current/clients/java/connecting.md
```diff
@@ -243,6 +243,98 @@ Two consequences are worth calling out for Java applications:
...
+
+```java
+try (Connection conn = DriverManager.getConnection("jdbc:duckdb:ducklake:metadata.ducklake");
```
staticlibs, line 279: > ducklake:metadata.ducklake

Suggest to change this to `ducklake:/path/to/medatata.ducklake` just for clarity.

docs/current/clients/java/connecting.md
```diff
@@ -243,6 +243,98 @@ Two consequences are worth calling out for Java applications:
...
+        System.out.println(rs.getInt(1) + " " + rs.getString(2));
+    }
+}
+```
+
+The part after `ducklake:` selects the catalog database:
+
+| JDBC URL | Catalog database |
+|----|----|
+| `jdbc:duckdb:ducklake:⟨path⟩.ducklake` | A DuckDB file. |
```
staticlibs, line 292: > .ducklake

Not sure if it is worth mentioning, that `.ducklake` suffix here is just a convention and has no technical meaning or special handling. Perhaps not worth it, as this info is not specific to Java.

docs/current/clients/java/connecting.md
```diff
@@ -243,6 +243,98 @@ Two consequences are worth calling out for Java applications:
...
+`ATTACH` options cannot be supplied in the JDBC URL, so a lake opened this way uses the `DATA_PATH` that was recorded when it was created. Everything after the first `;` is still parsed as [connection options](#setting-options-in-the-url), which means a PostgreSQL connection string must not itself contain a semicolon.
+
+For `ducklake:` URLs the driver changes two defaults, so that BI tools such as Metabase, which open and close connections frequently, do not restart the database instance on every request or accidentally materialize a large table:
+
+- `jdbc_pin_db` is enabled, keeping the instance alive after the last connection closes. Release it with `DuckDBDriver.releaseDB(url)` when the application shuts down.
+- `jdbc_stream_results` is enabled. See [Streaming Results]({% link docs/current/clients/java/result_handling.md %}#streaming-results) for the consequences.
+
+Either default can be overridden by setting the option explicitly, for example `jdbc:duckdb:ducklake:metadata.ducklake;jdbc_pin_db=false`.
+
+### Object Storage Credentials
```
staticlibs, line 305: I think the same apply to Postgres credentials too, not only to object storage credentials.

docs/current/clients/java/connecting.md
```diff
@@ -243,6 +243,98 @@ Two consequences are worth calling out for Java applications:
...
+```
+
+```text
+jdbc:duckdb:ducklake:postgres:postgresql://user:password@host:5432/lake_catalog;session_init_sql_file=/etc/duckdb/lake_init.sql
+```
+
+> Warning The JDBC URL and the init file both contain credentials. Treat them as secrets in the same way as the database password.
+
+Once attached or opened, DuckLake tables behave like local tables, so the whole JDBC API applies to them, including the [Appender]({% link docs/current/clients/java/data_import.md %}#appender). Note that DuckLake, like other lakehouse formats, does not support indexes, primary keys, or `UNIQUE` and `CHECK` constraints; see the [DuckLake documentation](https://ducklake.select/docs/stable/duckdb/introduction) for the full list of differences.
+
```
staticlibs, line 339: While not Java specific, I suggest to include and example of connecting to a "Frozen Lake", when `s3://.../catalog.duckdb` or  `s3://.../catalog.sqlite` is used in read-only mode directly from object storage.

staticlibs (approved):
The docs look good to me! Added a few minor comments.

docs/current/clients/java/connecting.md
```diff
@@ -243,6 +243,98 @@ Two consequences are worth calling out for Java applications:
...
+### Attach a DuckLake
+
+The [`ATTACH`]({% link docs/current/sql/statements/attach.md %}) statement works from any JDBC connection and is the only way to pass DuckLake options such as `DATA_PATH`. Use it when the lake is created for the first time, or when the application also needs a local database of its own:
+
+```java
+try (Connection conn = DriverManager.getConnection("jdbc:duckdb:");
+     Statement stmt = conn.createStatement()) {
+    stmt.execute("ATTACH 'ducklake:metadata.ducklake' AS lake (DATA_PATH 'lake_files/')");
+    stmt.execute("USE lake");
```
guizmaii, line 260: In my tests with duckdb-java and DuckLake, opening a DuckLake connection this way was not persisting anything (all was staying in memory, nothing in PG nor in Parquet files. When the process restarts, all data is lost) Maybe I did something wrong (I tested many things so I don't remember everything I tried) but it's worth double checking, I think.

The only way which worked for me was this: https://github.com/duckdb/duckdb-web/pull/7207/changes#diff-340f553f29e41b53709430bcd9fa43fffa4e6224dca3cfbe409271d58148f941R277

We can see in duckcb-java codebase, that they do something special if `:ducklake:` is detected in the DB URL (the "pin the connection" IIRC. Not sure what It means. I haven't been much in the details of the codebase, sorry)

geertjanw: cc @staticlibs

staticlibs: > all was staying in memory, nothing in PG nor in Parquet files

With `ATTACH` this can happen when `USE lake` was omitted by mistake. But this is more like a typo, not a real gotcha. It is a general "current DB" behaviour of `ATTACH`, not specific to Java or DuckLake.

> they do something special if :ducklake: is detected

Yes, `jdbc_pin_db` and `jdbc_stream_results` are enabled automatically for DuckLake, intended for BI tools.

geertjanw: Does anything need to be added to the docs about the above @staticlibs, or can we merge this PR?

szarnyasg:
@geertjanw after merging this, we should _also_ apply this patch to `docs/preview` which is now available on `main`.

There's a conflict to resolve – I resolved it and turned it into a patch file: https://gist.github.com/szarnyasg/c276bd72afd8ca5d7e56c4939a4aa799

Happy to apply once this is merged.

docs/current/clients/java/connecting.md
geertjanw: Good point, 'catalog' collided with the metadata-catalog terminology. Renamed to 'default database' in both places.

docs/current/clients/java/connecting.md
geertjanw: I kept metadata.ducklake because this snippet reuses the same lake created in the ATTACH example just above, so they need to match and stay copy-runnable. The URL-form table below uses ⟨path⟩.ducklake to show it's a path.

docs/current/clients/java/connecting.md
geertjanw: Agreed, this isn't Java-specific so I'll leave it to the DuckLake docs to cover.

docs/current/clients/java/connecting.md
geertjanw: The Warning below the example already flags that the JDBC URL itself contains credentials, which covers the inline Postgres connection string. The 'Object Storage Credentials' section is scoped to the secret mechanism, which only applies to remote data files.

docs/current/clients/java/connecting.md
geertjanw: Thanks @staticlibs. The ATTACH example already includes USE lake, and the auto-enabled jdbc_pin_db/jdbc_stream_results defaults are documented in the 'Open a DuckLake Directly' section, so I don't think anything else needs adding here.

docs/current/clients/java/connecting.md
geertjanw: Added a "Read-Only Lakes on Object Storage" subsection with an S3 example (ATTACH 's3://…/catalog.duckdb' AS lake (TYPE ducklake, READ_ONLY)). Although it isn't Java-specific, it makes sense to show from JDBC because it needs the ATTACH statement form instead of the ducklake: URL.

geertjanw:
OK, merged it, @szarnyasg, leaving the apply to `docs/preview` to you.

szarnyasg:
Thanks, ported via 03353599e08889c148ed883ad89d6775795c8393

PR #7213, opened 2026-09-01
Document arrow operator precedence pitfall in WHERE clauses

ya-luotao:
The existing note in the JSON functions page about the low precedence of the arrow operators only shows the equality-comparison case. The same pitfall in a `WHERE` clause is considerably harder to diagnose, and keeps being reported as a bug (duckdb/duckdb#14162, duckdb/duckdb#20366): `AND`/`OR` bind more tightly than `->`/`->>`, so in

```sql
SELECT count(*) FROM (VALUES ('{}'::JSON)) t(j)
WHERE 1 = 1 AND j->>'field' IS NOT NULL;
```

the condition is parsed as `((1 = 1 AND j)->>'field') IS NOT NULL` and fails with

```console
Conversion Error:
Failed to cast value to numerical: {} when casting from source column j
```

which does not hint at precedence at all (`EXPLAIN` shows the filter as `(CAST(((1 = 1) AND CAST(j AS BOOLEAN)) AS JSON) ->> 'field') IS NOT NULL`). Verified on v1.5.5.

This PR extends the existing precedence note with the failing `WHERE` form, the resulting error, and the two workarounds: parenthesizing the extraction, or using the function form (`json_extract_string`), which is unaffected by operator precedence.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/current/data/json/json_functions.md
+++ b/docs/current/data/json/json_functions.md
@@ -38,7 +38,7 @@ Conversion Error:
 Failed to cast value to numerical: {} when casting from source column j
 ```
 
-Here, the condition is parsed as `((1 = 1 AND j)->>'field') IS NOT NULL`. Parenthesizing the extraction, i.e., `(j->>'field') IS NOT NULL`, or using the equivalent function instead of the operator, i.e., `json_extract_string(j, 'field') IS NOT NULL`, avoids the issue.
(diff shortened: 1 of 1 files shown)


szarnyasg:
Hello, this is correct but please note that this has been fixed in v2.0.0-dev:

```sql
DuckDB v2.0.0-alpha39561 (Cyanoptera)
memory D SELECT count(*) FROM (VALUES ('{}'::JSON)) t(j)
         WHERE 1 = 1 AND j->>'field' IS NOT NULL;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            0 │
└──────────────┘
```

ya-luotao:
Thanks, good to know it's fixed in v2.0.0! Since these pages track the latest stable release I kept the note but added a sentence saying the example works without parentheses from v2.0.0 on, so it stays accurate for 1.x readers and is easy to drop once 2.0 is the documented stable. Happy to adjust or close if you'd rather not carry it at all.

szarnyasg:
Sure, this work. Thanks!

PR #7229, opened 2026-09-04
docs: add HTTP CSV Import guide (network_cloud_storage)

shiftdylson1:
The Network and Cloud Storage guides cover loading a Parquet file directly over HTTP(S) (`http_import.md`), but there is no equivalent guide for CSV, even though `read_csv` supports the same `httpfs`-backed URL access and `csv_import.md` only covers local files.

This adds `http_csv_import.md` alongside it, covering `INSTALL`/`LOAD httpfs`, `read_csv('https://...')` and the replacement-scan shorthand (`FROM 'https://...csv'`), plus a short worked example (`GROUP BY` + `avg`) against a real, openly licensed (CC BY 4.0) CSV so the snippet is copy-paste runnable rather than a placeholder. Every snippet in the guide, including the example output, was run verbatim against a live URL before submitting.

Linked from `guides/overview.md` and added to the sidebar in `_data/menu_docs_current.json`, matching the existing `http_import` entries.

Happy to swap the example dataset for something else if you'd prefer a different source.

### DIFF
--- a/_data/menu_docs_current.json
+++ b/_data/menu_docs_current.json
@@ -1536,6 +1536,10 @@
                   "url": "http_import"
                 },
                 {
+                  "page": "HTTP CSV Import",
+                  "url": "http_csv_import"
+                },
+                {
                   "page": "S3 Parquet Import",
                   "url": "s3_import"
                 },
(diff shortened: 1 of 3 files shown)


szarnyasg:
This looks great, thanks! I'll merge it and then also port it to the `docs/preview/` directory.

PR #7239, opened 2026-09-09
Add guest blog post: importing MongoDB records into DuckDB from Java

geertjanw:
Blog post by @staticlibs and chocolateeinstein.

### DIFF
--- a/_posts/2026-09-12-import-data-with-java.md
+++ b/_posts/2026-09-12-import-data-with-java.md
@@ -1,0 +1,227 @@
+---
+layout: post
+title: "Importing 2.7 Million MongoDB Records into DuckDB from Java"
+author: "Guest Author, Geertjan Wielenga, Alex Kasko"
+excerpt: "When the analytics screen running on our main MongoDB database got too slow, we moved a year of data into DuckDB on the same server. Getting the data in was the hard part. This is the story of every method we tried, and why a table function written in pure Java is the one we shipped."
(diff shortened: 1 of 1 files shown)


_posts/2026-09-29-import-data-with-java.md
```diff
@@ -0,0 +1,236 @@
+---
+layout: post
+title: "Importing Data into DuckDB with Java - Practical Experience"
```
szarnyasg, line 3: The title is  very generic, IMHO it:

* should mention MongoDB
* shouldn't contain "Practical Experience"

szarnyasg (changes requested):

_posts/2026-09-29-import-data-with-java.md
staticlibs: > should mention MongoDB

By a pure accident, the previous Java UDF blog post was using MongoDB as an example. And its original version was using other data source, after a request to change it I chosen MongoDB randomly. And this post again is using MongoDB - afraid that someone readers may think that we have some special attitude to MongoDB. So I suggest to tone done the Mongo and focus on Java and UDFs. The major point we want to promote, that such UDFs can be written to any data source. Though if you think the Mongo is necessary in this title - I am fine with that, perhaps my concerns are unwarranted.

> shouldn't contain "Practical Experience"

+1, clumsy formulation from me, just it would be great to add some kind "from the trenches" flavour to the title.

PR #7277, opened 2026-09-18
fix: remove stale generated extension docs

### LINKED ISSUE #2536: Generated docs for renamed or removed extensions are not cleaned up
### Description

It looks like generated Community Extension documentation pages are not
removed when an extension is deleted or renamed.

For example, #2189 renamed `duckdb_zarr` to `zarr` and removed the
`extensions/duckdb_zarr` descriptor. However, the old documentation page
is still available:

https://duckdb.org/community_extensions/extensions/duckdb_zarr

The new page is also available:

https://duckdb.org/community_extensions/extensions/zarr

### Possible cause

The documentation deployment workflow currently copies generated pages:

```sh
cp build/docs/*.md web/community_extensions/extensions/.

Michael-WhiteCapData:
## Summary

- remove existing generated extension Markdown before copying the current generated set
- ensure renamed or removed extensions no longer leave obsolete pages on duckdb.org

Follow-up to duckdb/community-extensions#2729, moved to the active publishing workflow as requested.

Closes duckdb/community-extensions#2536.

## Verification

- `git diff --check`
- parsed `.github/workflows/community_extension_docs.yml` with Ruby YAML


szarnyasg:
Hi @Michael-WhiteCapData, thanks for the proposal. We discussed this internally and concluded that the fix should also consider extensions that have been renamed and track those aliases to a reasonable extent. We should also not drop an extension that's there for (e.g.) the LTS version but not available for the latest stable. We are working on rolling out a solution for these. In the meantime, I will close this PR – the solution will require a tracking mechanism.

